PR #9, opened 2016-04-05
No bug - Pass stdout through to console. r=me

ncalexan:


victorporof:
This is already the case now. Not sure when this merged, but it did.

Issue #27, opened 2016-04-06
Establish dedicated BrowserWindow for main process communication


ncalexan:
A number of things are exposed to content processes via HTML5 APIs, but not directly to the main process.  (For example, https://github.com/electron/electron/blob/master/docs/tutorial/online-offline-events.md.)

Rather than negotiating 0/1/many content windows, I suggest we add a single long-lived hidden rendering process to handle funneling HTML5 APIs to the main process.

jsantell:
For non-BW specific states, like online/offline, this sounds great; are there any other APIs that you think would be useful to have in the main process?

jsantell:
This issue is no longer relevant I believe. Please reopen if you feel otherwise.

PR #28, opened 2016-04-07
Issue #27: Establish hidden `BrowserWindow` for main process communication

ncalexan:
@Mossop comments?

It's not clear that a separate directory and scripts will pay off, but it's done.

I'm not 100% clear on the best way to have A keep a reference to B *without* exposing the reference.  It's also not clear to me when module level const defines go out of scope -- perhaps the efforts I made aren't actually necessary?  (Or effective?)

### DIFF
--- a/app/services/hidden-window.js
+++ b/app/services/hidden-window.js
@@ -1,0 +1,56 @@
+/* eslint no-console: 0 */
+
+/**
+ * The Electron main process does not receive many HTML5 events, including
+ * online and offline status changes.  The recommended approach is to
+ * pass-through the HTML5 events from a content process to the main process: see
+ * https://github.com/electron/electron/blob/master/docs/tutorial/online-offline-events.md.
+ *
+ * We implement this approach here in a single, extensible place.  We maintain a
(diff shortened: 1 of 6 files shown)


app/services/hidden-window.js
```diff
@@ -0,0 +1,56 @@
...
+
+app.on('ready', function() {
+  console.log('ready');
+  hiddenBrowserWindow = new BrowserWindow({ width: 0, height: 0, show: false });
+  hiddenBrowserWindow.loadURL('file://' + __dirname + '/hidden-window/hidden-window.html');
+
+  // // To debug the hidden window JS, uncomment the following:
+  // if (!isProduction()) {
+  //   hiddenBrowserWindow.openDevTools({ detach: true });
+  // }
```
ncalexan: I agree.  I haven't landed this yet because the consumer I expected to use it doesn't yet use it.  I'm thinking we may end up doing some storage stuff in this background process, too, so just leaving it hanging.

ncalexan:
I'm going to close this.  We can re-open and rebase if we need.

PR #34, opened 2016-04-07
Make the navbar more of a presentation component that does not define

jsantell:
its own handlers, and does not receive a dispatch method. Also clean up
the react-redux connect method used for some components and not others,
and only handle this at the root BrowserWindow state.

Things like `Page` and `TabBar` still use dispatch directly, which should be handled soon as well

### DIFF
--- a/ui/browser/views/browser.jsx
+++ b/ui/browser/views/browser.jsx
@@ -11,15 +11,18 @@ specific language governing permissions and limitations under the License.
 */
 
 import React, { PropTypes, Component } from 'react';
-import { connect } from 'react-redux';
 
 import browserDB from '../../shared/browser-db';
 import TabBar from './tabbar/tabbar.jsx';
 import NavBar from './navbar/navbar.jsx';
 import Page from './page/page.jsx';
 
-import { updateMenu } from '../actions/external';
-import { createTab, attachTab, setPageDetails } from '../actions/main-actions';
+import {
+  bookmark, unbookmark, menuLocationContext,
+  updateMenu, menuBrowser, maximize, minimize, close
+} from '../actions/external';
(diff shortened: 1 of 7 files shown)


ui/browser/views/browser.jsx
```diff
@@ -34,21 +37,39 @@ class BrowserWindow extends Component {
   }
 
   render() {
-    const { pages, currentPageIndex, pageOrder, ipcRenderer } = this.props;
+    const { dispatch, pages, currentPageIndex, pageOrder, ipcRenderer } = this.props;
     const platformClass = `platform-${platform}`;
+    const navBack = e => getCurrentWebView(e.target.ownerDocument).goBack();
```
bgrins: Any reason that some of these are defined above the jsx and others are in line (like bookmark)?

ui/browser/views/navbar/navbar.jsx
```diff
@@ -10,60 +10,76 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
...
+  navBack: PropTypes.func.isRequired,
+  navForward: PropTypes.func.isRequired,
+  navRefresh: PropTypes.func.isRequired,
+  minimize: PropTypes.func.isRequired,
+  maximize: PropTypes.func.isRequired,
+  close: PropTypes.func.isRequired,
+  openMenu: PropTypes.func.isRequired,
+
+  // For <Location>
+  bookmark: PropTypes.func.isRequired,
```
bgrins: Passing these props all the way down the component hierarchy might get noisy / annoying (having to copy/paste the same thing in a number of places).  But, seems like the right direction to go in for testability and simplifying the components so let's go with it

jsantell:
Handled in #47

Issue #17, opened 2016-04-06
Add open state for 'Pages' section


bgrins:
When clicking the 'Pages' button let's toggle a placeholder area

PR #36, opened 2016-04-07
Start wiring up buttons to use real assets in the navbar.

jsantell:
r? @bgrins

### DIFF
--- a/ui/browser/views/navbar/btn.jsx
+++ b/ui/browser/views/navbar/btn.jsx
@@ -12,21 +12,41 @@ specific language governing permissions and limitations under the License.
 
 import React, { PropTypes } from 'react';
 
+const HEIGHT = "20px";
+const WIDTH = "20px";
+
 /**
  * A button for the NavBar.
- * `icon` must be a valid font-awesome icon name
+ * `image` must be a valid CSS path to an image.
  */
-const Btn = ({ disabled, title, icon, onClick }) => (
-  <a href="#" className={disabled ? 'disabled' : ''} {...{ title, onClick }}>
-    <i className={`fa fa-${icon}`} />
+const Btn = ({ disabled, title, image, onClick }) => (
+  <a href="#"
+     style={{
+       backgroundImage: image ? `url(assets/${image})` : `url(${DEFAULT})`,
+       backgroundRepeat: "no-repeat",
+       backgroundSize: `${HEIGHT} ${WIDTH}`,
+
+       opacity: (disabled ? "0.5" : "1"),
+       cursor: (disabled ? "default" : "pointer"),
+
+       margin: "8px 3px 0px 3px",
+       width: WIDTH,
+       height: HEIGHT,
+       display: "inline-block",
+       padding: "0",
+     }}
+     {...{ title, onClick }}>
   </a>
 );
 
 Btn.propTypes = {
   disabled: PropTypes.bool,
   title: PropTypes.string.isRequired,
-  icon: PropTypes.string.isRequired,
+  image: PropTypes.string.isRequired,
   onClick: PropTypes.func.isRequired,
 };
 
 export default Btn;
+
+// Use a data URI to a Mozilla "M" for now if we don't have an asset for a button.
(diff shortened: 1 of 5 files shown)


ui/browser/views/navbar/btn.jsx
```diff
@@ -12,21 +12,41 @@ specific language governing permissions and limitations under the License.
...
...c7AZPoJCGxjiWjsKikq6vbmTSYsAS49Nwe58HnzjXMyrBvm1tUGjHzc28FSIqKSwjTm8tnUdP41dlzzpu3MTsPF7Fd4TjxJzaPuFl68+at2Vr16NHPkVExRrNFLHsCgkIxED4P07aSHYBJrOADyz8ipPWZ2b29vaKYd7q2Dp4JjxW7cDYSQVJWdu7joaG5XbhDmhH5IcL0mg4eOoKLb2IFkmF6IwJnhDsiHvjgsUaTRcQ9Aa4EIpjbu690zlahg3CZCJsI02uqrKp5KzBEnFIvlRbXGkWv4EtIXKoLN4s1bAajBffsKitrnA4PdqBYijC9pn+cqBIRppRVqeKaBx+Ql2dTaURLq+JCX+aGbOcNQ00wbi4QJhfCtDJljegWYpujUIl24IPacORpnbcKqyRiTcLkYzDheymwoIjomXAwQpjmKUz4kkIMG2EiTOJsEZC48jaY8F08hMn3YDpRWe2FMO3bX0aYCJM4MGHxxZehESYfg6mhwY7aNG+DCWVYfwsMJUw+BhO+XTMgSOZtMKFYhTD5HkxNTc2BwYSJMBEmwkSYCBNhIkyEiTARJsJEmAgTYSJMhIkwESbCRJgIE2EiTISJMBEmwkSYCBNhIkyEiTARJsJEmAgTYSJMhIkwEab/tlM3KQ2DURhGt6RNY2KsRcSRFBE3YQWd+4MuwIquQnGouIXiTpJWCLVCAhmaPeSropyHu4B3cLgwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDBBBNMMMEEE0wwwQQTTDDB9EuYHp+e16Jkf3TY/baGu8fj0+ALp9P3KM6CLGwvTgfnl9fdV728vq1HaahVvX56O7n/D5h68ebo4Kj7be/sjU/OVvGZ4mQQZGF7STa8uLoJgqklHmpVfyOb3D38eUxVVeV5MZ9/dL9iNivLMvjCpmlCLWwvL4rF4rP7qrquQ67Ki+Xya8WYJJgEk2CSYBJMgkkwSTAJJsEkwSSYBJNgkmASTIJJgkkwCSbBJMEkmASTBJNgEkyCSYJJMAkmCSbBJJgEkwSTYBJMEkyCSTAJJgkm/XzfL07KBwH/1dEAAAAASUVORK5CYII=";
```
bgrins: Why is this at the bottom of the file?  Also, how about we don't set a background image if it's not passed in instead of using a default (which shouldn't ever be used)?  I was planning to convert the 'pages' button to a Btn for #17 so it could have arbitrary child elements with no icon.  Maybe I should instead make a new component for text buttons but I was thinking we could re-use things like the disabled / active state.

jsantell: I have no strong feelings one way or another; just wanted to display anything for buttons that lack assets (like bookmarks in this case)

ui/browser/views/navbar/btn.jsx
```diff
@@ -12,21 +12,41 @@ specific language governing permissions and limitations under the License.
...
- * `icon` must be a valid font-awesome icon name
+ * `image` must be a valid CSS path to an image.
  */
-const Btn = ({ disabled, title, icon, onClick }) => (
-  <a href="#" className={disabled ? 'disabled' : ''} {...{ title, onClick }}>
-    <i className={`fa fa-${icon}`} />
+const Btn = ({ disabled, title, image, onClick }) => (
+  <a href="#"
+     style={{
+       backgroundImage: image ? `url(assets/${image})` : `url(${DEFAULT})`,
```
bgrins: I'm not a fan of the implications of sticking all of the styles into JS.  In this case I'm on board with passing in 'image' which sets inline style for background image instead of using icon class, but I'd expect the rest of the styles to live in CSS.  Specifically:

* For theming, how would that work?  Would all the various colors / font sizes be passed in as props to the component?  Seems easier to override the styles using CSS selectors.
* What about hover states?
* What about animations?

That said, I haven't used CSS in this way before so maybe I'm missing obvious answers and I'm open to be convinced.  But if it involves pulling in more libraries to handle these things, I wonder if it's adding enough value to justify it.  So it's not an r- but I want to discuss further about what the benefit is.

jsantell: @victorporof can explain the reasoning here better, this was more to try it out, I think he wants to discuss at tomorrows meeting

jsantell:
Migrated to #51 due to repo changes

victorporof:
There's no problem doing CSS transitions as inlined styles. Hover states are a bit trickier, but manageable with mixins. As discussed yesterday, a hybrid approach will do.

PR #42, opened 2016-04-08
Instrument

ncalexan:
@Mossop would you like to review?

@canuckistani will this address the product team's instrumentation needs?

### DIFF
--- a/app/main/main.js
+++ b/app/main/main.js
@@ -18,6 +18,7 @@ require('babel-register')();
 
 const path = require('path');
 const electron = require('electron');
+const instrument = require('../services/instrument');
 const server = require('../services/server');
 const app = electron.app; // control application life.
 const BrowserWindow = electron.BrowserWindow;  // create native browser window.
@@ -137,6 +138,11 @@ app.on('activate', () => {
   if (mainWindows.length === 0) {
     createWindow();
   }
+});
+
+ipc.on('instrument-event', (event, args) => {
+  // Until we transpile app/, we can't destructure in the argument list or inline here.
+  instrument.event(args.name, args.method, args.label, args.value);
 });
 
 ipc.on('new-window', createWindow);
--- a/app/main/main.js
+++ b/app/main/main.js
@@ -13,6 +13,9 @@ specific language governing permissions and limitations under the License.
 /* eslint-disable strict */
 'use strict';
 
+// Must go before any require statements.
+const browserStartTime = Date.now();
+
 require('babel-polyfill');
 require('babel-register')();
 
@@ -68,6 +71,9 @@ function createWindow(tabInfo) {
   });
 
   browser.webContents.once('did-finish-load', () => {
+    const browserDidFinishLoadTime = Date.now();
+    instrument.event('browser', 'READY', 'ms', browserDidFinishLoadTime - browserStartTime);
+
     browser.show();
 
     if (tabInfo) {
@@ -100,12 +106,18 @@ function registerFileProtocol(scheme, mapper) {
   });
 }
 
+const appStartupTime = Date.now();
+instrument.event('app', 'STARTUP');
+
 // Kick off the internal webserver so we have HMR
 server.serve(staticDir);
 
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 app.on('ready', () => {
+  const appReadyTime = Date.now();
+  instrument.event('app', 'READY', 'ms', appReadyTime - appStartupTime);
+
   registerFileProtocol('atom', url => {
     if (url.indexOf('/') === '-1') {
       url = `${url}/index.html`;
--- a/ui/browser/actions/main-actions.js
+++ b/ui/browser/actions/main-actions.js
@@ -13,32 +13,32 @@ specific language governing permissions and limitations under the License.
 import * as types from '../constants/action-types';
 
 export function createTab(location) {
-  return { type: types.CREATE_TAB, location };
+  return { type: types.CREATE_TAB, location, instrument: true };
 }
 
 export function duplicateTab(pageIndex) {
-  return { type: types.DUPLICATE_TAB, pageIndex };
+  return { type: types.DUPLICATE_TAB, pageIndex, instrument: true };
 }
 export function attachTab(page) {
-  return { type: types.ATTACH_TAB, page };
+  return { type: types.ATTACH_TAB, page, instrument: true };
 }
 
 export function closeTab(pageIndex) {
-  return { type: types.CLOSE_TAB, pageIndex };
+  return { type: types.CLOSE_TAB, pageIndex, instrument: true };
 }
 
 export function setLocation(userTyped) {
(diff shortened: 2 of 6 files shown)


app/main/main.js
```diff
@@ -67,6 +69,9 @@ function createWindow(tabInfo) {
   });
 
   browser.webContents.once('did-finish-load', () => {
+    let browserDidFinishLoadTime = new Date();
```
jsantell: I think `Date.now()` might be more intuitive, or could seem that we're passing in Date objects to the event method (when both Dates are cast to numbers anyway)

ui/browser/actions/main-actions.js
```diff
@@ -13,32 +13,32 @@ specific language governing permissions and limitations under the License.
 import * as types from '../constants/action-types';
 
 export function createTab(location) {
-  return { type: types.CREATE_TAB, location };
+  return { type: types.CREATE_TAB, location, instrument: 1 };
```
jsantell: What's `instrument` here?

app/main/main.js
bgrins: What about `performance.now()`?

bgrins: well, i guess we don't care about sub-milisecond precision here

jsantell: We also don't have web APIs here

bgrins: That's what I thought too, but I checked inside one of the components and it was defined

ui/browser/actions/main-actions.js
ncalexan: On Thu, Apr 7, 2016 at 5:21 PM, Jordan Santell <notifications@github.com>
wrote:

> In ui/browser/actions/main-actions.js
> <https://github.com/mozilla/tofino/pull/42#discussion_r58965928>:
>
> > @@ -13,32 +13,32 @@ specific language governing permissions and limitations under the License.
> >  import * as types from '../constants/action-types';
> >
> >  export function createTab(location) {
> > -  return { type: types.CREATE_TAB, location };
> > +  return { type: types.CREATE_TAB, location, instrument: 1 };
>
> What's instrument here?
>
I renamed this a few times before settling on shorter.  It's a flag
indicating whether to report ("instrument") these events to our metrics
service (GA for now).

I'd be happy with `shouldInstrument`, or really anything you care to
counter with.

LegNeato:
This may be of help as well:

https://github.com/facebook/nuclide/tree/master/pkg/nuclide-analytics

ncalexan:
@LegNeato thanks, nifty.  Flow and Rx make it hard for us to use, but some nice ideas in there.

app/main/main.js
```diff
@@ -99,12 +104,19 @@ function registerFileProtocol(scheme, mapper) {
   });
 }
 
+const appStartupTime = new Date();
```
Mossop: This needs to go right up to the top before any require statements to give us a full measure

app/main/main.js
Mossop: I agree that Date.now() makes more sense.

build/config-builder.js
```diff
@@ -42,6 +42,9 @@ const BASE_CONFIG = {
 
   // Whether or not the build is packaged in an electron distributable
   packaged: false,
+
+  // The Google Analytics tracking identifier to use when instrumenting.
+  google_analytics_UA: 'UA-76122102-1'
```
Mossop: Don't we want this to be secret-ish? At least coming from somewhere outside of the tree means every fork won't report to our account.

ui/browser/store/store.js
```diff
@@ -13,6 +13,14 @@ specific language governing permissions and limitations under the License.
 import { applyMiddleware, createStore } from 'redux';
 import createLogger from 'redux-logger';
 import rootReducer from '../reducers';
+import * as instrument from '../../shared/instrument';
+
+const instrumenter = store => next => action => {
```
Mossop: This syntax makes my head hurt, is there really not a better way to write this?

ui/browser/actions/main-actions.js
Mossop: If it's a flag why not a boolean?

ui/browser/store/store.js
jsantell: It's a bit hard to look at, but is the recommended way for middleware FWIW

ncalexan: On Fri, Apr 8, 2016 at 3:51 PM, Jordan Santell <notifications@github.com>
wrote:

> In ui/browser/store/store.js
> <https://github.com/mozilla/tofino/pull/42#discussion_r59099690>:
>
> > @@ -13,6 +13,14 @@ specific language governing permissions and limitations under the License.
> >  import { applyMiddleware, createStore } from 'redux';
> >  import createLogger from 'redux-logger';
> >  import rootReducer from '../reducers';
> > +import * as instrument from '../../shared/instrument';
> > +
> > +const instrumenter = store => next => action => {
>
> It's a bit hard to look at, but is the recommended way for middleware FWIW
>

Yeah, straight out of the docs.  It's like 10000 spoons when all you want
is to curry transparently...

Mossop: Well I warn you, if we do this more than once I'm going to wrap it in something easier on the eyes ;)

build/config-builder.js
ncalexan: I have no particularly strong opinion.  I see that Atom includes there tracking identifier in source https://github.com/atom/metrics/blob/master/lib/reporter.coffee#L109.  It wouldn't be hard to extract from the code.  This is just a personal test account tracking identifier; let's land it as is, and hide it if necessary when we move to a Mozilla-funded account.

ui/browser/actions/main-actions.js
ncalexan: It should be.  Will update.

ncalexan:
@Mossop re-review?  I changed to use `Date.now()`; started timing app startup earlier; and replaced `1` with `true`.

ui/shared/instrument.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {string} name - Event name or category; equivalent to UI Telemetry's event.
+ * @param {string} method - Event method or action; equivalent to UI Telemetry's method.
+ * @param {string} [label] - Optional label.
+ * @param {integer} [value] - Optional value.
+ * @return {Promise} A Promise resolving to null.
+ */
+export function event(name, method, label = null, value = null) {
+  ipcRenderer.send('instrument-event', { name, method, label, value });
+}
```
Mossop: So one option to avoid having to change this argument list in so many places is just to make this:

    export function event(...args) {
      ipcRenderer.send('instrument-event', args)
    }

Then up where it is received you can just call instrument.event.apply(instrument, args);

Mossop:
R+ with the lint failures fixed, take my other comment or not, whichever you prefer.

ui/shared/instrument.js
ncalexan: I considered this, but it loses IDE completion :(

PR #47, opened 2016-04-08
Make the navbar more of a presentation component that does not define

jsantell:
its own handlers, and does not receive a dispatch method. Also clean up
the react-redux connect method used for some components and not others,
and only handle this at the root BrowserWindow state.

(Having to recreate this PR due to repo name changes, privacy changes, etc), original: https://github.com/mozilla/tofino/pull/34


jsantell:
r+'d @bgrins in #34

PR #4, opened 2016-04-05
Bump electron-prebuilt version

bgrins:
Signed-off-by: Brian Grinstead <briangrinstead@gmail.com>

I was getting an error before updating:

    > electron-prebuilt@0.37.2 postinstall /fawkes/node_modules/electron-prebuilt
    > node install.js

    module.js:327
        throw err;
        ^

    Error: Cannot find module 'process-nextick-args'

PR #49, opened 2016-04-08
Bump electron-prebuilt version

bgrins:
Since I filed #4 the a new version has been released.  So here's a new version bump.  r? @rnewman


rnewman:
Ship it!

PR #52, opened 2016-04-08
Make travis run tests.

Mossop:
Travis doesn't support node_js containers for OSX so instead this just uses the
C++ container and then uses nvm to install an appropriate node version to run
with.


jsantell:
r+

Issue #54, opened 2016-04-08
Browser chrome doesn't interact


jonathanKingston:
All of the interface of the top bar is not clickable (url bar, reload, close etc) and I get the grabbing hand cursor.

The tabs and new button are clickable and selectable.

I'm running this on Ubuntu 16.04 with Node 5.8.0.

PR #56, opened 2016-04-08
Issue #54: Fixing clickable navbar

jonathanKingston:


jsantell:
Works for me without this patch in OSX, but I do find it weird that I can drag via buttons/input, but that decision was made due to the navbar being mostly buttons. I think the long term solution will probably be some UX work to make sure we have draggable space visible anyway, but can tackle that later.

Thanks @jonathanKingston !

PR #57, opened 2016-04-08
Enhance URL handling.  Fixes #7.

### LINKED ISSUE #7: Better URI detection for user input in nav bar
See https://github.com/mozilla/fawkes/pull/5#issue-145877708 about coming up with a better detection for data URIs.  Here are some relevant looking utilities:

* http://mxr.mozilla.org/gaia/source/shared/js/url_helper.js
* http://mxr.mozilla.org/gaia/source/apps/sharedtest/test/unit/url_helper_test.js
* https://github.com/brave/browser-laptop/blob/master/js/lib/urlutil.js

bgrins:
Adds support for data and view-source URLs, and expands test coverage
for both the new url-util file and the browser-util file that converts
search terms into search URLs.

r? @joewalker can we include MPL source code here (url-util.js)?  If not we could use this one with modifications (which is already Apache): http://mxr.mozilla.org/gaia/source/shared/js/url_helper.js


bgrins:
I got confirmation from dcamp that MPL2 is compatible with Apache 2

bgrins:
OK updated the patch with a comment about origin of the url-util file.  Redirecting r? to @jsantell since licensing is figured out.  Jordan, the util code includes some functions that we aren't using right now around view source.  Let me know if you think they should be removed.

test/ui/browser/test-browser-util.js
```diff
@@ -10,6 +10,11 @@ describe('utils', function() {
       const UNCHANGED_URIS = [
         'data:text/html,',
         'http://example.com',
+        'http://foo',
+        'http://foo.com',
```
jsantell: Should add an `https` here too

test/ui/browser/test-browser-util.js
```diff
@@ -18,7 +23,12 @@ describe('utils', function() {
 
     it('should append protocol onto URIs', function() {
       const CHANGED_URIS = [
-        ['example.com', 'http://example.com'],
+        ['example.com', 'http://example.com/'],
+        ['example.net', 'http://example.net/'],
+        ['foo.bar', 'http://foo.bar/'],
+        ['blerg.co.uk', 'http://blerg.co.uk/'],
+        ['blach.com', 'http://blach.com/'],
+        ['www.blah.com', 'http://www.blah.com/'],
```
jsantell: We should ensure that something like 'example.com blah' doesn't add a protocol (like current Firefox, it'd redirect to a search query)

jsantell: NM that's handled below

jsantell: We should test leading with a space, like '   mozilla.org' -> 'http://mozilla.org'

test/ui/shared/test-url-util.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+  });
+
+  describe('getScheme', function() {
+    it('is null for empty', function() {
+      expect(UrlUtil.getScheme('/file/path/to/file')).toEqual(null);
+    });
+
+    it('is null for localhost', function() {
+      expect(UrlUtil.getScheme('localhost://127.0.0.1')).toEqual(null);
```
jsantell: I think the scheme for this should be `localhost:`, no?

test/ui/shared/test-url-util.js
```diff
@@ -0,0 +1,61 @@
...
+    it('defaults to http://', function() {
+      expect(UrlUtil.prependScheme('www.mozilla.org')).toEqual('http://www.mozilla.org');
+    });
+
+    it('keeps schema if already exists', function() {
+      expect(UrlUtil.prependScheme('https://www.mozilla.org')).toEqual('https://www.mozilla.org');
+    });
+  });
+
+  describe('isURL', function () {
```
jsantell: Should test the falsy scenario here, and aren't these technically URI's, not URLs?

ui/browser/browser-util.js
```diff
@@ -23,12 +25,16 @@ export function getSearchURL(query) {
  * Convert whatever the user typed into the URL bar into an actual URL
  */
 export function fixURL(typed) {
-  if (typed.includes('://') || typed.trim().startsWith('data:')) {
+  if (UrlUtil.isDataUrl(typed)) {
```
jsantell: should consistently uppercase `URL`, like `isURL`, or make `isURL` camel case

ui/shared/url-util.js
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Prepends file scheme for file paths, otherwise the default scheme
+   * @param {String} input path, with opetional schema
+   * @returns {String} path with a scheme
+   */
+  prependScheme(input) {
+    // expand relative path
+    if (input.startsWith('~/')) {
+      input = input.replace(/^~/, os.homedir());
```
jsantell: How does `os.homedir()` work when this runs in content process? Or in the main process for that matter? These are all going to be transpiled during build and not processed at runtime, leaving most `os` commands useless or misleading. To detect current os/system/arch, we should use `build-config.js` (see current `shared/util.js` using `development` or `packaged`)

ui/shared/url-util.js
```diff
@@ -0,0 +1,177 @@
...
+
+    return input;
+  },
+
+  /**
+   * Checks if a string is not a URL.
+   * @param {String} input The input value.
+   * @returns {Boolean} Returns true if this is not a valid URL.
+   */
+  isNotURL(input) {
```
jsantell: This method makes my head hurt. Should be moved to `isURL` and this removed

ui/shared/url-util.js
```diff
@@ -0,0 +1,177 @@
...
+    return !this.isNotURL(input);
+  },
+
+  /**
+   * Checks if a URL is a view-source URL.
+   * @param {String} input The input URL.
+   * @returns {Boolean} Whether or not this is a view-source URL.
+   */
+  isViewSourceUrl(url) {
+    return url.toLowerCase().startsWith('view-source:');
```
jsantell: We should do `(url || "")` or default arg or something and not throw if we pass in undefined/null

ui/shared/url-util.js
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Converts a view-source url into a standard url.
+   * @param {String} input The view-source url.
+   * @returns {String} A normal url.
+   */
+  getUrlFromViewSourceUrl(input) {
+    if (!this.isViewSourceUrl(input)) {
+      return input;
+    }
+    return this.getUrlFromInput(input.substring('view-source:'.length));
```
jsantell: "view-source:" should be a constant, it's used all over this file

jsantell:
I have many criticisms about the utils, but can handle in a follow up, my main concern is changing/removing the methods in the future once they're already being used, more so than the nits inside the file itself (inconsistent casing, boolean methods via negation)

jsantell:
And we almost definitely need to remove `os` usage, that should probably be blacklisted from any use anywhere in this app

bgrins:
Good points - I hadn't spent much time cleaning that file up.  I'll trim the file down and address the notes

test/ui/browser/test-browser-util.js
bgrins: Done

test/ui/browser/test-browser-util.js
bgrins: Oh, that's actually failing with the current url utils implementation, i'll take a look at that

test/ui/shared/test-url-util.js
bgrins: There's special case in the imported url utils for `scheme === 'localhost://'`.  I'm not sure why

bgrins:
@jsantell I think all the notes should be addressed with the latest commits.   The API surface area is smaller and new test cases have been added.  Can you have another look when you have a chance?

test/ui/browser/test-browser-util.js
```diff
@@ -6,32 +6,75 @@ import { fixURL, getSearchURL } from '../../../ui/browser/browser-util';
...
         'data:text/html,',
         'http://example.com',
+        'http://foo.com',
+        'https://foo.com',
+        'https://foo.com/?a=%20b',
+        'data:about',
+        'view-source:http://foo.com/',
+        'rtsp://100.100.100.100/rtsp.mp4',
+
+        // @TODO Handle spaces in GET parameters
```
jsantell: For tests, we should add pending tests -- which is `it()` with only one string argument, which will show up pending

test/ui/shared/test-url-util.js
```diff
@@ -0,0 +1,61 @@
+/* global describe, it */
```
joewalker: We could probably have a custom .eslintrc to just say this in one place. We probably don't need to do that in this PR though

victorporof: The test folder has its own eslintrc. Should probably add this to the environment, or create a globals array.

But I'm not even sure we need those rules. Aren't those part of the mocha environment, which we already specify?

bgrins: You are right, they were unneeded so I removed them

victorporof:
@bgrins @jsantell What are we doing with this PR?

bgrins:
@jsantell do you remember what we decided with this?  I can rebase and land if that works for you

victorporof:
@jsantell reping

jsantell:
@bgrins this needs a rebase, it's rather out of date

bgrins:
Realized this was using node from frontend.  Need to rethink and hopefully simplify the approach.

PR #60, opened 2016-04-10
Pass NODE_ENV through the env option for exec.

### LINKED ISSUE #59: gulp run fails on windows
It attempts to set the NODE_ENV environment variable as if running under a unix shell.

Mossop:
Fixes #59.


jsantell:
r+, but I don't think the build uses NODE_ENV anymore -- 'production' and 'development' environments don't really make sense in a desktop environment, as we currently have 3 different modes: building locally with hot module reloading, building locally without HMR, and packaging. These scenarios are represented in build-config as two settings, 'development' and 'packaged'. That being said, if this makes it work on windows, let's get it in and figure that out later

PR #62, opened 2016-04-10
Stop using the custom build of electron.

Mossop:
When we were experimenting I had us using a custom build of electron both to
test how easy that would be (not too bad) and because I was adding some features
to electron. We don't need those features right now and I intend to try to
upstream them. Using a custom build breaks attempts to compile binary node
modules for electron. All the documented ways of doing that compile against the
published electron releases and in particular the version number matters and I
had messed with that.

This changeset removes all the additional code that tries to use my custom build
and instead just relies on the electron-prebuilt package to handle the electron
download for us.


jsantell:
r+

PR #69, opened 2016-04-11
Update suggested commit wording to automatically close issues when landing

bgrins:
With "Fixes #N" an the issue N will automatically close once it's pushed: https://help.github.com/articles/closing-issues-via-commit-messages/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ Guidelines:
   spelunkers.
   For example:
   ```
-  Issue #6 - Frobnicate the URL bazzer before flattening pilchard. r=mossop,rnewman
+  Frobnicate the URL bazzer before flattening pilchard. Fixes #6. r=mossop,rnewman
 
   The frobnication method used is as described in Podder's Miscellany, page 15.
   Note that this pull request doesn't include tests, because we're bad people.


victorporof:
r+

victorporof:
Note: that commit message style doesn't automatically close the original issue once it's pushed, you need to reference the issue number inside the PR body.

See https://github.com/mozilla/tofino/pull/71 for an example.

We should update the readme to suggest doing just that. @bgrins care to do it?

victorporof:
I also suggest making a not about asking for review. Pinging somebody in a comment usually works, but it feels better to just assign somebody to a PR if you want their review.

bgrins:
> I also suggest making a note about how to ask for review. Pinging somebody in a comment usually works, but it feels better to just assign somebody to a PR if you want their review.

I'm all for getting rid of email pings as a way to track reviewers, but it may be a separate discussion / issue depending on a couple questions:

1) Does everyone have permissions to set the assignee to anyone else in PRs?
2) Does a user get notified when they are assigned to a PR?

bgrins:
> Note: that commit message style doesn't automatically close the original issue once it's pushed, you need to reference the issue number inside the PR body.

According to https://help.github.com/articles/closing-issues-via-commit-messages/ it looks like it should be enough.

PR #74, opened 2016-04-12
Remove electron from .gitignore r=victorporof

gabrielluong:
Since we no longer use a custom built of electron, we no longer need electron in .gitignore


victorporof:
r+

PR #85, opened 2016-04-12
Fix npm task commands documented in README.md

lmorchard:
Just cloned the repo and tried running through the tasks in the README. Looks like they're a bit off with respect to `package.json` and npm scripts.


jsantell:
Looks good, thanks @lmorchard !

PR #90, opened 2016-04-12
Add ability to toggle 'pages' section (aka tabbar).  Fixes #17. r=jsantell

### LINKED ISSUE #17: Add open state for 'Pages' section

bgrins:
Adds a new TOGGLE_PAGES_SECTION action that allows it to be collapsed.  The state is toggled when the pages button is clicked, and becomes automatically true if a new tab is added.

I converted the Btn component to allow arbitrary children and also optionally to not include an image.  Let me know if you'd rather instead that we not make the pages section a 'Btn' and I'll undo that part - since it might ultimately need very different styling (not sure about that).


ui/browser/views/navbar/navbar.jsx
```diff
@@ -20,8 +20,8 @@ import Location from './location.jsx';
 function NavBar(props) {
   const {
     navBack, navForward, navRefresh, page, pages, openMenu, ipcRenderer,
```
ncalexan: Comment, not issue: this list of props seems untenable over time.  Is this really how React Apps are built?

ui/browser/reducers/main-reducers.js
```diff
@@ -52,6 +53,9 @@ export default function basic(state = initialState, action) {
     case types.SET_PAGE_ORDER:
       return setPageOrder(state, action.pageOrder);
 
+    case types.TOGGLE_PAGES_SECTION:
+      return togglePagesSection(state, action.opened);
```
jsantell: I really don't like the past tense "opened" here -- it's unclear what the future state will be, as the actions cause a change in state via reducer, and seems to imply it's referring to the current state, and the 'toggle' reference could imply the inverse. Either way, I think as a whole, this action/state should be closer to `SET_PAGES_VIEW_VISIBILITY` or something, where it doesn't reference a previous state (which 'toggle' and a past tense verb seem to imply).

ui/browser/reducers/main-reducers.js
```diff
@@ -135,3 +142,10 @@ function setCurrentTab(state, pageIndex) {
 function setPageOrder(state, pageOrder) {
   return state.set('pageOrder', pageOrder);
 }
+
+function togglePagesSection(state, opened) {
```
jsantell: I don't think this should handle toggling at all -- either take true or false, and make the state change. Yes we could get states where it's already open, and we try to open it again, but that atleast makes the recorded ACTION more obvious

ui/browser/actions/main-actions.js
```diff
@@ -42,3 +42,7 @@ export function setCurrentTab(pageIndex) {
 export function setPageOrder(pageOrder) {
   return { type: types.SET_PAGE_ORDER, pageOrder };
 }
+
+export function togglePagesSection(opened) {
+  return { type: types.TOGGLE_PAGES_SECTION, opened };
```
ncalexan: I feel like you're straddling two action models.  On the one hand, you want to say "toggled"; on the other, you want to say "set to ...".  You have some code checking for `undefined` later to handle the former, which feels wrong to me.

Is there value supporting both verbs?   With one action?  In general, I prefer to have actions that are facts: "opened pages section", "closed pages section", and therefore don't rely on previous state to interpret.  Is that not a Redux-y thing to do?

ui/browser/views/navbar/btn.jsx
```diff
@@ -20,36 +20,42 @@ const noop = function() {};
...
+};
 
 Btn.propTypes = {
   disabled: PropTypes.bool,
+  active: PropTypes.bool,
   id: PropTypes.string,
   image: PropTypes.string,
   title: PropTypes.string.isRequired,
+  icon: PropTypes.string,
   clickHandler: PropTypes.func.isRequired,
```
jsantell: Need propTypes for `children`

ui/browser/reducers/main-reducers.js
jsantell: And in the case where we noop because we're attempting to open an already opened pages tab, then nothing will happen in our views as state didn't really change

ui/browser/views/tabbar/tabbar.jsx
```diff
@@ -20,9 +20,14 @@ import { TabBarDragDrop, TabDragDrop } from './dragdrop';
  * The TabBar runs across the top of each browser window, containing the
  * window management buttons, and browser tabs
  */
-const TabBar = ({ pageOrder, pages, currentPageIndex, dispatch }) => {
+const TabBar = ({ pageOrder, pages, currentPageIndex, dispatch, pageSectionOpened }) => {
```
jsantell: Add `pageSectionOpened` to propTypes

ui/browser/views/navbar/navbar.jsx
bgrins: >  this list of props seems untenable over time

agreed, if you look at diff there's copy/paste required across lots of files to add an action like this.  We should discuss further to see if this is how we want to continue as more actions get added

jsantell:
r+ with the toggling name changes!

PR #96, opened 2016-04-12
Fix CmdOrCtrlW close tab message not hitting the listener. Fixes #92

### LINKED ISSUE #92: Fix CmdOrCtrlW close tab message not hitting the listener
webContents.send() sends a message to the renderer process, and thus we actually need a ipcRenderer to listen on the rederer process.

gabrielluong:
Fixes #92.

r? @jsantell


jsantell:
R+

PR #98, opened 2016-04-12
[WIP] Prevent transpiling node modules. Fixes #58

### LINKED ISSUE #58: Should prevent transpiling of node modules
Maybe we can do this via eslint.

Since we transpile our code before packaging, these values get turned into strings of the current machine, not the machine where a distributable is actually run. Most of the time it would be ok (like the `os.platform()` value), but there may be other discrepencies. Either way, it's a bit confusing.

If you need to query similar info, we should do it at build time -- look at `build/config-builder.js` at how the `build-config.js` is created during build time to have build-specific configurations, not runtime specific.

I'm sure I could be missing some nuisances, but this seems to make sense. How do others feel about this? Is this nothing to worry about?

linclark:
Just putting this up for discussion. Will add notes to #58

### DIFF
--- a/build/webpack.config.default.js
+++ b/build/webpack.config.default.js
@@ -5,6 +5,11 @@
 
 const path = require('path');
 
+const ExternalsPlugin = require('../node_modules/webpack/lib/ExternalsPlugin');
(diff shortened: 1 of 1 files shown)


linclark:
Webpack was removed (#157) so this PR is no longer relevant

PR #99, opened 2016-04-13
Support toggling between 'title' and 'url bar'

jsantell:
This adds the toggling between 'title' and 'url bar' as shown in UX mockups

r? @bgrins @victorporof


ui/browser/views/navbar/btn.jsx
```diff
@@ -12,25 +12,26 @@ specific language governing permissions and limitations under the License.
...
     cursor: (disabled ? 'default' : 'pointer'),
-    margin: '8px 3px 0px 3px',
+    margin: '10px 6px 0px 6px',
     height: HEIGHT,
     display: 'inline-block',
     padding: '0',
     border: '0',
     backgroundColor: (active ? '#DFE1E4' : 'transparent'),
-  };
+  }, style || Object.create(null));
```
victorporof: Instead of object assign, simply use the spread operator:

const styles = {
  foo: bar,
  ...style
}

ain't that nice?

ui/browser/views/navbar/location.jsx
```diff
@@ -20,15 +20,52 @@ import { fixURL, getCurrentWebView } from '../../browser-util';
  * to ipcRenderer on mount, and because the key-down event handler is non
  * trivial.
  */
-class Location extends Component {
-  constructor() {
-    super();
-    this.handleKeyDown = this.handleKeyDown.bind(this);
-  }
+const Location = createClass({
```
victorporof: Can't we use a proper class here?

ui/browser/views/navbar/location.jsx
```diff
@@ -20,15 +20,52 @@ import { fixURL, getCurrentWebView } from '../../browser-util';
...
   componentDidMount() {
     this.props.ipcRenderer.on('focus-urlbar', () => this.refs.input.select());
-  }
+  },
+
+  componentDidUpdate() {
+    // If we're showing the URL bar, it should be focused. The scenario
+    // where this isn't true is immediately after displaying the URL bar,
+    // so give it focus
+    if (this.state.showURLBar && document.activeElement !== this.refs.input) {
```
victorporof: Ugh, checking globals here. Maybe we should have `activeElement` part of the global state.

ui/browser/views/navbar/location.jsx
```diff
@@ -20,15 +20,52 @@ import { fixURL, getCurrentWebView } from '../../browser-util';
...
+
+  onTitleFocus() {
+    this.setState({ showURLBar: true });
+  },
+
+  onURLBarFocus() {
+    this.refs.input.select();
+  },
+
+  onURLBarBlur() {
```
victorporof: The convention seems to have event listeners in react be named handleFooBar instead of onFooBar.

victorporof:
r+ with comments addressed.

ui/browser/views/navbar/btn.jsx
jsantell: works for me!

ui/browser/views/navbar/location.jsx
jsantell: Extending the React.Component class resulted in many warnings from React, it seems not all methods are supported there (getDefaultState() was the one that started it, IIRC)

ui/browser/views/navbar/location.jsx
jsantell: Yeah, I don't like this either. Or if there's way to determine if an element has focus without the document

ui/browser/views/navbar/location.jsx
victorporof: React docs only mention getDefaultState. Need to put those in the class constructor, like this:

class Foo extends React.Component {
  constructor(props) {
    super(props);
    this.state = { ... };
  }
}

Since this is all documented and already a standard among the react croud, let's go with that.

ui/browser/views/navbar/location.jsx
victorporof: Have the App component put this stuff in the redux store, then propagate it down to all components.

victorporof:
Waiting for comments to be addressed, will merge after that.

ui/browser/views/navbar/location.jsx
jsantell: Yuck, there's gotta be a better way. I'll handle this in a follow up

jsantell:
Comments addressed, some finicky changes due to Location always rerendering, and persisting URLs when it shouldn't, and the opposite when it shouldn't in edge cases. This is complex, doing in a follow up.

victorporof:
+1

PR #105, opened 2016-04-13
Rename npm script 'start-dev' task to 'dev'. Fixes #104. r=jsantell

### LINKED ISSUE #104: Rename npm script 'start-dev' task to 'dev'
I think the original intentions was that the npm scripts was gonna be called 'npm start-dev', but in reality, you need to use 'npm run' to run scripts. Instead of calling 'npm run start-dev', let's simplify to 'npm run dev'

gabrielluong:
#104 

r? @jsantell

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ npm install
 ### Tasks
 
 * `npm start` - Runs the build. Builds the client/renderer code via webpack in `production` mode. Should be rather similar to a packaged build.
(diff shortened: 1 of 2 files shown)


victorporof:
Original intention was just dropping gulp, so these can be renamed to anything.

victorporof:
I'm just gonna land this, this is fine.

PR #107, opened 2016-04-13
Allow user to close a tab by middle mouse button

locosoft1986:
Close#106
Allow user to close a tab by middle mouse button.
Resolve the lint problem in the location.jsx, the line 70 is too long.

Signed-off-by: Jeremy Weasley <locosoft1986@gmail.com>

### DIFF
--- a/ui/browser/views/navbar/location.jsx
+++ b/ui/browser/views/navbar/location.jsx
@@ -67,7 +67,8 @@ class Location extends Component {
           onKeyDown={this.handleKeyDown} />
 
         <Btn title="Bookmark"
-          image={page.isBookmarked ? 'glyph-bookmark-filled-16.svg' : 'glyph-bookmark-hollow-16.svg'}
+          image={page.isBookmarked ? 'glyph-bookmark-filled-16.svg'
+                                        : 'glyph-bookmark-hollow-16.svg'}
           clickHandler={onBookmark} />
       </div>
     );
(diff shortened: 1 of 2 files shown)


victorporof:
I had no idea there was such a thing as closing tabs with middle click. Ping @phlsa for his thoughts on this.

The eslint error was fixed by https://github.com/mozilla/tofino/commit/27fa39ff504c37c39bc1a852dea31d4aa7dac217 Generally it's a good idea to separate such changes into different commits, so please keep that in mind in the future.

locosoft1986:
@victorporof  Got it.
@phlsa Hi, what do you think on this? It is convenient for user to click the wheel button of the mouse to close a tab, since Chrome, Firefox and Vivaldi can do so.

victorporof:
This commit is competing with https://github.com/mozilla/tofino/pull/109

phlsa:
I'm generally all for including functionality like this, as it provides consistency and is known to make people's lives easier!

I do have some concerns about timing and size of the code base though. If we are going to try very different approaches to very different parts of the browser, that will mean lots of changing code. The more code we have in the code base, the more potentially needs to change.

I don't want to limit our ability to move fast, so perhaps it's better to pool things like this for now until the code base has somewhat stabilized?
I'm not an engineer and this seems like a pretty small patch so it's probably fine, but I wanted to discuss this issue in general. I'll leave it up to @victorporof whether it is an issue in this regard.

victorporof:
I'm duping this to https://github.com/mozilla/tofino/pull/109 since that has a more up to date PR. Sorry @locosoft1986!

PR #111, opened 2016-04-13
Extend .eslintrc with recommended react rules

victorporof:
Since the react-eslint plugin already extends a "good practices" recommended configuration, we should extend that in our .eslintrc. 

Additionally, we seem to already be following a pattern with our current components. It'd be a good idea to make the current style a standard we keep in the future. There are many rules we could add (see https://github.com/yannickcr/eslint-plugin-react), but I just picked a few:

* react/jsx-indent -- make sure we indent jsx markup with 2 spaces (already the case everywhere)
* react/jsx-indent-props -- make sure we indent every "attribute" (prop) with 2 spaces (already the case almost everywhere)
* react/jsx-max-props-per-line -- make sure we have one "attribute" (prop) on every line (already the case almost everywhere)
* react/jsx-equals-spacing and react/jsx-curly-spacing -- make sure we have prop={value} without spaces (already the case everywhere)
* react/jsx-handler-names -- make sure we adhere to the onFoo={this.handleFoo} de-facto convention for naming event handlers

Ping @joewalker for review.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "airbnb",
+  "extends": ["airbnb", "plugin:react/recommended"],
 
   "parser": "babel-eslint",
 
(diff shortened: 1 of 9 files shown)


joewalker:
LGTM

victorporof:
Thanks!

PR #121, opened 2016-04-14
Take a dependency on sqlite, stubbing out a storage module.

rnewman:
I also fixed a lint warning in `main.js`, and stubbed out usage so we'll die hard if loading the module fails. That part'll get overwritten quite quickly, but until then we have `foo.db`.

You'll need to run `npm install` again to pull in `sqlite3`.

### DIFF
--- a/app/services/storage.js
+++ b/app/services/storage.js
@@ -10,21 +10,27 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
+/* eslint no-console: 0 */
+
 /**
- * Disable the eslint "strict" and "no-commonjs" rules, since this is a script
- * that runs in node without any babel support, so it's not not a module
- * implicitly running in strict mode, and we can't use the import syntax.
+ * The storage module that makes up the majority of the profile service.
+ *
+ * @module Storage
  */
 
-/* eslint-disable strict */
-/* eslint-disable import/no-commonjs */
+import sqlite3 from 'sqlite3';
 
-'use strict';
+export function open(path) {
+  sqlite3.verbose();
 
-// Must go before any require statements.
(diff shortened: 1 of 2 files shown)


app/services/storage.js
```diff
@@ -0,0 +1,34 @@
...
+
+/* eslint no-console: 0 */
+
+/**
+ * The storage module that makes up the majority of the profile service.
+ *
+ * @module Storage
+ */
+
+// const { app } = require('electron');
```
jsantell: extra commented line

package.json
```diff
@@ -29,7 +29,8 @@
     "react-dom": "0.14.7",
     "react-redux": "4.4.1",
     "redux": "3.3.1",
-    "redux-logger": "2.6.1"
+    "redux-logger": "2.6.1",
+    "sqlite3": "^3.1.3"
```
jsantell: Other than one exception, all other modules have versions explicitly specified, any reason not to do that here?

app/main/main.js
```diff
@@ -130,6 +138,7 @@ app.on('ready', () => {
   });
   registerFileProtocol('libs', url => path.normalize(path.join(root, 'libs', url)));
 
+  storage.open('/tmp/foo.db');
```
jsantell: Probably should use path module here so it works in windows; also Electron has an [API](http://electron.atom.io/docs/latest/api/app/#appgetpathname) for getting temp directories in a cross platform way that we could use here

victorporof:
@rnewman you'll have to rebase since https://github.com/mozilla/tofino/pull/125 landed. The new main.js code is in a sibling file now, to allow webpacking. See the PR.

rnewman:
Rebased, feedback addressed; rubberstamp plz, @jsantell.

PR #125, opened 2016-04-14
Use native modules everywhere, r=jwalker. Fixes #124

### LINKED ISSUE #124: Use native modules everywhere
Currently we have a mix of require and imports. We can use native modules everywhere except in main.js for the build system and main.js in app.

victorporof:
https://github.com/mozilla/tofino/issues/124

Individual commits are easier to follow.


victorporof:
Ping @joewalker

package.json
```diff
@@ -39,6 +39,7 @@
     "enzyme": "2.2.0",
     "eslint": "2.4.0",
     "eslint-config-airbnb": "6.2.0",
+    "eslint-plugin-import": "^1.4.0",
```
joewalker: I think we should use `"1.4.0"` to avoid different people using different versions of the module.

victorporof: Agreed.

build/tasks.js
```diff
@@ -1,41 +1,46 @@
...
+ * Use need to allow lazy loading of modules for all tasks. Generally,
+ * we assume `build-config.js` exists, however *during* the build process
+ * this might not be true, which will break certain native module imports that
+ * try to eagerly load this file.
+ */
+
+export default {
   async config(options) {
-    await require('./task-config-builder')(options);
+    await require('./task-config-builder').default(options);
```
joewalker: I'm surprised this doesn't need an `/* eslint-disable import/no-commonjs */`. What did I miss?

victorporof: Inline requires are still allowed, since the `import` eslint plugin assumes they're function calls. There's a few more occurrences of this in app/ code. Not sure what I could do about this, except disallow naming functions "require", or not assuming a node environment anymore.

joewalker:
LGTM.

victorporof:
Thanks Joe!

victorporof:
I made most of the code in app/main/main.js go through babel as well, so that we can use proper module syntax there too. So currently we only still use require wherever we need lazy/hot loading of modules, and all the places are documented.

Issue #20, opened 2016-04-06
Expose webpack failures when running in dev mode


jsantell:
Right now, if there's a build failure when running `gulp run-dev`, we just get a useless 404 error for `static/built/browser.js`.

When running in `gulp run`, we can open the tools and see "Uncaught Error: Cannot find module "./navbar/navbar.jsx". We need to display this information somewhere.

jsantell:
The new build scripts don't report any errors if webpack build fails, and just fails silently.

jsantell:
Didn't mean to close

victorporof:
Could it be because we're using promises for build tasks now, and not catching the errors anywhere? Currently the promises should reject for everything wrong thay happens, but those rejections get swallowed. I'll tale this.

PR #126, opened 2016-04-14
Expose build failures (#20)

victorporof:

Indeed existing promise rejections weren't handled in the top level build code, and I found a few other call sites that didn't properly reject or throw when errors occurred.

Ping @jsantell for review.


jsantell:
:shipit:

PR #129, opened 2016-04-15
Test actions effects on state

jsantell:
Tests for action creators mostly, also removing `pageOrder` state. This completely disables the tab dragdrop, but it already wasn't working and we're not sure what that UX will even be.

Next steps are cleaning up the actions and make enforce the action flow (tracked in other issues)

### DIFF
--- a/test/ui/browser/actions/test-duplicate-tab.js
+++ b/test/ui/browser/actions/test-duplicate-tab.js
@@ -1,0 +1,28 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+import expect from 'expect';
+import configureStore from '../../../../ui/browser/store/store';
+import * as actions from '../../../../ui/browser/actions/main-actions';
+
+describe('Action - DUPLICATE_TAB', () => {
+  beforeEach(function() {
+    this.store = configureStore();
+  });
+
+  it('Should create a new tab with source\'s location and select it', function() {
+    const { getState, dispatch } = this.store;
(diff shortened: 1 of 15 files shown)


test/ui/browser/actions/test-duplicate-tab.js
```diff
@@ -0,0 +1,28 @@
...
+    expect(getState().currentPageIndex).toEqual(0);
+
+    dispatch(actions.createTab('https://moz.org/1'));
+    dispatch(actions.createTab('https://moz.org/2'));
+    dispatch(actions.createTab('https://moz.org/3'));
+
+    dispatch(actions.duplicateTab(1));
+    expect(getState().pages.size).toEqual(5);
+    expect(getState().pages.get(4).location).toEqual('https://moz.org/1',
+      'DUPLICATE_TAB creates an appends a new tab with source\'s location');
```
bgrins: Do we have to use single quotes even when there's a single quote in the string or is eslint smart enough to not complain about this?

bgrins:
The tests look good to me

test/ui/browser/actions/test-duplicate-tab.js
jsantell: Hmm not sure, but not sure if it's preferable to have double quotes instead of an escape in this case

victorporof:
shipit!

victorporof:
Your branch had a merge commit in it so I had to squash to maintain a clean history. In the future, please rebase your dev branches instead of merging upstream master into them. See the readme on contributing.

PR #137, opened 2016-04-15
Fix SET_PAGE_ORDER and write tests ensuring selected pages remain con…

jsantell:
…sistent.

r? anyone

### DIFF
--- a/test/ui/browser/actions/test-set-page-order.js
+++ b/test/ui/browser/actions/test-set-page-order.js
@@ -1,14 +1,94 @@
 // Any copyright is dedicated to the Public Domain.
 // http://creativecommons.org/publicdomain/zero/1.0/
 
-import expect from 'expect'; // eslint-disable-line
+import expect from 'expect';
 import configureStore from '../../../../ui/browser/store/store';
-import * as actions from '../../../../ui/browser/actions/main-actions'; // eslint-disable-line
+import * as actions from '../../../../ui/browser/actions/main-actions';
+
+/**
+ * Takes a pages List and maps the location ('http://moz1.org')
+ * to just it's number, '1', and into a string for testing ordering here.
+ */
+function serialize(list) {
+  return list.map(p => /moz(\d)\.org/.exec(p.location)[1]).join(',');
+}
+
+function getSelectedPageURL(state) {
+  return state.pages.get(state.currentPageIndex).location;
+}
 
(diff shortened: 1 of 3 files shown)


ui/browser/reducers/main-reducers.js
```diff
@@ -125,10 +125,28 @@ function setCurrentTab(state, pageIndex) {
...
+      currentPageIndex--;
+    }
+    if (updatedIndex <= currentPageIndex) {
+      currentPageIndex++;
+    }
+  }
+
+  // First remove the page we want to move, then add back to the array,
+  // maybe we want to do this in a faster way?
+  let pages = state.pages.splice(pageIndex, 1);
```
ncalexan: Here's where we want to use `withMutations` and save some garbage collection.

test/ui/browser/actions/test-set-page-order.js
```diff
@@ -1,14 +1,90 @@
...
+
+  it('Should still reference currently selected page when moving page before', function() {
+    const { dispatch, getState } = this.store;
+
+    dispatch(actions.setCurrentTab(2));
+    dispatch(actions.setPageOrder(4, 0));
+    expect(serialize(getState().pages)).toEqual([
+      '5', '1', '2', '3', '4',
+    ].toString());
+    expect(getState().currentPageIndex).toEqual(3,
```
ncalexan: It would make it easier to reason if you also assert the current tab's title.

test/ui/browser/actions/test-set-page-order.js
```diff
@@ -1,14 +1,90 @@
 // Any copyright is dedicated to the Public Domain.
 // http://creativecommons.org/publicdomain/zero/1.0/
 
-import expect from 'expect'; // eslint-disable-line
+import expect from 'expect';
 import configureStore from '../../../../ui/browser/store/store';
-import * as actions from '../../../../ui/browser/actions/main-actions'; // eslint-disable-line
+import * as actions from '../../../../ui/browser/actions/main-actions';
+
+/**
```
ncalexan: This might be simpler if you used tab titles, rather than extracting things from the URLs.

ncalexan:
I made some suggestions, but they don't need to be addressed if you choose otherwise.

test/ui/browser/actions/test-set-page-order.js
jsantell: :+1:

PR #142, opened 2016-04-16
Transpile all the main process scripts and drop hot-module-loading.

### LINKED ISSUE #14: Transpile main process code via webpack
Right now we're doing this in runtime. We need to transpile this code (put it in `app-built/app.js`? no idea) and then move all the babel deps to devDependency so we don't put it in the distributed package.

[I'm definitely not working on this]

### LINKED ISSUE #40: Make hotload not bring App to the foreground
I hit save in Emacs, a lot.  I manage my git tree in Emacs, a lot.  We hot reload the App front-end even when it's not the foreground App; and that brings the window to the front.  Can we not do that?

### LINKED ISSUE #61: On first startup Windows Firewall throws up a dialog asking whether to allow electron to do things
I'm guessing this is the webserver we're starting for hot-module reloading. Maybe if we only listen on localhost it won't throw up this?

### LINKED ISSUE #134: Remove code handling hot module reload
There doesn't seem to be a compelling need for this at this time (we have ctrl+alt+r for reloading the window anyway), and nobody seems to know exactly how it works.

Mossop:
This uses babel to transpile all of the main process scripts and also removes
all of the hot-module support and the web server that it required. All generated
scripts go into the `lib` directory and the app runs from there using file links
which will make using preload scripts much easier.

For the main process scripts we only transpile if there has been a change. The
copy function could probably be faster but there isn't one built into node O_o.

fixes #14, fixes #40, fixes #61, fixes #134.


package.json
```diff
@@ -50,6 +48,7 @@
     "mocha": "2.4.5",
     "mocha-eslint": "2.0.2",
     "mocha-jsdom": "1.1.0",
+    "mz": "2.4.0",
```
jsantell: Any advantages of `mz` over `fs-promise`? They both look popular, just familiar with fs-promise

Mossop: Mainly it's what I was used to, mz seemed nice as a one-stop-shop for all node APIs. But it looks like fs-promise supports using fs-extra to make a proper copy function so maybe I'll switch to that.

build/task-build.js
```diff
@@ -68,12 +69,13 @@ async function babelDir(source, target) {
 
     if (sourceFile.endsWith('.js')) {
       const { code } = await transpile(sourceFile, { sourceMaps: true });
-      await fs.writeFile(targetFile, code);
+      promises.push(fs.writeFile(targetFile, code));
     } else {
-      const buffer = await fs.readFile(sourceFile);
-      await fs.writeFile(targetFile, buffer);
+      promises.push(fs.copy(sourceFile, targetFile));
```
victorporof: Why not 'await' on all of these? Don't we care about order?

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "description": "Experimental prototype",
   "main": "lib/main/main.js",
+  "repository": "https://github.com/mozilla/tofino",
```
victorporof: Would be better to have this in a separate commot but whatever.

app/main/browser.js
```diff
@@ -1,167 +0,0 @@
-/*
```
victorporof: We're losing a lot of history by doing this. Can't this be renamed in a commit then changed in another, if git can't figure out the move?

victorporof:
Yeah so mainly I don't understand how jsx files are treated; if we're still using babel-register to load those, that's sucky. Let's transpile evetything. 
Everything else looks good.

victorporof:
Might also have to update the "import/no-unresolved" rule in .eslintrc

victorporof:
Should we still package dev dependencies now? See task-package.js. I'm not really sure why we were doing that in the first place.

jsantell:
@victorporof  `devDependencies` are explicitly ignored in `task-package.js`

victorporof:
@jsantell Oh right! I had that the other way around. Carry on :)

jsantell:
I'm not sure if `lib` is a good name for all the built things, but have been trying to think of a better name for literally over a week now at this point so :yolo:

Mossop:
> I'm not sure if lib is a good name for all the built things, but have been trying to think of a better name for literally over a week now at this point so :yolo:

That was basically where I got to too.

build/task-build.js
Mossop: Nope, the files can be processed in whatever order, might as well try to do it in parallel where possible

Mossop:
@victorporof I think this is all up to date from your comments. Look good to you?

victorporof:
r++

Mossop:
Landed in 15d6b658a61d771d83ccdc2df28af5d3b7e8f6ad

Issue #147, opened 2016-04-17
Class names are not passed to webviews


victorporof:
...so all code that relies on `webview.webview-{index}` selectors (e.g. tab ordering, drag & drop etc.) doesn't work.

victorporof:
Turns out drag'n'drop doesn't work for other reasons (it's a WIP), but the class names still aren't properly passed in, so we should fix that.

Issue #113, opened 2016-04-13
Consolidate the way we write CSS


victorporof:
There have been ongoing discussions the last week about how we're going to go about writing CSS. Currently things are ok, but they are inconsistent and suffer from the age-old problem of CSS: cascading, improper modularity, lack of namespacing etc.

Inline styles make things way better, but they suffer from a few problems. The more objective ones are:
* pseudo-selectors are clunky, :hover, :visited, etc. are hard to use, unless using event handlers
* shadow dom is almost impossible to style (e.g. browser-defined insides of some nodes)
* hey make the DOM bigger and hard to inspect
* theming information would require being part of application state.

The general conclusion was that a mix of css files and inlined styles would be nice. React-free-style could help us with that: https://github.com/blakeembrey/react-free-style. Alternatively we could also write our own thing.

This issue is about unifying the way we do things.

victorporof:
Also relevant to leave here: https://github.com/blakeembrey/free-style#why

victorporof:
Fixed by https://github.com/mozilla/tofino/pull/163

PR #148, opened 2016-04-17
Class names are not passed to webviews

victorporof:
https://github.com/mozilla/tofino/issues/147


victorporof:
@jsantell take a quick look at this please. Class names are now properly passed into the `webview` element in the shadow dom. I'll fix the "I have no idea why we need this" comment in https://github.com/mozilla/tofino/issues/113, it's related to this problem.

jsantell:
Wow, good eye. I guess we don't have any ways of testing WebViews yet?

r+

PR #150, opened 2016-04-17
Navigation commands for WebView should be handled by Redux actions vi…

### LINKED ISSUE #119: Navigation should be an action in navbar, Location
Right now, we just report every location change as an action, and no action for actually navigating.

We should make navigating an action, and optionally the incremental typing in the location bar may send info to the main process for frecency, etc.

jsantell:
…a state changes observed by Page. r=vp,linclark

Fixes #119.

This removes most of the usage of different components querying the DOM for the webview, and so only the Page object manages its WebView*. There are still other things grabbing the webview with `getCurrentWebView` but these are for non-navigation components like copy/paste, may need a different solution to clean that up

This feels more Reduxy, and commands are executed as the most recent 'command' (refresh, back, forward, etc) is stored in the Page state, so the Page knows if it needs to fire off a command or not. Is this the right way to do this? Feels nice, having these pretty big deal actions actually reflected in the redux state, though.

Can't assign to @linclark for some reason (?), but would like both of ya @victorporof to take a look at this, because I think this problem will come up again so good to have a pattern for dealing with it

### DIFF
--- a/ui/browser/actions/external.js
+++ b/ui/browser/actions/external.js
@@ -10,7 +10,9 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
-import { createTab, setPageDetails, duplicateTab, closeTab } from './main-actions';
+import {
+  createTab, setPageDetails, duplicateTab, closeTab, navigatePageTo,
+} from './main-actions';
 import { getCurrentWebView, fixURL } from '../browser-util';
 import { remote, clipboard, ipcRenderer } from 'electron';
 
@@ -84,14 +86,12 @@ export function menuLocationContext(input, dispatch) {
 
   menu.append(new MenuItem({
     label: 'Paste and Go',
-    click: ev => {
+    click: () => {
       const before = input.value.slice(0, input.selectionStart);
       const after = input.value.slice(input.selectionEnd);
       const loc = before + clipboard.readText() + after;
       dispatch(setPageDetails(-1, { userTyped: loc }));
-
-      const webview = getCurrentWebView(ev.target.ownerDocument);
-      webview.setAttribute('src', fixURL(loc));
+      dispatch(navigatePageTo(-1, fixURL(loc)));
     },
   }));
 
--- a/ui/browser/browser-util.js
+++ b/ui/browser/browser-util.js
@@ -71,3 +71,13 @@ export function getWebView(document, pageIndex) {
   }
   return element.webview;
 }
+
+/**
+ * Takes an object and attaches a unique `id` to it.
+ *
+ * @TODO Should we start using UUIDs?
+ */
+let ATTACH_UNIQUE_ID = 0;
+export function attachUnique(obj) {
+  return Object.assign(obj, { id: ATTACH_UNIQUE_ID++ });
+}
--- a/ui/browser/views/browser.jsx
+++ b/ui/browser/views/browser.jsx
@@ -24,12 +24,12 @@ import {
 import {
   setPageAreaVisibility as setPageAreaVisibilityAction,
   createTab, attachTab, closeTab, setPageDetails,
+  navigatePageBack, navigatePageForward, navigatePageRefresh, navigatePageTo,
 } from '../actions/main-actions';
 
 import * as mainActions from '../actions/main-actions';
 import * as profileCommands from '../../../app/shared/profile-commands';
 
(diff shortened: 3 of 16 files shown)


jsantell:
That coverage diff is awesome.

ui/browser/actions/external.js
```diff
@@ -11,7 +11,9 @@ specific language governing permissions and limitations under the License.
 */
 
 import browserDB from '../../shared/browser-db';
-import { createTab, setPageDetails, duplicateTab, closeTab } from './main-actions';
+import {
+  createTab, setPageDetails, duplicateTab, closeTab, navigatePageTo,
+} from './main-actions';
```
victorporof: We really need to stop abusing destructuring when importing or handling `props`. Not a problem here, big problem in other places.

ui/browser/views/browser.jsx
```diff
@@ -63,7 +63,7 @@ class BrowserWindow extends Component {
           {...{ pages, navBack, navForward, navRefresh, minimize, maximize,
             close, openMenu, onLocationChange, onLocationContextMenu,
             onLocationReset, bookmark, unbookmark, pageAreaVisible, ipcRenderer,
-            setPageAreaVisibility }} />
+            setPageAreaVisibility, navigateTo }} />
```
victorporof: This is getting more and more horrible. Why not have some kind of object coalescing all these props instead of passing them one by one?

ui/browser/browser-util.js
```diff
@@ -71,3 +71,13 @@ export function getWebView(document, pageIndex) {
   }
   return element.webview;
 }
+
+/**
+ * Takes an object and attaches a unique `id` to it.
+ *
+ * @TODO Should we start using UUIDs?
```
victorporof: Incremental ids are unique anyway, as long as no other part of the application has a different generator. Could pull in some UUID library though, especially if we're saving things to databases where we truly want unique ids, there's tons on npm and supposedly pretty fast.

victorporof:
This is so hot.

ui/browser/actions/external.js
jsantell: Was talking to @bgrins about this; I think the best solution to this is just pass in the entire props object to the children:

```
<ComponentX { ...this.props } />
<ComponentY { ...this.props } />
```

Yeah a lot of the properties will be discarded, but I don't think that'll be an issue.

linclark:
I think the state might need to be slightly restructured because of React's batching.

What would we want to have happen if multiple state changes to the page were happening in the same batch and each was trying to execute a different command? Would we want each one of the commands to be executed, or just the last one? With the current code, I think just the last one might be executed.

After a component is updated, its `componentDidUpdate` method is [added to a queue](https://github.com/facebook/react/blob/ae569105738de9bd732d121b0098e0a2ed01baa5/src/renderers/shared/reconciler/ReactCompositeComponent.js#L824) (rather than invoked directly). Because of this, I believe there is no guarantee that each one of the states we set will reach `componentDidUpdate`.

If we restructured the state, I think it would be safer.

```js
// Pages is a map with incremental IDs for keys
pages: {
  15: { ... },
  16: { ... },
  22: { ... },
},
// Commands is a map using the same keys as pages does
commands: {
  15: [
    'back',
    'back',
    'forward'
  ],
  16: []
  22: []
}
```

In the `connect()` function for the page, we would get both the page and the corresponding commands array. Then on `componentDidUpdate`, we could check the length of the commands array against the previous length and run any commands that had been added to the array since the last pass through.

jsantell:
@linclark Ah, good to know about the batching! That certainly affects things here. I'll update the PR to store a command queue per page to see how that will look -- adding it to the root state object might make more sense if/when we add more commands? But at the same time, these commands are specifically to manipulate the webview... we'll see how this updated PR looks

jsantell:
So unless we drain the queue, this command list will keep growing unless the Page also fires a command indicating to clear out command queue (which seems like a wasteful action) -- I think that was the original reason of having the unique ID on the command to ensure the command wasn't already executed. Is there a good way to handle this?

linclark:
This is an idea. I don't particularly like it, but just to put it out there...

We could store the ID of the last executed command as state on the Page. Then, we could pass that in with future actions and prune anything before it.

I'm thinking it might be premature to optimize that way at this point, though, for two reasons.

1. I'm thinking webview might become more React friendly once the GitHub teams are working with React more.
2. Long term, we might want to consider integrating something else (like Relay) which might change the way we handle actions

Issue #153, opened 2016-04-18
Consider not using webpack at all, ot at least being consistent about what we webpack and what we don't


victorporof:
Now that we're not gulping (https://github.com/mozilla/tofino/issues/76), and as of today babeling everything (https://github.com/mozilla/tofino/pull/142), we should consider not using webpack anymore. AFAICT, currently the only benefit of webpack is that we're able to use json loaders and import directly from json files. Since we're probably not going to rely on any other loaders [js and jsx are handled by babel, css is handled by react-free-style (https://github.com/mozilla/tofino/issues/113)], webpack has no real benefit.

victorporof:
@jsantell @Mossop what do you think?

Mossop:
Node already supports loading from json directly IIRC. My understanding is that the only thing webpack gives us at this point is packing all the module dependencies into one file. I don't know what that buys us. Maybe some perf? Probably only the same as if we used electron's asar thing but it complicated things.

I have heard that webpack has some support for dead code removal so maybe that might be useful, but at this early stage I don't think it is worth doing work for file size optimizations.

victorporof:
At the very least we should be consistent. We're webpacking the browser ui code, but not the main process code.

victorporof:
@Mossop Can we use require for files loaded in html documents? Currently we use webpack to bundle everything into `lib/ui/browser.js`, which is included as a script in `static/browser.html`, which is loaded by the main process code in `app/main/browser.js`. If we don't use webpack, do you know if modules would just work? Maybe we can't get rid of webpack after all.

Mossop:
Code loaded in BrowserWindows can use require and other node stuff just fine (unless you turn off nodeIntegration). Electron creates a node context and then appears to copy the `require`, `process` and other bits into the global for the html document.

victorporof:
Nice!

PR #157, opened 2016-04-18
Kill webpack.

Mossop:
@victorporof this is what getting rid of webpack roughly looks like. It fails eslint tests but it runs and works. It moves a lot of stuff to make it work so I'm interested in what you think of the final directory structure.


Mossop:
This fixes #153

Mossop:
@ncalexan You might want to look at this too Nick. Now all the main source code is in the `app` directory (maybe we should rename it to `src` at this point). The `README.md` has some notes about what is what there.

Mossop:
So a problem with this is that it then allows us to use node modules inside the renderer UI. It sounds like we don't want that so we might need to stick with webpack.

victorporof:
@Mossop It "allows" us, but we don't have to, right? We can have strict eslint rules that prevent us from accidentally doing that.

victorporof:
Didn't mean to close.

victorporof:
Code looks good, I'd personally like to see this merged.

Mossop:
Yeah I need to get the eslint failures fixed which I think involves writing a new npm module :cry:

victorporof:
I can take over if you don't want to do that 👍 What do you mean by writing a new npm module?

Mossop:
So we transform .jsx -> .js files. To make import statements work in both cases I dropped the ".jsx" part from the import statements. But then the eslint import plugin can't find the .jsx files so it flags a no-unresolved error. Looks like you can extend the plugin with a resolver module to find new files but it looks like no-one has written a resolver for jsx files yet. https://github.com/benmosher/eslint-plugin-import#resolvers

Mossop:
I guess we could just rename the jsx files to js, I don't know if babel cares about the extension. But I bet other tools might

victorporof:
Ugh 😢

linclark:
Not 100% sure, but I've seen `.js` used interchangeably with `.jsx` a lot in the community, so I don't think there are any major tools that depend on the file extension.

Mossop:
Turns out I just needed to read deeper into the docs to find the setting to turn on jsx support!. I'll get this rebased and ready to go

Mossop:
This is updated and should now pass tests

app/ui/browser/reducers/main-reducers.js
```diff
@@ -11,11 +11,12 @@ specific language governing permissions and limitations under the License.
 */
 
 import assert from 'assert';
+import 'babel-polyfill';
```
victorporof: Why do we need this here?

Mossop: This file uses Object.entries which isn't supported in node.

build/task-build.js
```diff
@@ -44,6 +27,9 @@ async function buildFile({ path: sourceFile, stats: sourceStats }, targetFile) {
 
   const extension = path.extname(sourceFile);
   if (extension === '.js' || extension === '.jsx') {
+    if (extension === '.jsx') {
+      targetFile = targetFile.substring(0, targetFile.length - 1);
```
victorporof: This looks a bit terrible. If by any chance we'd like to also support files with the extension .foo at some point, we'd have to substring -3 and add a .js etc.

Extracting the basename and patch and attaching an extension seems like a better way to do this, but I can totally understand why you'd not want to do it now.

app/ui/browser/reducers/main-reducers.js
victorporof: So we'd have to include babel-polyfill in every file that needs a polyfill? That sounds yucky.

victorporof:
Really happy to see this patch. My only real concern is about babel-polyfill.

app/ui/browser/reducers/main-reducers.js
Mossop: Chatted with Jordan and let's just land this for now and we can see if that becomes an issue

jsantell: Yucky, yeah -- but let's land it and deal with that later. Sounds like we'd either need a build step or not use these new methods in this case

victorporof:
This was merged in https://github.com/mozilla/tofino/commit/3ca2ded897e095aecf321d9d477984fa169855ff

PR #159, opened 2016-04-19
[WIP] SQLite promises

rnewman:
Early days. Trying to commit as I go.

### DIFF
--- a/app/services/sqlite.js
+++ b/app/services/sqlite.js
@@ -1,0 +1,68 @@
+/*
+ Copyright 2016 Mozilla
+
+ Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ this file except in compliance with the License. You may obtain a copy of the
+ License at http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software distributed
+ under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+
+/**
+ * When life gives you callbacks, write your own promise wrapper.
+ *
+ * @module SQLite
+ */
+
+import sqlite3 from 'sqlite3';
+import thenifyAll from 'thenify-all';
+
+const Promise = global.Promise;
+
+export function verbose() {
+  sqlite3.verbose();
+}
+
+/**
+ * Turn methods on `source` that return callbacks into methods on `dest` that are bound to
+ * `source` as if invoked as method calls.
+ *
+ * @param source an object with methods named by `methods`.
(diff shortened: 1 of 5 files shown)


app/services/sqlite.js
```diff
@@ -0,0 +1,60 @@
...
+import sqlite3 from 'sqlite3';
+import thenifyAll from 'thenify-all';
+
+const Promise = global.Promise;
+
+export function verbose() {
+  sqlite3.verbose();
+}
+
+function thenifyMethods(source, dest, methods) {
```
ncalexan: Worth a little comment.

app/services/storage.js
```diff
@@ -18,28 +18,49 @@
...
+const mkdirp = thenify(cbmkdirp);
+
+export class Storage {
+  constructor(db) {
+    this.db = db;
+  }
+
+  /**
+   * Open a database and return a promise-based wrapper.
+   * @param path
```
ncalexan: Update these comments, please.

app/services/storage.js
```diff
@@ -18,28 +18,49 @@
  * @module Storage
  */
 
-import * as sqlite from './sqlite';
+import cbmkdirp from 'less-mkdirp';
```
ncalexan: Did we already take this dep?  (I don't see it locally, and it's not in this patch.)

app/services/storage.js
```diff
@@ -18,28 +18,49 @@
...
-        console.log(`Unable to fetch DB user_version: ${err}.`);
-      });
-  });
-
-  return o;
+import { DB, verbose } from './sqlite';
+
+const mkdirp = thenify(cbmkdirp);
+
+export class Storage {
```
ncalexan: Class comment -- how is `Storage` different from ... Why is it not `ProfileStorage` or `EventStore` or ...

app/services/sqlite.js
```diff
@@ -41,18 +41,19 @@ export class DB {
   /**
    *
    * @param filename the file path to open
-   * @param mode any mode flags to pass to `sqlite3.Database`.
+   * @param mode any mode flags to pass to `sqlite3.Database`. Defaults to OPEN_CREATE | OPEN_READWRITE.
    * @returns Promise that resolves to the opened database.
    */
   static open(filename, mode) {
```
ncalexan: JS has default arguments now.

ncalexan:
Couple nits and requests.  Might as well fold this into something more coherent before landing.

app/services/storage.js
rnewman: Yes: https://github.com/mozilla/tofino/pull/159/commits/994d93ae28b7686b7718490a0bfc69cb9e6570b4

app/services/storage.js
rnewman: I think this'll drop out later. It's really `ProfileStorage`, so I'll change the name.

PR #160, opened 2016-04-19
catch instrumentation failures so as to not crash the process if analytics is unreachable.  r=nalexander

bgrins:
Signed-off-by: Brian Grinstead <briangrinstead@gmail.com>


ncalexan:
wfm.

PR #162, opened 2016-04-19
Update electron-packager for security issue. r=mossop

linclark:
See security notice https://github.com/electron-userland/electron-packager/issues/333

Marking this as a WIP because I tried testing this, but when I packaged with either version I got an error about not being able to find "immutable" in bookmarks-reducers.js at line 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "babel-register": "6.7.2",
     "chokidar": "1.4.3",
     "coveralls": "2.11.9",
-    "electron-packager": "5.2.1",
+    "electron-packager": "7.0.0",
     "electron-prebuilt": "0.37.6",
     "electron-rebuild": "1.1.3",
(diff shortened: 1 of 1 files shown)


jsantell:
Trying this out -- got an error with `immutable`.

Turns out, immutable's main is located at `dist/immutable.js`, and we ignore `/dist` in `build/task-package.js` when copying over to the electron app -- I think if we make our ignore regexes more strict here, it could fix the issue

bgrins:
This immutable error looks like it might be happening even without the PR: https://github.com/mozilla/tofino/issues/185#issue-149626425

linclark:
Yes, I tried to say that in the original comment, but it might not have been clear.

I tested it with the current version of electron-packager and the new version. It failed with both. The error is independent of this PR.

Mossop:
LGTM

PR #163, opened 2016-04-19
Consolidate the way we write CSS

victorporof:
https://github.com/mozilla/tofino/issues/113

Ping @jsantell for review. Please take a close look at the CSS. I ended up rewriting all the styles, everything used to be really messy and layout rules were quite dubious. I'll update the readme soon with how we should write CSS.


victorporof:
@bgrins @linclark @joewalker Feel free to take a look at this too if you have opinions about css.

static/css/common.css
```diff
@@ -1,3 +1,7 @@
-.hidden {
+div {
+  display: flex;
```
joewalker: That's going to surprise people that don't know about our divs, right?, or did I misunderstand a scope? `div.flex { ... }`?

victorporof: Indeed this is global. I don't think there's any scenario when we don't want divs to display as flex boxes. Developer intent is probably never "make this a block where children are displayed following dubious layout rules that nobody remembers". Pretty much every time I see a div I end up making it display as flex. But I agree that this might be subjective so I can change this (and end up making all our divs explicitly flexbox anyway).

jsantell:
This looks great. r+, not sure if others have strong opinions on styling, but I like this.

victorporof:
Landing. I'll address additional comments if there's anything else added here after merging.

PR #172, opened 2016-04-19
Use 'value' in the url bar instead of defaultValue? Fixes #170

### LINKED ISSUE #170: Location bar input value doesn't update when switching tabs
STR:

* Open two tabs
* Focus location bar and enter 'foo'
* Switch to the other tab
* Focus location bar

Expected:
The location bar has the URL for the second tab

Actual:
The location bar has the text 'foo'

bgrins:


ui/browser/views/navbar/location.jsx
```diff
@@ -158,11 +158,12 @@ class Location extends Component {
           </span>
         </div>
         <input id="urlbar-input"
-          className={INPUT_BAR_STYLE}
-          hidden={!showURLBar}
+          is="tofino-input"
```
bgrins: It turns out using is="" here isn't necessary, and the fix is due to the change from value to defaultValue

bgrins:
OK, so using 'value' messes up the undo state after all

bgrins:
Closing this for now

PR #173, opened 2016-04-19
No bug - Pass command line arguments through `npm run`.

ncalexan:
My intended use case is to specify the path the browser loads per-window
UI from.

r? @Mossop


Mossop:
LGTM

PR #175, opened 2016-04-19
Use new action for setUserTypedLocation instead of wrapping it into setPageDetails

bgrins:
r? @jsantell


jsantell:
Looks good; I'd prohibit userTyped being modified by setPageDetails, and consider consistency of pageIndex between these APIs (if this is a property, it should also be a prop on setPageDetails, navigate etc)

bgrins:
> I'd prohibit userTyped being modified by setPageDetails

Done

> consider consistency of pageIndex between these APIs (if this is a property, it should also be a prop on setPageDetails, navigate etc)

Agreed - i'll do that in a follow up

PR #176, opened 2016-04-19
Use payload format for setPageDetails and setUserTypedLocation instea…

bgrins:
…d of separate pageIndex. r= @jsantell


jsantell:
r+

PR #183, opened 2016-04-19
Bump appveyor.yml to VS 2015.

rnewman:


Mossop:
We can close this right @rnewman?

PR #184, opened 2016-04-20
Send user typed locations over IPC and return a set of completions.

bgrins:
r= @jsantell 

Includes thunk library to trigger IPC communication inside of the action creator


app/main/reducers/location-reducers.js
```diff
@@ -0,0 +1,31 @@
...
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+
+import * as profileCommandTypes from '../../shared/constants/profile-command-types';
+
+import Immutable from 'immutable';
+
+export default function locationReducer(state, command) {
+  if (typeof state === 'undefined') {
```
jsantell: Why would the state be `'undefined'` as a string?

app/shared/profile-commands.js
```diff
@@ -31,3 +31,8 @@ export function unbookmark(url) {
 export function visited(url) {
   return { type: profileCommandTypes.VISITED, payload: { url } };
 }
+
+export function setUserTypedLocation(text) {
+  return { type: profileCommandTypes.SET_USER_TYPED_LOCATION,
```
jsantell:
```
return {
  type: ...,
  payload: { },
}
```

app/main/reducers/location-reducers.js
jsantell: nevermind, `typeof`, couldn't we just check `state == void 0`?

PR #186, opened 2016-04-20
Remove babel useless imports

victorporof:
I'm not really sure why we need them, and tests seem to run just fine without.

@ncalexan: I remember you seemed to have to add these in a test a while ago, I wonder why.
@Mossop: You did some work lately on removing webpack and using babel for things, and I don't think it was at all related to how we test things. But apparently we don't need these imports.

### DIFF
--- a/test/profile/test-storage.js
+++ b/test/profile/test-storage.js
@@ -3,9 +3,6 @@
 
 /* eslint no-console: 0 */
 
-import 'babel-polyfill';
-import 'babel-register';
-
 import expect from 'expect';
 
 import fs from 'fs';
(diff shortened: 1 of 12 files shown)


jsantell:
We need these if a file uses Object.entries anywhere in code etc.

victorporof:
@jsantell True, but we shouldn't import 'babel-polyfill' otherwise, no? At least 'babel-register' really has no meaning now.

Mossop:
LGTM.

babel-register is already imported by mocha.opts so you shouldn't need to explicitly include that. babel-polyfill should only be needed if you're using some Object or Array static methods that Node doesn't support.

ncalexan:
We were, but `util.js:mapObject` is gone now.

PR #189, opened 2016-04-20
Remove bookmarks dead code. Fixes #188

### LINKED ISSUE #188: Bookmarks content code doesn't seem to be used anywhere
The entire app/content/bookmarks doesn't seem to be loaded anywhere, and it's following the old model with global css files, not using jsx and is pretty much code rot. We should build the UI for these inside the app/ui folder anyway.

victorporof:
https://github.com/mozilla/tofino/issues/188


victorporof:
Only commit fb15dac is relevant, as this branch was based off https://github.com/mozilla/tofino/pull/186

jsantell:
r+ for removal, but not sure who wrote this code; would be a good use case if it works of just displaying static content outside of the browser chrome and profile service if it works

victorporof:
Rebased on master.

PR #192, opened 2016-04-20
Remove blue glow on buttons when selected

victorporof:

Seeing this every time we click on a button is probably not very nice:
screen shot 2016-04-20 at 7 48 46 am


jsantell:
r+

PR #196, opened 2016-04-20
Fix NYC to check all sources.

Mossop:


package.json
```diff
@@ -66,5 +66,11 @@
   },
   "engines": {
     "node": ">=5.8.0"
+  },
+  "nyc": {
+    "all": true,
+    "require": "babel-register",
+    "include": ["app/**"],
+    "extension": [".jsx"]
```
victorporof: Strange that this array doesn't contain '.js' in it.

.babelrc
```diff
@@ -1,3 +1,4 @@
 {
-  "presets": [ "es2015-node5", "stage-2", "react" ]
+  "presets": [ "es2015-node5", "stage-2", "react" ],
+  "sourceMaps": "inline"
```
victorporof: Kinda' sad that we have to inline sourcemaps.

Mossop: It looks like this is only applying to when babel-register is used, we can certainly override that for the actual code we run in task-build.js

package.json
Mossop: Yeah, seems to work though.

ncalexan:
Between this and #209, this looks good.

PR #204, opened 2016-04-20
Document profile service.

ncalexan:
@bgrins I'm not much for readable prose.  Feel free to correct me.

@rnewman I folded most of your formalism in, just changing notation.

Between the two of you, r?


rnewman:
Documenting in-person discussion: I'd like the commands from the front-end to the profile service to be phrased as events, not as 'set'. So we'd have `DID_UNSTAR_PAGE` & `DID_STAR_PAGE`, or `DID_CHANGE_STARRED`, rather than `SET_*`.

ncalexan:
Landed.

PR #205, opened 2016-04-20
Fix the package task. Fixes #185. r=mossop

### LINKED ISSUE #185: Linux packaged version fails to start
I ran `npm run package` and then extracted `dist/tofino-0.1.0-linux-x64.zip` into a new directory. From there, running `./tofino` fails with:

```
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find module 'immutable'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.Module._load (module.js:290:25)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (/home/fabrice/sandbox/tofino/resources/app/lib/main/reducers/bookmarks-reducers.js:12:18)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
```

linclark:
This PR:
- makes the paths more explicit so that it doesn't ignore things like `node_modules/immutable/dist/immutable.js`
- un-ignores the ui directory. I'm not sure what change happened to necessitate this, but there's no UI without the ui dir

Is there a better way to keep this list up to date?

### DIFF
--- a/build/task-package.js
+++ b/build/task-package.js
@@ -16,18 +16,14 @@ const PLATFORM = os.platform();
 
 const IGNORE = [
   // Ignore build stuff
-  '/appveyor.yml',
-  '/branding($|/)',
-  '/build($|/)',
-  '/dist($|/)',
+  '/\/appveyor.yml',
+  '/\/branding($|/)',
+  '/\/build($|/)',
+  '/\/dist($|/)',
   '/electron($|/)',
-  '/gulpfile.js',
   '/README.md',
   '/scripts($|/)',
   '/test($|/)',
-
-  // Ignore source code
-  '/ui($|/)',
 
   // Ignore `/app` and `/shared` once we compile main process code
   // ahead of time, rather than during runtime


build/task-package.js
```diff
@@ -16,19 +16,15 @@ const PLATFORM = os.platform();
...
   '/README.md',
   '/scripts($|/)',
   '/test($|/)',
 
-  // Ignore source code
-  '/ui($|/)',
-
   // Ignore `/app` and `/shared` once we compile main process code
   // ahead of time, rather than during runtime
   // '/app($|/)',
```
victorporof: We are now compiling main process code ahead of time. Shouldn't this comment be updated?

linclark:
it looks like we can't ignore shared yet. If we do, then [this import](https://github.com/mozilla/tofino/blob/master/app/main/reducers/bookmarks-reducers.js#L13) fails.

linclark:
also, which `shared` are we trying to ignore there? There are multiple `shared` dirs. Both of them are in the `app` directory, so should be covered by the ignore on `app`

PR #206, opened 2016-04-20
Wrap history recording in a transaction.

rnewman:
This introduces an `inTransaction` method, and then takes care to track IDs and chain them correctly.

### DIFF
--- a/test/profile/test-storage.js
+++ b/test/profile/test-storage.js
@@ -128,4 +128,28 @@ describe('open', () => {
       done();
     }());
   });
+
+  it('Saves visits.', (done) => {
+    (async function () {
+      const tempDir = tmp.dirSync({}).name;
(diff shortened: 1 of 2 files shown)


test/profile/test-storage.js
```diff
@@ -128,4 +128,28 @@ describe('open', () => {
       done();
     }());
   });
+
+  it('Saves visits.', (done) => {
+    (async function () {
+      const tempDir = tmp.dirSync({}).name;
```
ncalexan: Can we figure out the Mocha `before` or `setup` function and use it?  I think we have `before` somewhere else in the tree.  (And `after`.)

ncalexan:
This looks good.

rnewman:
Landed as part of #207.

PR #207, opened 2016-04-20
Substring visit matching.

rnewman:
@bgrins!

### DIFF
--- a/app/services/storage.js
+++ b/app/services/storage.js
@@ -58,6 +58,12 @@ export class ProfileStorage {
     return new ProfileStorage(db).init();
   }
 
+  close() {
+    this.places.clear();
+    this.nextPlace = 0;
+    return this.db.close();
+  }
+
   /**
    * Load the place map out of the DB.
    */
@@ -76,6 +82,20 @@ export class ProfileStorage {
 
     this.nextPlace = max + 1;
     return max;
+  }
+
+  async init() {
+    const schema = new ProfileStorageSchemaV1();
+    const v = await schema.createOrUpdate(this);
+
+    if (v !== schema.version) {
+      throw new Error(`Incorrect version ${v}; expected ${schema.version}.`);
+    }
+
+    const loaded = await this.loadPlaces();
+    console.log(`Loaded ${loaded} places.`);
+
+    return this;
   }
 
   async savePlace(url, now = microtime.now()) {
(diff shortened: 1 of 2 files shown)


app/services/storage.js
```diff
@@ -180,10 +180,37 @@ export class ProfileStorage {
...
+      out.push(row.url);
+    }
+    return out;
+  }
 
+  async visitedMatches(substring, since = 0, limit = 10) {
+    // Fetch all places visited, with the latest timestamp for each.
+    // We can't limit here, because we don't know matching URLs.
+    // TODO: inverting this query might work better if `since` is long ago.
+    const sub = `SELECT place, MAX(ts) AS latest
```
ncalexan: There's a good deal of duplication here.  Can we streamline?

Perhaps this mostly falls away when we start maintaining our in-memory views?

ncalexan:
These look good.

app/services/storage.js
rnewman: This'll get reworked a little when we have to string match on more things, and yes, there will be some intersection of in-DB views, in-DB materialized views, and in-memory materialized views.

app/services/storage.js
```diff
@@ -78,7 +84,51 @@ export class ProfileStorage {
...
+    return this;
+  }
+
+  /**
+   * Begins a transaction and runs `f`. If `f` returns a promise that rejects,
+   * rolls back the transaction and passes through the rejection. If `f` resolves
+   * to a value, commits the transaction and passes through the resolved value.
+   *
+   * @param f the function to call within a transaction.
+   * @returns {*} the result or rejection of `f`.
```
bgrins: All of these comments say returns `{*}` but really they return a `Promise` that resolves with `{*}`

app/services/storage.js
```diff
@@ -78,7 +84,51 @@ export class ProfileStorage {
...
+   */
+  inTransaction(f) {
+    return this.db
+               .run('BEGIN TRANSACTION')
+               .then(f)
+               .then((result) =>
+                 this.db
+                     .run('COMMIT')
+                     .then(() => Promise.resolve(result)))
+               .catch((err) =>
```
bgrins: Is there a good way to add a test case for failures on functions in this file? It can be hard to follow all of the promise rejections / catching to figure out what the expected return values are in the case of a failure

Issue #200, opened 2016-04-20
Show autocomplete suggestions in UI


bgrins:
After #184 suggestions will be returned for url bar input.  We should show this in the UI like: http://people.mozilla.org/~shorlander/stuff/images/fawkes-explorations-06-05-search.png

bgrins:
This was landed

PR #208, opened 2016-04-20
Show completions in the UI. r=vporof. Fixes Issue #200

bgrins:
r? @victorporof

### DIFF
--- a/app/ui/browser/views/navbar/location.jsx
+++ b/app/ui/browser/views/navbar/location.jsx
@@ -163,49 +163,49 @@ class Location extends Component {
 
     let completions = null;
     if (SHOW_COMPLETIONS && profile.completions.get(urlValue) && focusedURLBar) {
-      const results = profile.completions.get(urlValue).map((completion) => {
-        return <div>{completion}</div>;
-      });
+      const results = profile.completions.get(urlValue).map((completion) =>
+        <div>{completion}</div>
+      );
       completions = <div className={LOCATION_BAR_AUTOCOMPLETE_STYLE}>{results}</div>;
     }
 
     return (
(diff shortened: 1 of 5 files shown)


app/ui/browser/views/navbar/location.jsx
```diff
@@ -17,6 +17,15 @@ import Btn from './btn';
 
 import { fixURL, getCurrentWebView } from '../../browser-util';
 
+// Pref off completions for now since the view doesn't do anything besides
+// show them.
+const SHOW_COMPLETIONS = false;
```
bgrins: If you want to try it locally you'd need to flip this bool

app/ui/browser/views/navbar/location.jsx
```diff
@@ -28,6 +37,18 @@ const LOCATION_BAR_STYLE = Style.registerStyle({
   borderRadius: '2px',
 });
 
+const LOCATION_BAR_AUTOCOMPLETE_STYLE = Style.registerStyle({
+  position: 'absolute',
```
victorporof: This makes it not look like the mockup. The code would also be simpler if we just made the autocompletion list part of the chrome, and not some floating thing.

https://people.mozilla.org/%7Eshorlander/stuff/images/fawkes-explorations-06-05-search.png

bgrins: I have lots of unanswered questions about that mock (like what happens with overflow), and trying to match it is proving tricky.  Are you OK with landing with the popup mode and then do styling updates in follow ups?

victorporof: As discussed on Slack, ok.

bgrins:
This was merged

PR #212, opened 2016-04-20
Store test

ncalexan:
The intention here is just to show that we can test the main process actions and reducers.


app/main/store/store.js
```diff
@@ -0,0 +1,21 @@
...
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+
+import { applyMiddleware, createStore } from 'redux';
+import rootReducer from '../reducers';
+import thunk from '../../shared/thunk';
+
+export default function configureStore() {
+  const store = createStore(rootReducer, applyMiddleware(thunk));
```
bgrins: Looks like middleware changed from logger to thunk, is that intentional?

PR #217, opened 2016-04-20
Testing for linux failures

jsantell:
https://travis-ci.org/mozilla/tofino/builds/124618508


jsantell:
Was an issue with other things dispatching to ipc that were not yet run since they're async, so just filters out the events that we're not interested in. Surprisingly this happened on OSX as well, and Linux sometimes, so yay our first intermittent

jsantell:
Replaced via #229

Issue #167, opened 2016-04-19
Make content scroll behind the title bar


victorporof:

victorporof:
See also https://github.com/mozilla/tofino/issues/166

PR #221, opened 2016-04-21
Fix browser chrome not being draggable in certain cases

victorporof:
Clicking close to the sides doesn't drag the app. This is because the child selectors previously used aren't relevant anymore.


victorporof:
Obviously only commit ccd9894 is relevant, as this is based off https://github.com/mozilla/tofino/issues/167

jsantell:
r+

PR #223, opened 2016-04-21
Add displayName to all stateless components. Fixes #222. r=vporof

### LINKED ISSUE #222: Add displayName for all stateless components
It turns out stateless components need the `displayName` to be manually set if you want them to be labeled properly in the Perf tool.

No displayName:
1before

With displayName:
2after

linclark:
2after


victorporof:
+100 to this. We should also enforce it via eslint. Care to look into that as well, Lin?

linclark:
👍 I can take a look at that

PR #224, opened 2016-04-21
Add ESLint rule: react/display-name. r=vporof

linclark:
This worked on local, but not exactly as expected. Going to see what CI does with it before landing #223 just to make sure it actually fails things correctly.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -57,6 +57,7 @@
     }],
     "flow-vars/define-flow-type": 1,
     "flow-vars/use-flow-type": 1,
+    "react/display-name": ["error", { "ignoreTranspilerName": true }],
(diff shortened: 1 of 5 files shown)


linclark:
r? vporof

Even though this is red, I don't think it's the cause of the failures, so I think this is ready to land.

victorporof:
r++!

victorporof:
Do you prefer to land this yourself, or should I squash? I'm asking because I don't know whether or not the commits are there to help reviewing or because they're relevant history.

(As a team, we all have different opinions about this. Some people prefer squashing, some carefully organize their PRs into multiple relevant commits that should be part of history.)

linclark:
I usually squash via the UI these days. I'll go ahead and do that.

PR #220, opened 2016-04-21
Make content scroll behind the title bar. Fixes #167 

victorporof:
https://github.com/mozilla/tofino/issues/167

Screencap: https://dl.dropboxusercontent.com/u/2388316/scroll.mp4

Yo @bgrins check this out.

PR #227, opened 2016-04-21
Make location bar look good on small browser window sizes

victorporof:
Commit 48cff5b and aad10c8 are relevant, as since this has branched off other PRs like https://github.com/mozilla/tofino/pull/221 and https://github.com/mozilla/tofino/pull/220


jsantell:
r+. We should probably think of our 'responsive' story.

PR #154, opened 2016-04-21
Packager includes source code from lib/

Mossop:
Now that we babel main process code, we should probably update our packager. There's some comments in there about ignoring app/ and shared/ as well. It looks like that entire thing needs to be updated.


Mossop:
This isn't ready to land yet.

jsantell:
r+

PR #233, opened 2016-04-21
Check that all npm dependencies are installed before starting any tasks.

Mossop:

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -8,30 +8,38 @@
  */
 /* eslint-disable strict */
 /* eslint-disable import/no-commonjs */
+/* eslint-disable no-console */
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


build/main.js
```diff
@@ -14,23 +14,31 @@
...
 
-const handleTaskFailed = e => {
-  /* eslint no-console: 0 */
-  console.error('Build failed.');
-  console.error(e);
-  process.exit(1);
-};
+require('check-dependencies')((result) => {
+  if (!result.depsWereOk) {
+    console.error('Dependency checks failed.');
```
bgrins: Need the eslint console comment here too?

Mossop: I think it just applies globally, I'll move it to the top so that is clear.

PR #234, opened 2016-04-21
Pass npm arguments through to mocha to allow selecting specific directories

### LINKED ISSUE #231: Split tests into type directories
We've talked about splitting tests into different directories based on the type and having different targets for each plus a top test command:

    test/functional
    test/lint
    test/end2end

Mossop:
This allows us to run individual suites like `npm test test/lint`. We could add a helper that prefixed `test/` on paths if that was helpful.

Fixes #231.

Is this all we need instead of the quick test stuff @joewalker?


linclark:
Tested this manually and it worked for me.

PR #236, opened 2016-04-21
Add keyboard control to completions

bgrins:


bgrins:
@victorporof I talked with @phlsa about the completion UI in http://people.mozilla.org/~shorlander/stuff/images/fawkes-explorations-06-05-search.png and sounds like the way to proceed for now is to treat it as a popup, instead of having it push down the content / extend the nav bar (although the background visually extends to the sides of the window instead of just to the width of the URL bar).

app/ui/browser/views/navbar/location.jsx
```diff
@@ -169,12 +171,22 @@ class Location extends Component {
...
   }
 
-  handleURLBarKeyDown(ev) {
+  handleURLBarKeyDown(maxCompletions, ev) {
     if (ev.keyCode === 13) { // enter
       this.props.navigateTo(fixURL(ev.target.value));
     } else if (ev.keyCode === 27) { // esc
       this.props.onLocationReset();
       ev.target.select();
+    } else if (ev.keyCode === 40 && maxCompletions > 0) { // down
```
jsantell: Let's use constants for these keycodes, we use them in a few places (can follow up)

app/ui/browser/views/navbar/location.jsx
```diff
@@ -221,7 +238,9 @@ class Location extends Component {
             onFocus={this.handleURLBarFocus}
             onBlur={this.handleURLBarBlur}
             onChange={this.props.onLocationChange}
-            onKeyDown={this.handleURLBarKeyDown}
+            onKeyDown={(e) =>
```
jsantell: event handler function signature should start with `e` as the first argument IMO

jsantell:
lgtm

victorporof:
@bgrins That's exactly what I was saying, there never was any suggestion for pushing down content...

PR #238, opened 2016-04-21
Use unique IDs for every tab, and update the related actions to take …

### LINKED ISSUE #202: Uniquely identify tabs
Right now we modify tabs via their index, which can change (reordering, closing tabs, etc.) -- we should just use an id for this.

Should these be created/managed on the backend?

jsantell:
…a tab ID rather than an index. r=bgrins

Closes #202

### DIFF
--- a/app/ui/browser/actions/main-actions.js
+++ b/app/ui/browser/actions/main-actions.js
@@ -26,37 +26,37 @@ export function createTab(location) {
   return { type: types.CREATE_TAB, location, instrument: true };
 }
 
-export function duplicateTab(pageIndex) {
-  return { type: types.DUPLICATE_TAB, pageIndex, instrument: true };
+export function duplicateTab(pageId) {
+  return { type: types.DUPLICATE_TAB, pageId, instrument: true };
 }
 export function attachTab(page) {
   return { type: types.ATTACH_TAB, page, instrument: true };
 }
 
-export function closeTab(pageIndex) {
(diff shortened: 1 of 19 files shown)


app/ui/browser/actions/main-actions.js
```diff
@@ -26,43 +26,43 @@ export function createTab(location) {
...
+export function setPageOrder(pageId, updatedIndex) {
+  return { type: types.SET_PAGE_ORDER, pageId, updatedIndex, instrument: true };
 }
 
 export function setPageAreaVisibility(visible) {
   return { type: types.SET_PAGE_AREA_VISIBILITY, visible };
 }
 
-export function setUserTypedLocation(payload) {
+export function setUserTypedLocation(pageId, data) {
```
bgrins: Keep the 'payload' name

app/ui/browser/model/index.js
```diff
@@ -52,4 +54,8 @@ export const Page = Immutable.Record({
   canRefresh: false,
   chromeMode: 'expanded',
   commands: Immutable.List(),
-});
+}) {
+  constructor(data) {
+    super(Object.assign({ id: uuid.v4() }, data));
```
jsantell: @rnewman here's where we create a uuid -- we can [configure the options](https://www.npmjs.com/package/uuid#uuidv4options--buffer--offset) to whatever and move this into a shared util used by all processes

PR #243, opened 2016-04-21
Fix switching between two tabs when clicking on a tab in the tabbar. …

jsantell:
…r=linclark

Signed-off-by: Jordan Santell <jsantell@mozilla.com>


bgrins:
works for me

PR #251, opened 2016-04-22
Transition navbar opacity quicker when switching into expanded mode

victorporof:

@jsantell I believe this fixes your first bullet point in https://github.com/mozilla/tofino/pull/220#issuecomment-213006017


jsantell:
nice, r+!

PR #256, opened 2016-04-22
Materialized views for visited history and starred pages.

rnewman:
This introduces two materialized views, replaces the accessors to use them, and maintains them as events are processed.


rnewman:
@ncalexan

ncalexan:
Code looks fine, but I think you should test the new functionality.

PR #257, opened 2016-04-22
move SHOW_COMPLETIONS pref into constants file. r=vporof

bgrins:
@victorporof what do you think about this?  Not sure if you were intending this ui constants file to be more for css constants but thought it might be good to consolidate 'prefs' in one place for this feature and future prototypes.


victorporof:
Definitely a good idea. Was planning to do just this but you were faster than me :)

PR #261, opened 2016-04-22
Fix array of render components in Location that need unique keys. r=l…

jsantell:
…inclark

Fix error: "Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Location`. See https://fb.me/react-warning-keys for more information."


linclark:
:shipit:

PR #262, opened 2016-04-22
Use constants for keyCodes. r=bgrins

jsantell:


app/ui/browser/actions/external.js
```diff
@@ -227,7 +228,7 @@ export function close() {
  * Perform the search if the user just pressed return
  */
 export function inPageSearch(ev) {
-  if (ev.keyCode === 13) {
+  if (ev.keyCode === KEY.ENTER) {
```
bgrins: What if we did `ev.key === "Enter"`, `ev.key === "KeyF"`, `ev.key === "Escape"`, etc instead of making constants?

victorporof: I second this.

jsantell: Is that easier? Are there going to be cross platform differences with using that? (I honestly don't know)

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Key_values

bgrins: I think it's a little easier since we won't need to add a new constant every time we handle a new key.  Not sure about cross platform issues but seems like there shouldn't be any

jsantell:
@bgrins updated to use `ev.key`

bgrins:
LGTM

PR #263, opened 2016-04-22
Adds a checker that forces an electron-rebuild if the electron version or native module versions have changed.

### LINKED ISSUE #260: Make it so we don't need to run electron-rebuild -f on every install

Mossop:
Fixes #260.

Can you review this @rnewman?


rnewman:
I love it. Ship it!

PR #265, opened 2016-04-22
Fix quicktest by not passing --test through to mocha and not clobbering the build config after setting quickTest.

Mossop:


joewalker:
LGTM

PR #266, opened 2016-04-22
Always build before running tests so compiled JS is available for electron to run.

Mossop:


Mossop:
This works so will let devs run end to end tests without packaging. We probably want to make CI run it against packages either instead of or as well though.

joewalker:
On phone in YVR, so this is from memory, but I think we might be able to remove getBuiltExecutable now?

Mossop:
I was going to leave it in so we could use it to run packaged builds in CI
in the future, I don't mind taking it out till then though

joewalker:
I think that getBuiltExecutable is easy to write and feels a bit custom for a utility anyway, so I'd vote to remove it.
LGTM otherwise.

PR #268, opened 2016-04-22
Add some flow suppress comments.

LegNeato:
This comes in handy to work around issues and later remove the
workarounds.

More details:

http://flowtype.org/docs/advanced-configuration.html#libs


victorporof:
Sure.

PR #270, opened 2016-04-22
[flow] Add typechecking to TravisCI

LegNeato:


Mossop:
Looks like we did this in #276.

PR #271, opened 2016-04-23
Remove the webview-wrapper, fixes #250

### LINKED ISSUE #250: Don't wrap webview in a custom element
See https://github.com/mozilla/tofino/blob/master/app/ui/shared/web-view.js#L16

We don't need this if we use react's `is`: https://github.com/bgrins/gecko-dev/blob/console-frontend/devtools/client/webconsole/new-console-output/components/message-types/console-api-call.js#L36

victorporof:

Fixes https://github.com/mozilla/tofino/issues/250


linclark:
Cleans things up nicely, LGTM 👍

jsantell:
Landing this to start checking out how to unit test with a webview

PR #273, opened 2016-04-25
Keep BrowserWindow list in main process Redux store.

ncalexan:
A few notes:
* there are more callers of `makeBrowserWindow` than I'd like.  I expect those to dispatch profile commands eventually.
* there are no writes to the storage layer yet.  Persisting windows and state is to follow.

r? @bgrins


app/main/reducers/browserWindows-reducers.js
```diff
@@ -0,0 +1,47 @@
...
+ * We diff the current and previous state.  However, after a BrowserWindow is closed, its id
+ * property is no longer available.  (The error will say "Object has been destroyed".)  Therefore,
+ * to allow to compare to a previous state with a BrowserWindow that is no longer alive, we key
+ * by the BrowserWindow.id and take care to not fetch the id of closed windows.
+ */
+type State = Immutable.Map<number, electron.BrowserWindow>;
+
+export default function browserWindowsReducer(state: State = new Immutable.Map(),
+                                              action: Object): State {
+  console.log(`reducing ${JSON.stringify(action)}`);
```
bgrins: Is this intentionally logging for all actions (even those that aren't handled by this reducer)?  It's actually useful, just was surprised when running it to see logs for all the actions, as it might get noisy.

app/main/reducers/profile-command-reducers.js
```diff
@@ -22,15 +23,15 @@ import { ProfileStorage } from '../../services/storage';
 
 export default async function commandHandler(storage: ProfileStorage,
                                              dispatch: any,
-                                             sessionId: number,
+                                             browserWindow: electron.BrowserWindow,
                                              command: profileActions.ProfileAction): Promise<void> {
-  console.log(`${sessionId}: Reducing ${JSON.stringify(command)}`);
+  console.log(`Reducing ${JSON.stringify(command)}`);
```
bgrins: Ditto - maybe there should be a prefix on this message like "profile-command-reducer" to make it clear where the log is coming from

app/main/actions/profile-actions.js
```diff
@@ -48,3 +50,21 @@ export function completionsFor(text: string,
...
+export function createBrowserWindow(browserWindow: electron.BrowserWindow): ProfileAction {
+  return {
+    type: profileActionTypes.DID_CREATE_BROWSER_WINDOW,
+    payload: {
+      browserWindow,
+    },
+  };
+}
+
+export function closeBrowserWindow(browserWindow: electron.BrowserWindow): ProfileAction {
```
bgrins: I think overloading action / command names is going to get more confusing as time goes on (DID_CLOSE_BROWSER_WINDOW and closeBrowserWindow), and I find the pattern used for IPC communication with 'new-window' / createBrowserWindow easier to follow.   That is, the ipc communication is using plain event listeners on server side, and the listener is in charge of emitting a profileAction (takes profileCommand out of the equation).  Why is closing a window a profile-command but opening one isn't?

app/main/browser.js
```diff
@@ -48,24 +47,41 @@ const globalShortcut = electron.globalShortcut;
...
-function sendDiffsToWindows() {
+function sendDiffsToWindows(): void {
   const previousState = currentState;
   currentState = store.getState();
 
+  // TODO: handle empty state.
+  if (previousState && !Immutable.is(currentState.browserWindows.keySeq(),
+                                     previousState.browserWindows.keySeq())) {
+    // Show new windows, taking care to use key IDs rather than possibly deleted ID members.
+    currentState.browserWindows.forEach((bw, id) => {
```
jsantell: We should probably yield on the showing and closing of windows for the entire function, but still run in parallel.

```js
let opening = browserWindows.map((bw,id) => currentState.has(id) ? null : bw.p.then(()=>bw.show()));
let closing = browserWindows.map((bw,id) => prevState.has(id) ? null : bw.p.then(()=>bw.close()));

yield Promise.all([opening, closing])
```

app/main/actions/profile-actions.js
ncalexan: @bgrins and I discussed this on Slack.  We're fairly sure the command separation will go away over time, since it's not carrying it's weight.

As for why closing is a command, but opening isn't -- that's just work in progress.  I expect all of these things to be commands soon enough.

PR #275, opened 2016-04-25
[WIP] Reduce time spent rerendering when interacting with the page. Fixes #253. r=jsantell

### LINKED ISSUE #253: Use connect() for Page components to reduce wasted time
Pages change pretty frequently. Whenever a user rolls over a link, `page.statusText` will change. Whenever a user types, `page.userTyped` will change.

Currently, we are diffing the whole tree when a page changes. We should be able to make this more efficient without too much work.

We are using connect() in one place, at the top of the tree, and we’re accessing the whole state tree. Redux automatically implements `shouldComponentUpdate()` for any container components (the higher order component that is returned from a call to `connect()`). However, it does not do that for the children. This means that a change to state causes the whole component tree to go through `_performComponentUpdate()`, which results in wasted time doing render tree construction and virtual DOM comparisons that do not result in a DOM modification

If we split this up and have containers at lower levels, we should see less wasted time in the perf tool. The basic concept of this is explained in a talk from [React.js Conf 2015](https://www.youtube.com/watch?v=KYzlpRvWZ6c), and Dan Abramov recently provided an [example of doing it with Redux](https://twitter.com/dan_abramov/status/719723513016426496).

There was some concern in the React components meeting that this would cause forced updates. It [does not cause a call to `forceUpdate()`](https://github.com/reactjs/redux/issues/1303#issuecomment-188618535). Instead, it works with React’s dirty component queue. It will check each connected component and if it’s possible that it will need an update, it will add it to the dirty component queue. This means that connected components can be added to the queue independent of their parents.

@jsantell @victorporof & anyone else, let me know if you want to talk through this before I start.

linclark:
As mentioned in #253, we are doing lots of rerenders when the page object changes, which it does frequently because of the `SET_PAGE_DETAIL` and `SET_USER_TYPED_LOCATION` action types.

This PR gives us a big decrease in the amount of time spent.

I created a test which creates 6 additional tabs (7 total) and then sets the status text 3 times.

Before (~37ms):
perf-tabreload-before

After (~9ms):
perf-tabreload-after

If you want to try it yourself, I posted a patch with the [test code](https://gist.github.com/linclark/4171e8483b035c425ed4103a97df4347). Feedback welcome on the testing technique.

### DIFF
--- a/app/ui/browser/actions/main-actions.js
+++ b/app/ui/browser/actions/main-actions.js
@@ -11,17 +11,20 @@ specific language governing permissions and limitations under the License.
 */
 
 import * as types from '../constants/action-types';
+import { Page } from '../model';
 import * as profileCommands from '../../../shared/profile-commands';
 import { ipcRenderer } from '../../../shared/electron';
 
 export function createTab(location) {
-  return { type: types.CREATE_TAB, location, instrument: true };
+  const page = location ? new Page({ location }) : new Page();
+  return { type: types.CREATE_TAB, page, instrument: true };
 }
 
 export function duplicateTab(pageId) {
   return { type: types.DUPLICATE_TAB, pageId, instrument: true };
 }
 export function attachTab(page) {
+  page = new Page(page);
   return { type: types.ATTACH_TAB, page, instrument: true };
 }
 
--- a/app/ui/browser/model/index.js
+++ b/app/ui/browser/model/index.js
@@ -34,6 +34,8 @@ export const State = Immutable.Record({
   // This is a list of Page objects (see below)
   pages: Immutable.List(),
 
+  pageIds: Immutable.Set(),
+
   // The currently displayed tab
   currentPageIndex: -1,
 
--- a/app/ui/browser/model/index.js
+++ b/app/ui/browser/model/index.js
@@ -14,6 +14,11 @@ import Immutable from 'immutable';
 import uuid from 'uuid';
 
 /**
+ * Fairly sure we should hard code this
+ */
+const HOME_PAGE = 'https://www.mozilla.org/';
+
+/**
  * The browser window's view of its profile.
  */
 export const Profile = Immutable.Record({
@@ -43,7 +48,7 @@ export const State = Immutable.Record({
  */
 export class Page extends Immutable.Record({
   id: null,
-  location: undefined,
+  location: HOME_PAGE,
   title: 'New Tab',
   statusText: false,
   userTyped: null,
--- a/app/ui/browser/reducers/main-reducers.js
+++ b/app/ui/browser/reducers/main-reducers.js
@@ -19,8 +19,10 @@ import { State, Page } from '../model';
 import * as profileDiffTypes from '../../../shared/constants/profile-diff-types';
 import { isUUID } from '../browser-util';
 
+const initialPage = new Page();
 const initialState = new State({
-  pages: Immutable.List.of(new Page()),
+  pages: Immutable.List.of(initialPage),
+  pageIds: Immutable.Set.of(initialPage.id),
   currentPageIndex: 0,
   pageAreaVisible: false,
 });
@@ -91,6 +93,8 @@ function createTab(state, page) {
     mut.update('pages', pages => pages.push(page));
     mut.set('currentPageIndex', state.pages.size);
     mut.set('pageAreaVisible', true);
+
+    mut.update('pageIds', pageIds => pageIds.add(page.id));
   });
 }
 
@@ -109,6 +113,8 @@ function attachTab(state, page) {
   return state.withMutations(mut => {
     mut.set('pages', Immutable.List.of(page));
     mut.set('currentPageIndex', 0);
+
+    mut.update('pageIds', pageIds => pageIds.add(page.id));
   });
 }
 
@@ -129,6 +135,8 @@ function closeTab(state, pageId) {
 
   return state.withMutations(mut => {
     mut.update('pages', pages => pages.delete(pageIndex));
+
+    mut.update('pageIds', pageIds => pageIds.delete(pageId));
 
     // If tab closed comes before our current tab, or if this is the right-most
     // tab and it's selected, decrement the current page index.
--- a/app/ui/browser/reducers/main-reducers.js
+++ b/app/ui/browser/reducers/main-reducers.js
@@ -19,13 +19,8 @@ import { State, Page } from '../model';
 import * as profileDiffTypes from '../../../shared/constants/profile-diff-types';
 import { isUUID } from '../browser-util';
 
-/**
- * Fairly sure we should hard code this
- */
-const HOME_PAGE = 'https://www.mozilla.org/';
-
 const initialState = new State({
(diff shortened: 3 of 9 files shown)


app/ui/browser/views/browser.jsx
```diff
@@ -162,7 +163,13 @@ BrowserWindow.propTypes = {
   pageAreaVisible: PropTypes.bool.isRequired,
 };
 
-export default BrowserWindow;
+const mapStateToProps = (state) => ({
+  // @TODO restructure the state so this list is automatically created as new todos
```
ncalexan: "new todos"?

linclark:
btw, this very likely breaks multiple tests. I haven't looked at them yet, wanted to get feedback on the approach before changing them

app/ui/browser/views/browser.jsx
```diff
@@ -11,6 +11,7 @@ specific language governing permissions and limitations under the License.
 */
 
 import React, { PropTypes, Component } from 'react';
+import { connect } from 'react-redux';
```
ncalexan: There's no motiviation for this change in the commit comment.  Can you explain why this is better?

app/ui/browser/views/navbar/navbar.jsx
```diff
@@ -229,4 +230,9 @@ NavBar.propTypes = {
   navigateTo: PropTypes.func.isRequired,
 };
 
-export default NavBar;
+const mapStateToProps = (state) => ({
+  // @TODO get this from the restructured reducer
```
ncalexan: In general, TODO comments that don't explain what happens now, what should happen, and why the one is better than the other, don't add much.

app/ui/browser/views/browser.jsx
linclark: sorry, copy pasta

app/ui/browser/actions/main-actions.js
```diff
@@ -11,17 +11,20 @@ specific language governing permissions and limitations under the License.
...
 export function createTab(location) {
-  return { type: types.CREATE_TAB, location, instrument: true };
+  const page = location ? new Page({ location }) : new Page();
+  return { type: types.CREATE_TAB, page, instrument: true };
 }
 
 export function duplicateTab(pageId) {
   return { type: types.DUPLICATE_TAB, pageId, instrument: true };
 }
 export function attachTab(page) {
```
ncalexan: Consider using Flow to make this explicit.

app/ui/browser/reducers/main-reducers.js
```diff
@@ -19,13 +19,8 @@ import { State, Page } from '../model';
...
 import { isUUID } from '../browser-util';
 
-/**
- * Fairly sure we should hard code this
- */
-const HOME_PAGE = 'https://www.mozilla.org/';
-
 const initialState = new State({
-  pages: Immutable.List.of(new Page({ location: HOME_PAGE })),
+  pages: Immutable.List.of(new Page()),
```
ncalexan: Nice, we handle the empty state now?

app/ui/browser/model/index.js
```diff
@@ -14,6 +14,11 @@ import Immutable from 'immutable';
 import uuid from 'uuid';
 
 /**
+ * Fairly sure we should hard code this
```
ncalexan: I'm actively pursuing having the main process inject the set of pages into each created window process for session restore.  I don't think that will impact your approach.

app/ui/browser/views/browser.jsx
```diff
@@ -163,9 +163,7 @@ BrowserWindow.propTypes = {
 };
 
 const mapStateToProps = (state) => ({
-  // @TODO restructure the state so this list is automatically created as new todos
-  // are added instead of generated on each update pass
-  pageIds: state.browserWindow.pages.toArray().map(page => page.id),
+  pageIds: state.browserWindow.pageIds,
```
ncalexan: Adding state to avoid a single `map` seems like the wrong trade-off to me.  Can you explain why this is better than the existing code, and better than an `OrderedMap` from `pageId` to `page`?

app/ui/browser/actions/main-actions.js
jsantell: ATTACH_TAB isn't implemented on the main process, and it's unclear how this should even work

app/ui/browser/actions/main-actions.js
```diff
@@ -11,17 +11,20 @@ specific language governing permissions and limitations under the License.
 */
 
 import * as types from '../constants/action-types';
+import { Page } from '../model';
 import * as profileCommands from '../../../shared/profile-commands';
 import { ipcRenderer } from '../../../shared/electron';
 
 export function createTab(location) {
-  return { type: types.CREATE_TAB, location, instrument: true };
+  const page = location ? new Page({ location }) : new Page();
```
jsantell: Does immutable set undefined? Like, I'd imagine/want `new Page({ location })` to use the default location and not need ternary operator -- if that's not the case, we should probably do something about that, making defaults less useful

app/ui/browser/views/navbar/navbar.jsx
```diff
@@ -207,7 +208,7 @@ NavBar.displayName = 'NavBar';
 
 NavBar.propTypes = {
   page: PropTypes.object,
-  pages: PropTypes.object.isRequired,
+  numPages: PropTypes.number.isRequired,
```
jsantell: Maybe `pageCount` or something? `numPages` isn't clear what it was representing to me

app/ui/browser/views/navbar/navbar.jsx
```diff
@@ -229,4 +230,13 @@ NavBar.propTypes = {
   navigateTo: PropTypes.func.isRequired,
 };
 
-export default NavBar;
+const mapStateToProps = (state) => {
+  const browserWindow = state.browserWindow;
```
jsantell: So `<NavBar/>` only attempts to render if either `numPages` or `page` state changes? Nice nice nice

app/ui/browser/model/index.js
```diff
@@ -29,6 +34,8 @@ export const State = Immutable.Record({
   // This is a list of Page objects (see below)
   pages: Immutable.List(),
 
+  pageIds: Immutable.Set(),
```
jsantell: After going through the code, I still don't quite get why this set exists; wouldn't this denormalizing duplicate what we have in pages, which we could do with `pages.map(p => p.id)`?

linclark: `pages.map` creates a new object, so you don't get the referential transparency you need to do a quick check in `shouldComponentUpdate`.

linclark: Another option that might work (and which could possibly eliminate the need for creating a `pageIds` set) is to move all of the frequently changing values out of the page object. For example, having a separate list of `statusTexts` which can be looked up via `page.id`. Then the page objects themselves shouldn't change as much and would give us the same effect as this `pageIds` array.

linclark:
I'm going to close this one out. We'll be going with a similar approach, but focusing on moving changes out of the page object. See #282.

PR #276, opened 2016-04-26
Run flow check in CI.

Mossop:


Mossop:
Looks like this passes @ncalexan

ncalexan:
Fine by me.  I wasn't comfortable committing to running Flow on every check-in, but now we have two devs who want to :)

Mossop:
I'm of the opinion that if we want to use type checking then we must pass the checks or it is worthless. So if people complain about running it in CI we should question whether we really want type checking at all.

I think we should also run this locally but I'm not set up to be able to figure that out.

PR #277, opened 2016-04-26
Fix the context menu.

Mossop:


jsantell:
LGTM -- the menu functions are starting to look like they should live in page.jsx, or a sibling utility, rather than an action

PR #278, opened 2016-04-26
Handle IPC requests from main process with redux actions.

jsantell:


jsantell:
@victorporof @linclark This is what I'm thinking. It does mean more actions to test, but if these things aren't actions, then we won't have replayability. We could reuse the same actions as `createTab()` for example, but in a history log, we wouldn't be able to tell if the action originated from a button click, or from the main process.

General thoughts? If this looks good I"ll follow up with tests.

app/ui/browser/reducers/main-reducers.js
```diff
@@ -37,15 +37,26 @@ function getPageIndexById(state, id) {
 export default function basic(state = initialState, action) {
   switch (action.type) {
     case types.CREATE_TAB:
```
linclark: What's the use case for `types.CREATE_TAB` vs `types.IPC_COMMAND_CREATE_TAB`? 

It looks like the action creator that uses `types.CREATE_TAB` is only used in `external.js`. It's used in various context menus to create a new tab. But the `BrowserMenu` uses IPC to do the same thing. Even within the context menus in `external.js`, some menu items use `ipcRenderer.send`instead of dispatching actions.

I'm not close enough to the details to know for sure, but it feels like we might want to consolidate those.

linclark: I guess it would probably be too slow to go over IPC for the contextual menu New Tab actions? I don't know much about the cost of IPC.

jsantell: types.CREATE_TAB is used in `<TabBar/>` (defined in `browser.jsx` as `handleNewTabClick`) which is the main local way to create a tab -- the context menus are still non-reduxy (and a lot of the things in `actions/external.js` for the most part) and some have the roundtrip IPC stuff, and those seem mostly legacy from the initial prototype.

I guess the main difference between the two types is distinguishing the source of the action, or what caused it, for better debugging, and possibly replayability

linclark: For posterity, there was a related short convo in Slack. 

My concern is that we have two action pathways here for the same thing:

- `type.CREATE_TAB` creates the tab locally and then pushes the data to the backend through instrumentation
- `type.IPC_COMMAND_CREATE_TAB` is the result of firing `new-tab` on the backend and then creating the tab locally

We discussed if it would be too expensive to go over IPC for all "new tab" events. This would mean we'd no longer have the local version (`types.CREATE_TAB`). All new tabs would be created by firing off a `new-tab` event.

We aren't sure whether it would be too expensive performance-wise to do this. It might be worth it for the simplification of only having one path. However, there were concerns.

Since there isn't consensus around using `new-tab` for all new tabs, we can go ahead with having 2 different actions.

app/ui/browser/views/browser.jsx
```diff
@@ -167,40 +168,22 @@ export default BrowserWindow;
...
-    dispatch(actions.createTab(bookmark.url));
-  });
-
-  ipcRenderer.on('tab-attach', (e, tabInfo) => {
-    const page = tabInfo.page;
-    page.guestInstanceId = tabInfo.guestInstanceId;
-    dispatch(actions.attachTab(page));
-  });
+  ipcRenderer.on('profile-diff', (_, args) => dispatch(args));
+  ipcRenderer.on('focus-url-bar', () => dispatch(ipcActions.focusURLBar()));
```
linclark: Is this working? Cmd-L isn't focusing the url bar for me. (it wasn't before either)

linclark:
Looks good, except could you run it through CI again? I'm guessing the fail was a fluke, but just in case. If that passes, then this is good to go.

app/ui/browser/views/browser.jsx
jsantell: The reducer/action doesn't do anything yet, but will add (I don't even know if this is wired up on the back)

jsantell:
Rebased and repushing

PR #279, opened 2016-04-26
Use a unique key when creating <Page> components so we reuse the corr…

jsantell:
…ect webview instance to prevent mismatching between webview page and <Page> page. r=linclark

Currently, create two tabs, let's say "a" and "b". Closing "a", and mouse over some links on "b", and the setPageDetails calls are using id "a", which throws in our reducer assertions. We were reusing the same page instance that "a" was using for "b", resulting in this error.


linclark:
Ah yes, I had noticed this key issue earlier. I expected it to have an impact on performance. I didn't realize it would have an impact on the functionality though... are we depending on the key somewhere in our code?

linclark:
Actually, I see where the issue would be (in the ref that it's holding on to). LGTM!

PR #281, opened 2016-04-27
Make "New Window" a profile command.

ncalexan:
It's worth noting the setImmediate() in browser-menu.js, which is
surprising.  I think I'm seeing an interaction with showing the browser
window and the child rendering event loops.

There's still clean-up to be done here: I expect to bring
`commandHandler` and `makeBrowserWindow` into a Browser class or
similar.

And I still don't know how to test these actions.


bgrins:
There's some flow errors on this push: https://travis-ci.org/mozilla/tofino/jobs/126184461#L529

Issue #283, opened 2016-04-27
Break up main-reducers.js and use combineReducers


linclark:
Currently, the `main-reducers.js` file contains all the reducers. We can split this up with `combineReducers`.

There is one concern... this can make it harder to access state between two reducers. However, in many cases this problem can be resolved by using selector functions. 

A selector function is called in `mapStateToProps` to select the data off of the state object. For example, `getPages(state)` would simply return `state.browserWindow.pages`. And if you need to combine data from sibling reducers, you can put a selector on a higher level reducer. The higher level reducer can call selectors on its two child reducers and combine the data as needed (as shown in the [shopping cart](https://github.com/reactjs/redux/blob/master/examples/shopping-cart/reducers/index.js) example.)

linclark:
@ncalexan do you think there's still a reason to keep profile, pages, etc, nested under `browserWindow`?

PR #286, opened 2016-04-27
Temporarily remove "Duplicate tab". r=jsantell

linclark:
Right now the "Duplicate tab" menu item doesn't pass in the `pageId` to the `duplicateTab()` action creator, so it doesn't duplicate the tab. I'd suggest that we temporarily remove this. It will make it easier to refactor the reducer in #283.

@jsantell, mind giving this a review?


jsantell:
Wouldn't it be easier to just pass it in `menuTabContext`? It looks like the page ID is being passed in, but the argument still has `pageIndex` in its name, if we want this functionality?

linclark:
I'm not sure whether we want to keep this or not. In other browsers, when you duplicate a tab it will copy over the history as well, so you can use the back button in the duplicated tab as well. We're not doing that, so this feature doesn't seem super useful at the moment. 

I think it might be better to kill it until we decide how we want it to work. But I could be wrong. Do others have opinions?

jsantell:
@linclark Good point, we have no idea if this is useful or if this'll be even used; sgtm!

PR #288, opened 2016-04-27
Webview commands in actions

jsantell:
This moves the webview interactions into function action creators and a way to mock webview interactions

### DIFF
--- a/app/ui/browser/actions/main-actions.js
+++ b/app/ui/browser/actions/main-actions.js
@@ -10,9 +10,31 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
+import assert from 'assert';
 import * as types from '../constants/action-types';
 import * as profileCommands from '../../../shared/profile-commands';
 import { ipcRenderer } from '../../../shared/electron';
+import { getWebViewById, fixURL, getPageIndexById, isUUID } from '../browser-util';
+
+/**
+ * Takes a function and action name. Calls the function/action-creator in
+ * a try block to stop progression of an action creator upon error,
+ * as well as rethrow the error with the related action type.
+ *
+ * @param {Function} fn
+ * @param {String} action
+ * @return {Function}
+ */
+function tagErrors(fn, action) {
+  return function(...args) {
+    try {
+      fn(...args);
+    } catch (e) {
+      e.message = `${action}: ${e.message}`;
+      throw e;
+    }
+  };
+}
 
 export function createTab(location) {
   return { type: types.CREATE_TAB, location, instrument: true };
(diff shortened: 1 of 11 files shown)


app/ui/browser/actions/main-actions.js
```diff
@@ -77,18 +99,54 @@ export function unbookmark(url) {
...
-  return { type: types.NAVIGATE_PAGE_REFRESH, pageId };
+// @TODO Linting https://github.com/eslint/eslint/issues/3587
+export function navigatePageRefresh(pageId, doc=document) { // eslint-disable-line
+  return tagErrors((dispatch, getState) => {
+    assert(isUUID(pageId), `Invalid page id: ${pageId}`);
+    const pageIndex = getPageIndexById(getState(), pageId);
+    assert(pageIndex >= 0, `Page ${pageId} not found in current state.`);
+    assert(getState().browserWindow.pages.get(pageIndex).canRefresh, 'Page cannot refresh.');
+
+    dispatch({ type: types.NAVIGATE_PAGE_REFRESH, pageId });
```
linclark: Since this action type is no longer handled by the reducer, is it still an action for the sake of logging?

jsantell: Correct; but not sure if that's the right way to go about it -- in that case, maybe it makes sense for the reducer to fire the side effect (webview navigation) for replayability?

linclark: Redux, and a lot of the ecosystem around Redux, assume that the reducer has no side effects. From the Redux docs:

> It’s very important that the reducer stays pure. Things you should never do inside a reducer:
>
>    Mutate its arguments;
>    Perform side effects like API calls and routing transitions;
>    Call non-pure functions, e.g. Date.now() or Math.random().

I don't think we'll be able to achieve replayability for these actions. I might be seeing it wrong, so if you think we can maybe we should walk through a use case.

jsantell: Good point -- not sure about this then. Either way, can move it elsewhere later for the replayability effect if it's even possible. Does it look good?

linclark: In this setup, we'd send actions to the store that don't affect the store. That feels a little off to me. It means we're doing extra work (going through the reducers, triggering the rerender process) just to get a log of something.

If you feel strongly that this is the approach we should take, I won't block it. But I think we should consider removing these actions and just handling webview commands with regular click handlers, and handle logging that some other way. I think that would reduce complexity.

test/ui/browser/utils.js
```diff
@@ -0,0 +1,37 @@
...
+  }).join(' ');
+
+  return new Promise(resolve => {
+    const inc = (el, attr) => () => el.setAttribute(attr, ~~el.getAttribute(attr) + 1);
+    jsdom.env({
+      html: domText,
+      done(err, window) {
+        const webviews = window.document.querySelectorAll('webview');
+        Array.prototype.forEach.call(webviews, webview => {
+          webview.goBack = inc(webview, 'back-count');
```
linclark: Would it make sense for us to use sinon test spies here? http://sinonjs.org/docs/#spies

jsantell:
Superceded via #358

PR #289, opened 2016-04-28
User a custom source-map-support in order to get meaningful stack tra…

ncalexan:
…ces.

The source-map-support package doesn't appreciate absolute file://
URLs.  I tried to address this in
https://github.com/evanw/node-source-map-support/issues/133 and made
some progress before discovering
https://github.com/evanw/node-source-map-support/pull/126.  I pushed
that work, and a small follow-up, to my local version used here.

With this, I get stack traces that referring to app/.

### DIFF
--- a/app/main/browser.js
+++ b/app/main/browser.js
@@ -13,18 +13,21 @@ specific language governing permissions and limitations under the License.
 */
 
 /* eslint import/imports-first: "off" */
+/* eslint no-console: 0 */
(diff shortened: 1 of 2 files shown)


ncalexan:
@rnewman, @Mossop -- non-standard but very useful locally.  r?

rnewman:
Works for me, AFAICT.

PR #291, opened 2016-04-28
Fetch recently starred pages from storage

rnewman:
This adds a `recentBookmarks` field to state, populates it from storage, makes the bookmark objects richer (title, location, etc.), and thus gives us a working bookmarks menu.


app/services/storage.js
```diff
@@ -375,10 +375,32 @@ export class ProfileStorage {
     return this.collectURLs(await this.db.all(query, [since, limit]));
   }
 
-  async starred(): Promise<[string]> {
+  async getStarredWithOrderByAndLimit(orderClause: string, limit: ?number): Promise<[string]> {
```
ncalexan: Wow, this is rough.  The caller needs to know the join involves `s` and `t`.  Maybe just special case ordering as `newestFirst: boolean`.

package.json
```diff
@@ -53,7 +53,7 @@
     "eslint-config-airbnb": "6.2.0",
     "eslint-plugin-import": "1.4.0",
     "eslint-plugin-react": "4.2.3",
-    "expect": "1.16.0",
+    "expect": "1.18.0",
```
ncalexan: Prefer to explain why we're upgrading now in the commit message.

app/main/browser.js
```diff
@@ -55,7 +55,13 @@ let currentState;
 
 function sendToAllWindows(event: string, args: Object): void {
   console.log(`browser.js: sendToAllWindows ${JSON.stringify(args)}`);
-  store.getState().browserWindows.forEach((id) => {
+  const windows = store.getState().browserWindows;
```
ncalexan: Hmm.  I'd prefer to guarantee that the state always includes an empty list of `browserWindows`, although initial states in Redux seem to be trickier than anticipated.  This is fine for now.

app/main/browser-menu.js
```diff
@@ -135,7 +135,8 @@ const BrowserMenu = {
           label: bookmark.title || bookmark.location,
           click(item, focusedWindow) {
             if (focusedWindow) {
-              focusedWindow.webContents.send('open-bookmark', bookmark);
```
ncalexan: Oh hey, this works!

app/ui/browser/views/browser.jsx
```diff
@@ -177,7 +177,7 @@ function attachIPCRendererListeners(browserView) {
   });
 
   ipcRenderer.on('open-bookmark', (e, bookmark) => {
-    dispatch(actions.createTab(bookmark.url));
+    dispatch(actions.navigatePageTo(browserView.props.currentPage.id, bookmark.location));
```
ncalexan: I'd be happy to standardize on `.location` throughout.

ncalexan:
@rnewman One real request, and then nits.  Also, remember to assign someone (me!) for reviews.

PR #298, opened 2016-04-28
Register tofino:// protocol for internal content pages. Initially, ju…

### LINKED ISSUE #296: Create about:* protocol for content pages

jsantell:
…st using tofino://mozilla now for testing. r=nalexander

Closes #296

### DIFF
--- a/app/main/browser.js
+++ b/app/main/browser.js
@@ -37,24 +37,24 @@ import * as instrument from '../services/instrument';
 import * as profileCommands from '../shared/profile-commands';
 import * as profileDiffs from '../shared/profile-diffs';
 import configureStore from './store/store';
+import registerAboutPages from './about-pages';
 import { ProfileStorage } from '../services/storage';
 import profileCommandHandler from './reducers/profile-command-reducers';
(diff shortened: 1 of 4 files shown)


app/main/browser.js
```diff
@@ -197,6 +186,9 @@ app.on('ready', async function() {
   const appReadyTime = Date.now();
   instrument.event('app', 'READY', 'ms', appReadyTime - appStartupTime);
 
+  // Register `about:*` protocols after app's 'ready' event
```
ncalexan: nit: full sentences.

ncalexan:
Route through @Mossop on tofino:// vs. about://.  (Chrome followed about://; not sure why we'd switch.)

I'm curious about what privileges these `tofino://` pages have.  Are they bound by the cross-origin policy, for example?

I'm fine with this.

jsantell:
`about://` doesn't work for some reason. Don't know why. Can change it later if we want, I don't think `tofino://` is a good name FWIW, but we can start building out bookmarks, home, newtab, etc

And agreed on the privileges front, wanted to get this working so we can see if we can even put electron things in the content (if we wanted to, anyway)

PR #299, opened 2016-04-28
Remove console.log statements from storage service and tests. r=rnewman

jsantell:


test/profile/test-storage.js
```diff
@@ -52,18 +50,14 @@ describe('ProfileStorage.open', () => {
...
       try {
         const parent = tmp.dirSync({}).name;
-        console.log(`Temp root: ${parent}`);
 
         // Ensure that subdirectories will be created if necessary.
         const tempDir = path.join(parent, '/foo/bar/');
         const storage = await ProfileStorage.open(tempDir);
 
         expect(storage instanceof ProfileStorage);
-        const v = await storage.userVersion();
```
rnewman: Switch this to `expect((await storage.userVersion) === 4)`.

app/services/storage.js
```diff
@@ -534,17 +532,14 @@ class ProfileStorageSchemaV4 {
...
       return v;
     }
 
     if (v === 0) {
-      console.log(`Creating storage at version ${this.version}.`);
       return this.create(storage);
     }
 
     if (v < this.version) {
-      console.log(`Updating storage from ${v} to ${this.version}.`);
```
rnewman: I'd like to keep the creation and update branches unless you have strong feelings otherwise. Knowing when DB updates and creation happen is useful, and should only happen on first launch.

rnewman:
r+ modulo those.

app/services/storage.js
jsantell: Would be hidden behind a `if (debug)` flag like similar files be suitable?

test/profile/test-storage.js
jsantell: :+1:

app/services/storage.js
jsantell: Without that, we get a lot of `"Creating storage at version 4."` in tests

rnewman: Yeah, a debug flag is fine by me.

PR #300, opened 2016-04-28
Remove 'Clear History' from menu.

rnewman:
Rebased on top of #297. We don't support clearing history.


jsantell:
r+ for the `app/ui/*` changes not in #297

PR #302, opened 2016-04-28
Less redux

ncalexan:
@bgrins @rnewman this folds the main process Redux profile actions into the profile command handling.

It's still using Redux to drive the loop, but that is easily removed.  I was pushed to this out of frustration when tracking sessions across browsing events; I'll come back to that tomorrow and see if I've improved matters by removing the action boilerplate.

Let me know what y'all think.

PR #303, opened 2016-04-29
Build the browser menu unconditionally

bgrins:
I'm not really sure why this was conditional, but it's causing me to not see the File menu with #302 applied


rnewman:
As discussed on IRC, I think the right solution is to always build the menu when the app launches, and thereafter when Recent Bookmarks changes, not to build the menu every time *anything* changes.

rnewman:
LGTM! We'll find out if this causes flicker or excess work later, I imagine.

PR #304, opened 2016-04-29
Make app state dependencies more explicit

linclark:
@jsantell I'm going to break up the reducer refactor into pieces to make it easier to review. This should also help us avoid stepping on each other's toes too much as we both work on the reducers.

This is the first piece.

### DIFF
--- a/app/ui/browser/reducers/main-reducers.js
+++ b/app/ui/browser/reducers/main-reducers.js
@@ -98,6 +98,27 @@ export default function basic(state = initialState, action) {
   }
 }
 
+export function getPages(state) {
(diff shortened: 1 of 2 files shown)


jsantell:
LGTM! Should these helpers be in reducers, though, if not used as a redux reducer? Even like `./reducers/helpers.js` maybe? I don't feel strongly though

linclark:
Good idea. We will probably have a lot more, so we probably don't want to put it in the same file. Looking back at the Redux docs for selectors, it looks like that uses a `selectors/` dir.

I'll do that as part of the follow ups to this one.

PR #313, opened 2016-04-29
Web Driver utilities and tests

jsantell:


jsantell:
Few things here:

* New homepage is `tofino://mozilla`, more predictable, faster in tests
* Page loading state is now an enum and is renderered as a data- property on the page so we can inspect for tests
* Wrapper around our driver, similar to [Brave's](https://github.com/brave/browser-laptop/blob/900a4da54794a803612352f196f0d4be80baa454/test/lib/brave.js) for testing -- we can add common utilities here to make writing e2e way easier.
* Static fixtures server for tests instead of mozilla.org

package.json
```diff
@@ -54,6 +54,7 @@
     "eslint-plugin-import": "1.4.0",
     "eslint-plugin-react": "4.2.3",
     "expect": "1.18.0",
+    "express": "^4.13.4",
```
Mossop: Use a precise version number here.

package.json
```diff
@@ -54,6 +54,7 @@
     "eslint-plugin-import": "1.4.0",
     "eslint-plugin-react": "4.2.3",
     "expect": "1.18.0",
+    "express": "4.13.4",
```
jsantell: Fixed

test/utils/driver.js
```diff
@@ -0,0 +1,97 @@
...
+      .moveToObject('#browser-navbar-back')
+      .leftClick('#browser-navbar-back');
+  },
+
+  navigateForward: async function() {
+    await this.app.client
+      .waitForVisible('#browser-navbar-forward')
+      .moveToObject('#browser-navbar-forward')
+      .leftClick('#browser-navbar-forward');
+  },
```
Mossop: Can we make a generic "clickButton" that does this?

Mossop:
Nice, LGTM

test/utils/driver.js
jsantell: :+1: We wouldn't always have to wait and move, but nice convenience

PR #315, opened 2016-04-29
Turns on 32-bit windows builds on AppVeyor.

### LINKED ISSUE #312: Support 32-bit builds, particularly for Windows 10 Home

Mossop:
Node has an oddity in that it uses "ia32" instead of "x86" as the identifier for
32-bit builds so we have to transplate the platform environment variable
accordingly.

Fixes #312.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,7 @@
-platform: x64
+platform:
+  - x86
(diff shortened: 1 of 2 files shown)


rnewman:
LGTM.

PR #316, opened 2016-04-29
Remove drag and drop functionality from tabs, as we're unsure if we w…

### LINKED ISSUE #284: dragging and dropping tabs isn't working
Open a couple tabs, then try to drag one to the left or right and I see:

tab.jsx:63 Uncaught TypeError: this.props.onDragStart is not a function

Screenshot: 
screenshot 2016-04-27 13 06 16

jsantell:
…ant this, and it's broken currently anyway. Fixes #284. r=bgrins


bgrins:
Works for me

PR #318, opened 2016-05-02
Remove SET_PAGE_ORDER action type. r=jsantell

linclark:
Now that #316 has landed, I don't think we need this code anymore, and it would make the reducer refactor easier if we remove it. @jsantell, what do you think?


jsantell:
There's a handful of logic in the page order reducer, if it looks like we won't be using this feature in whatever first version we're targeting, this sounds good and we can always revive it if we need, r+

PR #322, opened 2016-05-02
Fix search (cmd-F). r=jsantell

linclark:
Currently cmd-F does not show the search box. This fixes that.

### DIFF
--- a/app/ui/browser/views/browser.jsx
+++ b/app/ui/browser/views/browser.jsx
@@ -55,9 +55,9 @@ class BrowserWindow extends Component {
   handleKeyDown(ev) {
     const { dispatch, currentPage } = this.props;
 
-    if (ev.metaKey && ev.key === 'KeyF') { // cmd+f
+    if (ev.metaKey && ev.code === 'KeyF') { // cmd+f
       dispatch(actions.setPageDetails(currentPage.id, { isSearching: true }));
-    } else if (ev.key === 'Escape') { // esc
+    } else if (ev.code === 'Escape') { // esc
       dispatch(actions.setPageDetails(currentPage.id, { isSearching: false }));
     }
   }


jsantell:
In #262, usage of `keyCode` was changed to just use `key` in key event objects -- should the other handlers also use `code` instead of `key`? Either way, how we handle them should be consistent (@bgrins and @victorporof have stronger feelings IIRC between which ones to use)

bgrins:
Ah hm, according to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key `key` should be working, and I know it is for certain cases (like in the URL bar) but seems to not work here for whatever reason.

And it looks like one thing [to be aware of with `code`]( https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) is:

> However, you need to be careful when you reference this value for printable key events. For example, "KeyQ" is set for the "q" key of the QWERTY keyboard layout, the "'" key of the Dvorak keyboard layout and the "a" key of the AZERTY keyboard layout.

Maybe we'll end up needing to store constants for key codes after all

victorporof:
@bgrins The question we should ask ourselves is, do we want the same keyboard shortcut "shapes" everywhere ("thumb an pinky over there, no matter what the keys actually are"), or do we actually care about them being the same keys (as in, e.g., "these combinations of letters and command keys") everywhere? Maybe good arguments can be brought for both approaches. I'd stick to the classic approach and have keyboard shortcuts be consistent across different keyboards, no matter the layout. That'd mean hard coding key codes, which is sad.

bgrins:
> I'd stick to the classic approach and have keyboard shortcuts be consistent across different keyboards, no matter the layout. That'd mean hard coding key codes, which is sad.

I agree

victorporof:
@linclark Are you still working on this? I can take over if you want me to.

victorporof:
I'll just land this for now, we need a working searchbox. We can settle the keyCode debate later.

PR #324, opened 2016-05-02
Select a completion when pressing Enter. Fixes #309. r=jsantell

### LINKED ISSUE #309: Modify URL bar text when navigating through completions
Follow on to #236

STR:
Search for 'mo' in the URL bar to show 'https://www.mozilla.org/en-US/'
Press down

Expected:
The URL bar becomes filled with 'https://www.mozilla.org/en-US/'

Note, this interaction is complex.  Possible first step is to make pressing 'enter' on a highlighted search result to fill the URL bar, instead of arrow navigation filling it.

bgrins:
@jsantell 

Still working on a test (looks like we don't have any test yet for the location component)

### DIFF
--- a/app/ui/browser/views/navbar/location.jsx
+++ b/app/ui/browser/views/navbar/location.jsx
@@ -139,6 +139,14 @@ class Location extends Component {
     return 'glyph-bookmark-hollow-16.svg';
   }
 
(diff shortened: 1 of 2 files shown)


jsantell:
Nice, lgtm!

PR #325, opened 2016-05-02
Do not clear out Page's title on load, as we don't always have a new …

jsantell:
…title for a page, like when using HTML5 history APIs. r=bgrins


bgrins:
Nice, r+

bgrins:
Looks like there's an error on CI:

1) navigation updates display title when navigating via HTML5 history:
     Error: element (.active-browser-page[data-page-state="loaded"]) still not visible after 5000ms

jsantell:
I think that happens intermittently -- on any wait command for webdriver unfortunately :sigh:

Issue #282, opened 2016-04-27
Move frequently changing values out of the page object


linclark:
![render-tree-statustext](https://cloud.githubusercontent.com/assets/105464/14864372/9aa20b46-0c87-11e6-807f-50d665a0180b.png)
![render-tree-usertyped](https://cloud.githubusercontent.com/assets/105464/14864371/9aa119b6-0c87-11e6-982e-71cd0670779a.png)

Discussed in the meeting today. Currently, these values are `statusText` and `userTyped`.

linclark:
This has been partially reintroduced. We are rerendering for each character pressed in the location bar. Will discuss at the meeting.

PR #327, opened 2016-05-03
Split main-reducers.js into pages reducer and UI state reducer

### LINKED ISSUE #283: Break up main-reducers.js and use combineReducers

linclark:
r? @jsantell 

Fixes #283. It also helps with #282 by moving `userTyped` off of the page object.

Two things I wanted to call out:

- I named the new reducer `ui-state`. Open to other suggestions for it.
- I removed two tests that were checking the model. My reasoning was that these would need to be updated quite frequently (which I'm not sure they were being) and that they wouldn't catch many bugs that weren't caught by the other tests. I'm open to reverting that, though.


app/ui/browser/model/index.js
```diff
@@ -65,3 +51,20 @@ export class Page extends Immutable.Record({
...
+  recentBookmarks: Immutable.List(),
+  completions: Immutable.Map(),
+});
+
+/**
+ * Frequently changing UI state.
+ */
+export const UIState = Immutable.Record({
+  // What the user has typed into the location bar, stored by page id
+  userTypedLocation: Immutable.Map(),
```
jsantell: nit: Maybe `pagesAreaVisible`, `currentPageIndex` would make more sense in UIState rather than Page state

jsantell:
Looks good! Maybe `ui-state` can just be `ui`?

w/r/t removal of model test, we shouldn't need to update too often once things settle down, but for the most part is just testing immutable -- only thing that I'd want to keep is testing that Pages get an ID when they're created, but pretty sure other tests test that implicitly

app/ui/browser/model/index.js
linclark: I'm glad you said that because I agree. I decided not to attempt it here because it adds a little complication and I wanted to split it with the smallest amount of extraneous changes, but since you agree I'll work on that in a follow up.

linclark:
> Looks good! Maybe ui-state can just be ui?

Another thing I was thinking was `chrome`? Maybe I'll keep it as uiState for now and address in a follow up.

For the model test, that makes sense. So we would test anything where we're setting a more complicated default. I reinstated the test and added a check for the ID creation.

PR #329, opened 2016-05-03
Support downloading a custom build of electron.

Mossop:
Drops the electron-prebuilt dependency and instead manually checks the installed
electron version and if incorrect downloads with electron-download. We cache
the download locally to tofino since the global cache could affect other
applications on a developer's machine. Currently this doesn't change where
electron is downloaded from but once we have builds available we can just set
the electron-download options to point to them.

### DIFF
--- a/build/task-build-deps.js
+++ b/build/task-build-deps.js
@@ -5,9 +5,36 @@
 
 import path from 'path';
 import fs from 'fs-promise';
+import os from 'os';
 
-import { BUILD_CONFIG_JSON } from './task-config-builder';
+import download from 'electron-download';
+import unzip from 'extract-zip';
+import thenify from 'thenify';
+
 import * as BuildUtils from './utils';
+
+async function downloadElectron() {
+  const tmpDir = path.join(os.tmpdir(), 'tofino-tmp');
+  await fs.mkdirs(tmpDir);
+
+  try {
+    const zipPath = await thenify(download)(BuildUtils.getDownloadOptions());
+
+    await thenify(unzip)(zipPath, { dir: tmpDir });
+
+    // Some tools like electron-rebuild rely on this to find the executable
+    await fs.writeFile(path.join(tmpDir, 'path.txt'),
+                       BuildUtils.ELECTRON_EXECUTABLE[os.platform()]);
+
+    const targetDir = BuildUtils.getElectronRoot();
+    await fs.remove(targetDir);
+    await fs.move(tmpDir, targetDir);
+  } finally {
+    if (await fs.exists(tmpDir)) {
+      fs.remove(tmpDir);
+    }
+  }
+}
 
 async function findNativeModules() {
   const nodeModules = path.join(__dirname, '..', 'node_modules');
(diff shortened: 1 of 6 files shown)


build/task-build-deps.js
```diff
@@ -5,10 +5,45 @@
...
+import os from 'os';
+
+import download from 'electron-download';
+import unzip from 'extract-zip';
 
-import { BUILD_CONFIG_JSON } from './task-config-builder';
 import * as BuildUtils from './utils';
 
+async function downloadElectron() {
+  const targetDir = BuildUtils.getElectronRoot();
```
ncalexan: Consider deleting only after we download and unpack successfully.

build/task-build-deps.js
```diff
@@ -43,19 +78,33 @@ async function findNativeModules() {
...
   try {
-    existingConfig = await fs.readJson(BUILD_CONFIG_JSON);
+    existingConfig = await fs.readJson(BuildUtils.getBuildConfigFile());
   } catch (e) {
     // Missing files mean we rebuild
   }
 
+  const electron = BuildUtils.getManifest()._electron;
+  try {
+    if (electron.version !== BuildUtils.getElectronVersion()) {
```
ncalexan: Let's not do this, or at least not outside of automation.  There's no sense pulling the same large ZIP file twice in order to discover there's something funky with the compression or whatever.

ncalexan: And by this, I mean: let's not re-download.

build/task-build-deps.js
```diff
@@ -5,10 +5,45 @@
...
 
-import { BUILD_CONFIG_JSON } from './task-config-builder';
 import * as BuildUtils from './utils';
 
+async function downloadElectron() {
+  const targetDir = BuildUtils.getElectronRoot();
+  await fs.remove(targetDir);
+  await fs.mkdirs(targetDir);
+
+  const zipPath = await new Promise((resolve, reject) => {
```
ncalexan: Consider using `thenify` to make this cleaner.

ncalexan:
Code looks good to me; just some small suggestions.  I have not tested this locally.

build/task-build-deps.js
Mossop: I don't understand, this is how developers get updated to the right version of electron on their machine.If they don't have the right electron what do we do instead?

ncalexan: Sorry, I'm still not clear.  I mean, don't catch and re-download on error.  If there's a bad ZIP, don't redownload it!  If the "error case" includes bad versions (which I don't think it does, since you're testing the version above) that's very misleading and should be a separate `catch` case at least.

build/task-build-deps.js
```diff
@@ -5,10 +5,37 @@
 
 import path from 'path';
 import fs from 'fs-promise';
+import os from 'os';
+
+import download from 'electron-download';
```
ncalexan: Slight preference for `thenify` at import time, but whatevs.

ncalexan:
Looks good!

PR #332, opened 2016-05-03
Move statusText off of the page object. r=jsantell

### LINKED ISSUE #282: Move frequently changing values out of the page object

linclark:
Fixes #282 

A simple way to see how this changes what gets rendered:

1. Add a debugger statement in the render function in page.jsx
2. `npm start`
3. go to twitter.com
4. get close to a tweet link
5. open dev tools
6. hover over link

Before, you would hit the debugger statement. Now, you will not because the page isn't being rerendered.


app/ui/browser/views/page/status.jsx
```diff
@@ -51,6 +55,13 @@ Status.displayName = 'Status';
 
 Status.propTypes = {
   page: PropTypes.object.isRequired,
+  statusText: PropTypes.string.isRequired,
 };
 
-export default Status;
+function mapStateToProps(state) {
+  return {
+    statusText: getStatusText(state),
```
jsantell: don't we also need to map `page` here?

jsantell:
nice! r+, some linting failures on CI, and if `page` needs to be added to `mapStateToProps`

app/ui/browser/views/page/status.jsx
linclark: I wanted to ask about this... since we're only showing one status text at the moment, do we want to keep track of all of them? It seems like we could get away with only storing one since the user will only be hovering over one link at a time. But if we want to show multiple, then we can definitely add a map here.

jsantell: I think I just misunderstood this, we should only need one status text at a time, if `page` is already handled without mapping, my understanding is complete and this makes sense to me now

PR #351, opened 2016-05-04
Completely remove drag-and-drop components from tab that are throwing…

### LINKED ISSUE #336: TabBar Error From Console
There are 4 `TabBar` errors displaying in the console, seen when opening DevTools.

tabbarerror

jsantell:
… validation errors. r=linclark

Fixes #336


linclark:
This looks good. One thing I noticed, there is a `WebkitAppRegion: 'drag'` in `navbar.jsx`... is that anything that should be changed?

Other than that question, LGTM!

jsantell:
The `WebkitAppRegion` property is for being able to drag the full window around in the OS, so that should stay in this case, I think.

Thanks!

PR #354, opened 2016-05-04
Fix #335 - Enable 'File' menu item so app is still usable when window is closed. r=jsantell

### LINKED ISSUE #335: OS X: Closing Window Leaves App Useless
Closing the browser window on OS X keeps the app open but without a way to open a new window, thus leaving the app in a weird limbo.

darkwing:
There's no reason to remove the `File` menu item when there are no windows open -- it merely makes the app unusable.

This is my proposal:
- Keep the `File` menu item visible at all times
- Open new window if user clicks `New Window` submenu item
- Open new window if user clicks `New Tab` submenu item and there is no window open.

Happy to write a test if this functionality proposal hits the mark.

### DIFF
--- a/app/main/browser-menu.js
+++ b/app/main/browser-menu.js
@@ -18,6 +18,15 @@ const Menu = electron.Menu;
 
 const HELP_LINK = 'https://github.com/mozilla/tofino';
 
+const openNewWindow = () => {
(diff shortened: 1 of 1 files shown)


jsantell:
I think this sounds good, I'll review it hopefully by end of the day!

ncalexan:
lgtm.

PR #358, opened 2016-05-04
Control webviews with an event emitter helper rather than explicit st…

jsantell:
…ate changes with redux actions. r=linclark

@linclark @ncalexan So this is a bit what I was talking about. I think it'll make more sense as a side effect once we have explicit state changes from history/navigation events.


app/ui/browser/actions/external.js
```diff
@@ -11,9 +11,9 @@ specific language governing permissions and limitations under the License.
 */
 
 import {
-  createTab, setUserTypedLocation, closeTab, navigatePageTo,
+  createTab, setUserTypedLocation, closeTab,
```
ncalexan: Prefer `import *`.  Less to update, easier to grep for, and better WebStorm linking.

app/ui/browser/reducers/pages.js
```diff
@@ -55,19 +52,6 @@ export default function basic(state = initialState, action) {
...
       return closeTab(state, action.pageId);
 
-    case types.NAVIGATE_PAGE_BACK:
-      return navigatePageBack(state, action.pageId);
-
-    case types.NAVIGATE_PAGE_FORWARD:
-      return navigatePageForward(state, action.pageId);
-
-    case types.NAVIGATE_PAGE_REFRESH:
-    case types.IPC_COMMAND_PAGE_REFRESH:
```
ncalexan: In some way, we need to be able to trigger a refresh from the main process.  Perhaps you re-instate this later, or perhaps that's follow-up.

app/ui/browser/webview-controller.js
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+import assert from 'assert';
+import { EventEmitter } from 'events';
+import { isUUID } from './browser-util';
+
+const getPageIndexById = (pages, id) => pages.findIndex(p => p.id === id);
+
+/**
+ * A utility class that extends EventEmitter to receive page IDs and
```
ncalexan: Comment on how this fits into the larger architecture.  I expected one controller per webview, but in fact there's a single controller for all the webviews.

app/ui/browser/webview-controller.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  navigateBack(pageId) {
+    const pages = this.getPages();
+    const index = getPageIndexById(pages, pageId);
+    assert(index >= 0, `Page ${pageId} not found in current state.`);
+    assert(isUUID(pageId), `Invalid page id: ${pageId}`);
+    assert(pages.get(index).canGoBack, 'Page cannot go back.');
+
+    this.emit('navigate-back', pageId);
```
ncalexan: Consider also emitting `navigate-back-${pageId}`, to make it easier to listen to just one webview's message.  (And then wrapping `on` to hide that detail.)

app/ui/browser/views/page/page.jsx
```diff
@@ -50,49 +50,24 @@ const WEB_VIEW_CHROME_COLLAPSED_STYLE = Style.registerStyle({
...
-   *
-   * @TODO This queue grows indefinitely. We should drain it somehow in a
-   * Reduxy way.
-   */
-  componentDidUpdate({ page }) {
-    const currentPage = this.props.page;
-    if (page.commands.size !== currentPage.commands.size) {
-      for (let i = page.commands.size; i < currentPage.commands.size; i++) {
-        this.executeCommand(currentPage.commands.get(i));
+    const filter = fn => (id, ...args) => {
```
ncalexan: This surprised me, but I know see the pattern.  Consider making `on` take an optional `pageId` to move this filtering into the controller.  Otherwise, there's a lot of filtering happening for most webview events.

app/ui/browser/actions/external.js
```diff
@@ -101,7 +101,9 @@ export function menuLocationContext(input, dispatch) {
         pageIndex: -1,
         userTyped: loc,
       }));
-      dispatch(navigatePageTo(-1, fixURL(loc)));
+
+      // TODO this doesn't work anyway, we have to pass in webViewController
+      // dispatch(navigatePageTo(-1, fixURL(loc)));
```
linclark: I'd be tempted to just delete the feature if it isn't working rather than leave a todo, but go with whatever you think.

ncalexan:
I posted some nits and code style comments inline.  This looks good overall, but I don't feel I really understand the benefit of this approach.  The diff churn looks about equal (similar + lines to - lines), after you forget the test files, so it's not syntactically simpler.  If you're happy with it being semantically simpler, rock on!

ncalexan:
Oooh, speaking of semantics: it would be instructive to see a follow-up patch to change how page title's are surfaced, since that would flow through the controller and out into a Redux action (and then out to the UA service).

app/ui/browser/reducers/pages.js
jsantell: Yeah it's handled in browser.jsx

app/ui/browser/views/page/page.jsx
jsantell: I considered that, but still wanted it possibly to listen to all events for tests. But that can be done by emitting both `navigate-back` and `navigate-back-${id}` if we wanted, so will do that

app/ui/browser/actions/external.js
jsantell: None of these are tested and probably don't work, going to tackle this next so it's just a list for me heh

jsantell:
@ncalexan this breaks out the webview interactions out of the redux loop; a page title changing is unaffected, this just no longer uses actions/action semantics when they aren't actions, and trigger empty reducers

linclark:
I like it. r+ if CI comes back green

PR #362, opened 2016-05-04
Reset selected autocomplete index when the typed input changes

bgrins:
r? @jsantell.

This fixes a situation where you have hovered / navigated to a result, then change the input so that there aren't as many results.  i.e. you can end up in a situation where the focused result = 4 but the number of results = 2.

We could possibly move the entire focusedResultIndex machinery up to redux but I'm not sure if that's what we'll want long term, so this is a quick and easy fix to improve the UX.


jsantell:
This makes sense and looks good, r+

Issue #337, opened 2016-05-04
Tab Bar should always be visible


jsantell:
Tracking for MVP milestone.

rnewman:
Fixed by #363.

PR #363, opened 2016-05-05
Always show the Pages bar. (#337)

rnewman:


rnewman:
As a follow-up for this: the pages bar overlays onto web content, which is now wrong.

rnewman:
This seems like a gloriously awful hack to me, but it seems to work, and it's what we do already…
I have no idea where `27` comes from, but the two bars are the same height when compressed, so I used that.

rnewman:
Note that I very hackily implemented an English plural form for the tab counter. This causes some left/right shifting which we probably want to fix, but I'm happy to just get rid of "1 Pages".

PR #364, opened 2016-05-05
Increase default homepage horizontal padding

darkwing:
The homepage text gets too close to the left and right browser edges, at least for my taste.  This increases the spacing, making things look a bit more spacious.

screenshot at may 04 20-12-05


jsantell:
lgtm!

PR #367, opened 2016-05-05
Refactor browser menu creation to make it easier to update as well as…

jsantell:
… prepare to listen to IPC calls for testing the accelerators. r=nalexander


ncalexan:
I see locally that the App and File menu are missing separator lines.  (There are separator spaces.)  Not all menus are missing separators, though.  (I can screenshot if necessary.)

In addition, after closing the last window, I have only the App and Help menu, and Alt+N and Alt+T no longer do anything.  I think @darkwing did some nice work to get New Tab and New Window working in this situation; I'd prefer not to lose it.

Code looks fine.

ncalexan:
While you're refactoring, consider making these templates dynamic.  The parameter would be a locale.  (I didn't do this refactor earlier since it's awkward, but you're re-working a good deal of stuff here.)

jsantell:
Ah yes, I need to add 'file', 'edit', and 'window' to the window-less menu set.

jsantell:
Fixed separator issue.

Once we figure out what our l10n story is, making these dynamic will be straight forward, but no point yet because no idea what kind of l10n data we're looking at.

PR #368, opened 2016-05-05
Track session IDs and ancestor IDs for tabs.

ncalexan:
See commit comments for process notes.


app/ui/browser/actions/main-actions.js
```diff
@@ -39,15 +39,19 @@ export function setStatusText(text) {
...
     // Only send request to profile service if there's a non-empty input.
     // Empty input could happen if a page finishes loading and the userTyped
     // state is going to be reset.
     if (payload.text) {
-      ipcRenderer.send('profile-command', profileCommands.setUserTypedLocation(payload.text));
+      profileCommands.request(profileCommands.setUserTypedLocation(payload.text))
+        .then(({ text, completionList }) => {
+          dispatch(profileDiffs.completions(text, completionList));
+        })
+        .catch(console.warn); // Ignore failures.  Not much to be done here.
```
bgrins: I don't think this will work how you want it to (although that's changing as of https://github.com/whatwg/console/issues/3 I don't think electron has caught up to it).  `console.warn.bind(console)` will do the trick

ncalexan: I think I'm just going to drop this entirely.  It *should* make `eslint` complain, but for reasons I don't understand it does in `app` but not in `app/ui`.

app/main/profile-command-handler.js
```diff
@@ -40,16 +45,35 @@ export async function handler(
       .set('recentBookmarks', new Immutable.List(recentBookmarks));
   }
 
+  async function startSession(scope: number, payload: Object = {}): Promise<void> {
+    const { ancestorId } = payload;
+    const sessionId = await storage.startSession(scope, ancestorId);
```
bgrins: Do we want error handling for this or is it fine to assume for now that the storage will never reject?  Looks like that's what we are doing anyway on app startup so it's probably fine.  It's just that the respond callback will never fire which might leave the UI in an odd state

app/main/profile-command-handler.js
```diff
@@ -72,23 +93,21 @@ export async function handler(
...
       break;
 
     case profileCommandTypes.DID_SET_USER_TYPED_LOCATION:
-      try {
-        const completions = await storage.visitedMatches(payload.text);
-        return userAgent.set('locations',
-          userAgent.locations.set(payload.text, new Immutable.List(completions)));
-      } catch (e) {
-        console.log(e); // TODO: do more than ignore failure.
+      if (respond) {
```
bgrins: This would be a weird scenario if the frontend sent this action without a respond callback.  I don't know that it's worth throwing, but maybe at least warning (unless if you can think of a reason this should happend)

app/ui/browser/index.jsx
```diff
@@ -18,18 +18,26 @@ import { ipcRenderer } from '../../shared/electron';
...
 const initialState = rootReducer(undefined, { type: null });
-initialState.profile = initialState.profile.withMutations(mut => {
-  mut.set('recentBookmarks', new Immutable.List(initialProfileState.recentBookmarks))
-     .set('bookmarks', new Immutable.Set(initialProfileState.bookmarks));
-});
 
 const store = configureStore(initialState);
 
+// Before rendering, add a fresh tab and prepare the cached profile data.  This is leading toward a
+// session restore model) where the main process can instantiate a mostly-formed browser window.
```
bgrins: extra )

app/ui/browser/reducers/pages.js
```diff
@@ -101,24 +126,38 @@ function closeTab(state, pageId) {
...
+  if (ancestorId && ancestorId >= 0) {
+    ancestorIndex = getPageIndexBySessionId(state, ancestorId);
   }
 
   return state.withMutations(mut => {
     mut.update('pages', pages => pages.delete(pageIndex));
 
+    const current = state.currentPageIndex;
+
+    // If we're currently looking at the closed tab and it has an ancestor that is still open,
```
bgrins: that's nice!  is there an existing place to add a test for this?

test/ui/browser/actions/test-close-tab.js
```diff
@@ -60,18 +76,37 @@ describe('Action - CLOSE_TAB', () => {
...
     dispatch(actions.closeTab(ids.get(3)));
+
     expect(getPages().size).toEqual(3);
     expect(getPages().get(2).location).toEqual('https://moz.org/2');
     expect(getCurrentPageIndex()).toEqual(2,
       'CLOSE_TAB on last selected tab selects the previous tab');
   });
 
-  it('Destroying the last tab should reset state', function() {
+  it('Should destroy a selected tab and select ancestor when it has an ancestor', function() {
```
ncalexan: @bgrins -- test for ancestor selection is here.  I'll add a test to ensure that this doesn't fail when the ancestor has been closed.  We could also ask the UA service for *all* ancestors, and try to find the most recent one...

bgrins:
Looks good to me

app/main/profile-command-handler.js
ncalexan: My current thinking is that these storage functions move to an HTTP or WS service and every API endpoint returns something, which might be an error.  That is, the User Agent service looks more and more like a regular web service.  That's what the UI (the web platform) is good at working with.

app/main/profile-command-handler.js
ncalexan: I agree.  I'm just quieting Flow's null checking here.  This gets more uniform when the function return is just for the response, and not for the main process state updates.  Soon!

rnewman:
@ncalexan: is this still relevant?

ncalexan:
Landed!

Issue #346, opened 2016-05-04
+W to Close Window


jsantell:
Tracking for MVP milestone.
Should not be considered complete without tests.

PR #369, opened 2016-05-05
Close window hotkey

ncalexan:
Addresses #346 but doesn't have tests.  @jsantell, relevant to your interests.


jsantell:
Looks good, but just looking at commit 44850b1

Also, good to point out the hotkey testing in #370 only covers menu hotkeys -- I'll look at adding the localshortcut-style hotkeys into the mix as well and add tests for it

jsantell:
Fixed in 25ac25d102caa69dcfe8bb9c59cf658c77832fba

PR #370, opened 2016-05-06
Menu hotkey testing

jsantell:
FINALLY!!

### DIFF
--- a/test/utils/driver.js
+++ b/test/utils/driver.js
@@ -72,6 +72,40 @@ const Driver = {
     this.app = this.port = this.fixturesURL = this._stopServer = this._bwHandle = null;
   },
 
+  /**
+   * Fetches the redux state of the first browser window and returns
+   * it as a plain JS object (de-immutable-ified).
+   *
+   * @return Promise<Object>
+   */
+  getReduxState: async function() {
+    await this.setTargetToBrowserWindow();
+
+    return await this.client.execute(() =>
+      JSON.stringify(require('electron').ipcRenderer.store.getState()))
+      .then(res => JSON.parse(res.value));
+  },
+
+  /**
+   * Polls the redux state of the first browser window
+   * until the passed in predicate, which receives the state as a
+   * plain JS Object, returns a truthy value, or resolves to a truthy value.
+   *
+   * @param {Function} predicate
+   * @return Promise
+   */
+  waitUntilReduxState: async function(predicate) {
+    let result = false;
+    do {
+      const state = await this.getReduxState();
+      result = await predicate(state);
+    } while (!result);
+  },
+
+  ipcSendToRenderer(channel, ...args) {
+    this.app.electron.ipcRenderer.emit(channel, {}, ...args);
+  },
+
   ipcSendToMain(channel, ...args) {
     this.app.electron.ipcRenderer.send(channel, ...args);
   },
--- a/test/utils/driver.js
+++ b/test/utils/driver.js
@@ -56,6 +56,7 @@ const Driver = {
     this._stopServer = stop;
     this.port = port;
     this.fixturesURL = `http://localhost:${this.port}`;
+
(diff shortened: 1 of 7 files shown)


ncalexan:
I have two thoughts about this PR:

1. I think we should work out how to add this type of test-only message up front, before it gets hard to do so.  Can we use Spectron to trigger the addition of the listener?  Or do we need to set an environment variable or a command line option?

2. I think it's not wise to test this type of interaction using the UI.  I tried to test link clicking (option-click, `target=_blank`) in this style and it's even harder to do using the UI.  I rigged up a Web Socket server to let the test ask all BWs for their state, which made for *much* better testing.  However, I think the better thing to do is find a way to add "test middleware" to the UI store which reflects to the test (which then captures) the *actions* (and not the state) flowing through the UI Redux loop, so that our tests can assert "interaction X produced Redux action Y".  This is better than "interaction X produced Redux state Y", since we then have freedom to change the details of what actions do.  And it's much better than "interaction X produced UI DOM Z", which is just frustrating to work with.

In many ways, we have an opportunity to tap the message stream of the browser at a very low level.  Imagine how much easier Firefox testing would be if we could do this!

jsantell:
> I have two thoughts about this PR:
> 
> I think we should work out how to add this type of test-only message up front, before it gets hard to do so. Can we use Spectron to trigger the addition of the listener? Or do we need to set an environment variable or a command line option?

If we can determine if we're in a testing environment in the main process, we can do that; but that depends on kinds of tests we're running. I don't think this is implemented yet AFAIK?


> I think it's not wise to test this type of interaction using the UI. I tried to test link clicking (option-click, target=_blank) in this style and it's even harder to do using the UI. I rigged up a Web Socket server to let the test ask all BWs for their state, which made for much better testing. However, I think the better thing to do is find a way to add "test middleware" to the UI store which reflects to the test (which then captures) the actions (and not the state) flowing through the UI Redux loop, so that our tests can assert "interaction X produced Redux action Y". This is better than "interaction X produced Redux state Y", since we then have freedom to change the details of what actions do. And it's much better than "interaction X produced UI DOM Z", which is just frustrating to work with.

We do this for action tests -- tap into action/state cycles. There's no way to use electron outside of spectron. The UI interaction here is just closing tabs (which I agree isn't great, but neither is having hotkeys break every week) and isn't looking deep into the browser chrome state. To deal with frustrating of windows being indistinguishable in webdriver, maybe we could expose an ipc command in a browser window to return state so that tests can use it. Also, dispatching actions via IPC (similar to main->renderer communication) to do things like close tabs -- is that what you're suggesting?

test/utils/driver.js
```diff
@@ -72,6 +72,40 @@ const Driver = {
...
+   * Fetches the redux state of the first browser window and returns
+   * it as a plain JS object (de-immutable-ified).
+   *
+   * @return Promise<Object>
+   */
+  getReduxState: async function() {
+    await this.setTargetToBrowserWindow();
+
+    return await this.client.execute(() =>
+      JSON.stringify(require('electron').ipcRenderer.store.getState()))
```
ncalexan: It's worth communicating that the Redux state does not transmit via JSON cleanly.  We will want `transit` and `transit-immutable` to rehydrate meaningful record types.

test/utils/driver.js
```diff
@@ -72,6 +72,40 @@ const Driver = {
...
+
+  /**
+   * Polls the redux state of the first browser window
+   * until the passed in predicate, which receives the state as a
+   * plain JS Object, returns a truthy value, or resolves to a truthy value.
+   *
+   * @param {Function} predicate
+   * @return Promise
+   */
+  waitUntilReduxState: async function(predicate) {
```
ncalexan: I'm fine with this approach.  I'm not a huge fan of polling in this way, but c'est la vie.

PR #373, opened 2016-05-06
Use `redux-immutable` to have the aggregate UI state be an immutable …

ncalexan:
…Record.

This is just hygiene, consistency, and future-looking uniformization.
Mostly I was irritated that I couldn't just .toJS() the UI state!

There's a bit of hoop-jumping to use a new Record type rather than an
immutable Map.  It's worth it to expose `.member`, rather than requiring
`.get('member')` everywhere.  I hope to make this type checked in the
future.

redux-immutable is quite light and has no further dependencies.  I see
no reason to write our own equivalent, or to fold it into our package
directly.

### DIFF
--- a/app/ui/browser/index.jsx
+++ b/app/ui/browser/index.jsx
@@ -17,16 +17,10 @@ import { ipcRenderer } from '../../shared/electron';
 
 import App from './views/app';
 import configureStore from './store/store';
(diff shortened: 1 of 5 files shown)


Mossop:
LGTM, fix the lint failure before landing.

PR #385, opened 2016-05-09
Set input value when userTypedLocation changes externally

### LINKED ISSUE #356: Address Bar Context Menu - "Paste" and "Paste and Go" Don't Work
On Mac OS X El Capitan the "Paste" and "Paste and Go" address bar context menu items do nothing -- the address bar stays empty after clicking either item.

bgrins:
Fixes #356.

@jsantell I can't think of a good way to test this.

* Context menu actions - AFAIK we can't trigger the context menus in our tests.  This seems like the kind of thing Flow should help with (the parameter order / types regressed at some point when calling functions that *do* have type annotations), but I couldn't get it to actually complain about much trying.
* The input value setting - this is hooking into componentWillReceiveProps to force the input's value to update if userTypedLocation has changed in some manner other than input being received in the text box.  Shallow rendering doesn't support refs though (https://facebook.github.io/react/docs/test-utils.html#shallow-rendering) so we can't cover it from our unit tests.

### DIFF
--- a/app/ui/browser/views/navbar/location.jsx
+++ b/app/ui/browser/views/navbar/location.jsx
@@ -117,6 +117,15 @@ export class Location extends Component {
     // depending on the eventual UX and if anywhere else would want to change it.
     if (this.props.userTypedLocation !== nextProps.userTypedLocation) {
       this.setState({ focusedResultIndex: -1 });
+
+      // this happens when userTypedLocation changed outside of a 'normal' input
+      // change i.e. 'paste' context menu.  Need to make sure the input is actually set.
+      // No tests for this right now, since shallow rendering doesn't support 'refs'.
+      if (this.refs.input &&
+          nextProps.userTypedLocation &&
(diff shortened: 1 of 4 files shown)


app/ui/browser/views/navbar/location.jsx
```diff
@@ -117,6 +117,15 @@ export class Location extends Component {
     // depending on the eventual UX and if anywhere else would want to change it.
     if (this.props.userTypedLocation !== nextProps.userTypedLocation) {
       this.setState({ focusedResultIndex: -1 });
+
+      // this happens when userTypedLocation changed outside of a 'normal' input
+      // change i.e. 'paste' context menu.  Need to make sure the input is actually set.
+      // No tests for this right now, since shallow rendering doesn't support 'refs'.
```
jsantell: You can use enzyme for non-shallow tests too, which this component should be able to independently test

PR #387, opened 2016-05-09
WIP: Rudimentary text + URL + title searching.

rnewman:


rnewman:
@bgrins, @ncalexan: save some page content from the menu and try this out?

rnewman:
It's expected not to pass lint yet, because we do dangerous things.

app/services/storage.js
```diff
@@ -373,6 +374,77 @@ export class ProfileStorage {
     return out;
   }
 
+  /**
+   * Search for places that match title, URL, or saved content
+   * since the specified timestamp.
+   *
+   * Snippets in the returned results are guaranteed to be HTML-escaped.
```
ncalexan: What does this mean?  That they can be interpolated into HTML as values?  Would other consumers find this irritating?

app/shared/types.js
```diff
@@ -23,3 +23,11 @@ export type ReadabilityResult = {
   length: number,
   excerpt: string,
 };
+
+export type AwesomebarMatch = {
```
ncalexan: I feel this pattern is not helping.  Maybe we can just use ES6 class definitions for this?

app/ui/browser/views/navbar/location.jsx
```diff
@@ -220,17 +227,41 @@ export class Location extends Component {
...
-        }}
-        style={this.state.focusedResultIndex === i ? { background: 'red' } : null}>
-        {completion}</div>
-      ));
+    const renderRow = (completion, i) => {
+      // We get safe, decorated (<b>foo</b>) HTML from the database.
+      const snippet = completion.snippet ? (
+        <div
+          className={LOCATION_BAR_RESULTS_SNIPPET_STYLE}
+          dangerouslySetInnerHTML={{          // eslint-disable-line react/no-danger
```
ncalexan: How can this rule not pun on "middle name"?

ncalexan:
This is looking pretty great!

app/services/storage.js
rnewman: Perhaps another way of phrasing this is that the results are valid HTML fragments.

app/ui/browser/views/navbar/location.jsx
rnewman: I'm waiting for a trolling rule whose name is more than 100 characters long.

app/services/storage.js
rnewman: And yes, the idea is that this API is generating HTML like it's 2007. The alternative is that we return some super awesome result format like:

```
[{t: "… some text containing "},
 {m: "a match"},
 {t: " then some more text."}]
```

app/shared/types.js
rnewman: I assume we'll be ditching this if we get rid of Flow…

PR #398, opened 2016-05-10
Make sure unnecessary modules aren't piling up in package.json

victorporof:
Fixes https://github.com/mozilla/tofino/issues/397

### DIFF
--- a/test/lint/packages.js
+++ b/test/lint/packages.js
@@ -1,0 +1,122 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+import expect from 'expect';
+import fs from 'fs-promise';
+import glob from 'glob';
+import difference from 'lodash/difference';
+
+const all = '/**/';
+const valid = '*.@(js|jsx)';
+
+const paths = [
+  `app${all}${valid}`,
+  `build${all}${valid}`,
+  `shared${all}${valid}`,
+  `test${all}${valid}`,
+  `ui${all}${valid}`,
+];
+
+const ignored = [
+  // Build modules not imported anywhere, but used at the CLI.
+  'electron-rebuild',
+
+  // Test modules not imported anywhere, but used at the CLI.
+  'mocha',
+  'coveralls',
+  'nyc',
+
+  // Modules used only for configuration.
+  'babel-preset-es2015-node5',
+  'babel-preset-react',
+  'babel-preset-stage-2',
+  'babel-eslint',
+  'eslint',
+  'eslint-config-airbnb',
+  'eslint-plugin-flow-vars',
+  'eslint-plugin-import',
+  'eslint-plugin-react',
+];
+
+// We can afford being a little more relaxed with the regexes here
+// since we have very strict eslint rules about whitespace.
+const requires = /require\('([\w-/]+)'\)/g;
+const imports = /import(?:.*?from)?\s+'([\w-/]+)'/g;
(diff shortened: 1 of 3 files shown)


victorporof:
Review ping @Mossop

test/lint/packages.js
```diff
@@ -0,0 +1,122 @@
...
+  'eslint-plugin-react',
+];
+
+// We can afford being a little more relaxed with the regexes here
+// since we have very strict eslint rules about whitespace.
+const requires = /require\('([\w-/]+)'\)/g;
+const imports = /import(?:.*?from)?\s+'([\w-/]+)'/g;
+
+describe('packages', () => {
+  it('should not be orpahned', test(async function() {
```
Mossop: s/orpahned/orphaned/

test/lint/packages.js
```diff
@@ -0,0 +1,122 @@
...
+import glob from 'glob';
+import difference from 'lodash/difference';
+
+const all = '/**/';
+const valid = '*.@(js|jsx)';
+
+const paths = [
+  `app${all}${valid}`,
+  `build${all}${valid}`,
+  `shared${all}${valid}`,
```
Mossop: This directory doesn't exist

test/lint/packages.js
```diff
@@ -0,0 +1,122 @@
...
+  `test${all}${valid}`,
+  `ui${all}${valid}`,
```
Mossop: Nor does this

test/lint/packages.js
```diff
@@ -0,0 +1,122 @@
...
+const requires = /require\('([\w-/]+)'\)/g;
+const imports = /import(?:.*?from)?\s+'([\w-/]+)'/g;
+
+describe('packages', () => {
+  it('should not be orpahned', test(async function() {
+    const manifest = await fs.readJson('package.json');
+    const modules = [
+      ...Object.keys(manifest.dependencies),
+      ...Object.keys(manifest.devDependencies),
+    ];
```
Mossop: It would be good, but could be a follow-up, to verify that app/** only uses modules from dependencies and not devDependencies.

test/lint/packages.js
```diff
@@ -0,0 +1,122 @@
...
+
+    const sources = await globMany(paths);
+    const matches = await regexFiles(sources, requires, imports);
+    const expected = Array.from(new Set([...matches, ...ignored]));
+
+    expect(difference(modules, expected)).toEqual([]);
+  }));
+});
+
+function test(runner) {
```
Mossop: What is this function for?

test/lint/packages.js
victorporof: Love this.

test/lint/packages.js
victorporof: Allows async functions to be passed as callbacks to `it` in a more elegant fashion.

Mossop: Mocha claims to already handle callbacks returning promises which is all an async function is anyway. Is there something broken with that?

victorporof: It's a bit annoying having to try catch every `await` and calling done() myself with the error argument to make the test fail, or with no arguments to make it pass. This utility function handles that for me. I'd actually suggest putting it in shared so more people use it.

Mossop: Oh I see, mocha doesn't consider rejected promises as test failures. That's kinda dumb :(

Mossop:
LGTM

PR #399, opened 2016-05-10
Websockets

ncalexan:
@rnewman, could you comment on the REST API for the UA service in https://github.com/mozilla/tofino/commit/f05d568467fc0b822db18dd2f67efa549e3bdc37.  I didn't spend a second thinking about it but together, we should :)


app/main/user-agent-service.js
```diff
@@ -0,0 +1,220 @@
...
+
+    const errors = req.validationErrors();
+    if (errors) {
+      res.status(401).json(errors);
+      return;
+    }
+
+    const { q } = req.query;
+    const results = yield storage.visitedMatches(q);
+    res.json({ results });
```
rnewman: We're probably going to want the richer types specified in `AwesomebarResult` in my current branch, or at least something more than URLs… unless we only want to support URL completion here.

app/main/browser.js
```diff
@@ -231,48 +127,93 @@ app.on('window-all-closed', () => {
...
   }
 });
 
-ipc.on('instrument-event', (event, args) => {
-  // Until we transpile app/, we can't destructure in the argument list or inline here.
-  instrument.event(args.name, args.method, args.label, args.value);
-});
+async function newBrowserWindow(tabInfo: ?Object) {
+  const bw = await makeBrowserWindow(tabInfo);
+  browserWindowIds.push(bw.id);
```
rnewman: Shouldn't this be returning something useful, like `bw`?

app/main/server.js
```diff
@@ -0,0 +1,10 @@
+import * as service from './user-agent-service';
```
rnewman: License block.

app/shared/profile-diffs.js
```diff
@@ -15,14 +15,14 @@
...
-export function bookmarks(bookmarkSet: [string]) {
+export function bookmarks(bookmarkSet: string[]) {
   return {
     type: profileDiffTypes.BOOKMARKS,
     payload: bookmarkSet,
   };
 }
 
-export function completions(text: string, completionList: [AwesomebarMatch]) {
+export function completions(text: string, completionList: AwesomebarMatch[]) {
```
rnewman: Thanks!

rnewman:
I hate to say "LGTM" on something this substantial, but it does.

bgrins:
I'm happy with the approach here

Issue #383, opened 2016-05-09
+[1-8] should select tabs [1-8]


jsantell:
Hotkeys for selecting the first 8 tabs. Should not be considered without tests.

jsantell:
Fixed in f839ecb225b5b4bd5d37f35d8a1c9fc39437f74f and 7d6517e2863018f26a526c46c9122b7a9282c5d4

PR #401, opened 2016-05-10
Numeral tab selection, CmdOrCtrl+[1-8]

jsantell:


jsantell:
Fixes #383

test/webdriver/test-hotkeys.js
```diff
@@ -45,4 +45,46 @@ describe('hotkeys', function() {
...
   });
+
+  it('CmdOrCtrl+[1-8]: Select Tab', async function() {
+    const MAX = 8;
+    const state = await Driver.getReduxState();
+    const initialPages = state.pages.pages;
+    expect(initialPages.length).toBeGreaterThan(0, 'Have atleast one tab to start.');
+
+    // Open tabs until we have 8
+    for (let i = initialPages.length; i < MAX; i++) {
```
jsantell: This test takes a little long, but shouldn't timeout on the spectron end of things (I think) -- we can also just fire several of these tab creations at once and redux will resolve them in hopefully the same loop to save some time?

app/main/hotkeys.js
```diff
@@ -0,0 +1,63 @@
...
+    shortcut.register(browserWindow, accelerator, handlerFactory(browserWindow));
+  });
+}
+
+/**
+ * To be used in responding to IPC 'synthesize-accelerator' messages, by looking
+ * through hotkeys defined and firing the message to the active browser window if found.
+ *
+ * Does not analyze the accelerator for the 'true' command used, for example,
+ * it just crudly maps "CmdOrCtrl+T" to "CmdOrCtrl+T" -- it would not match
```
bgrins: "crudely"

test/webdriver/test-hotkeys.js
bgrins: I guess let's roll with it as-is and see if we have issues with timeouts

bgrins: Actually since this test isn't about cmdorctrl+t how about we do the `Driver.ipcSendToMain('synthesize-accelerator', 'CmdOrCtrl+T');` inside the loop and then have a single `await Driver.waitUntilReduxState(currentState => currentState.pages.pages.length === MAX);` after the loop?

test/webdriver/test-hotkeys.js
```diff
@@ -45,4 +45,46 @@ describe('hotkeys', function() {
...
+      Driver.ipcSendToMain('synthesize-accelerator', `CmdOrCtrl+${i + 1}`);
+      await Driver.waitUntilReduxState(currentState =>
+        currentState.pages.currentPageIndex === i);
+      expect(true).toEqual(true, `Selected tab index ${i} when firing CmdOrCtrl+${i + 1}`);
+    }
+
+    // Close all tabs except the first
+    for (let i = MAX; i > 1; i--) {
+      Driver.ipcSendToRenderer('close-tab');
+      await Driver.waitUntilReduxState(currentState =>
```
bgrins: Ditto here - I think we don't need to await inside the loop since the final await will put us in the right state (unless if we want to test the 'close-tab' functionality directly here)

app/main/hotkeys.js
jsantell: woops :poop:

test/webdriver/test-hotkeys.js
jsantell: I have a patch for that; doesn't seem to save too much time (if any at all), but every bit counts

jsantell:
Merged in f839ecb225b5b4bd5d37f35d8a1c9fc39437f74f / 7d6517e2863018f26a526c46c9122b7a9282c5d4

PR #403, opened 2016-05-10
Add tests for closing tabs, windows and opening windows. r=bgrins

jsantell:


app/ui/browser/views/browser.jsx
```diff
@@ -170,10 +170,6 @@ function attachIPCRendererListeners(browserView) {
   ipcRenderer.on('profile-diff', (_, args) => dispatch(args));
   ipcRenderer.on('focus-url-bar', () => dispatch(ipcActions.focusURLBar()));
   ipcRenderer.on('new-tab', () => dispatch(actions.createTab()));
-
-  // @TODO main process should be sending an id to close a tab
```
bgrins: Is this TODO resolved?  We are still sending a close-tab event without an ID from the main process via the Window menu

bgrins:
Changes look good to me but please review the comment that's removed and make sure it actually should be fully removed

app/ui/browser/views/browser.jsx
jsantell: Correct, but we don't track that information and need to close a specific tab from the main process at this point since we have a better idea of how this thing is architected (i wrote the initial comment FWIW)

PR #331, opened 2016-05-11
Status text sometimes displays link title

rnewman:
For example, on Twitter rolling over all of the tweet titles will show the title in the lower left corner, not the link.


linclark:
👍

The tests are red on CI, but ran them locally and they work.

victorporof:
@rnewman @linclark should we land this?

PR #415, opened 2016-05-11
Urlbar fixes

bgrins:
This has fixes for:
* Initial state (previously it would show an empty value on the home page)
* Multiple tabs (now the value is set when switching between tabs)

It also includes a simple addition to the navigation e2e test to make sure the value gets set

r? @jsantell

### DIFF
--- a/app/ui/browser/views/navbar/location.jsx
+++ b/app/ui/browser/views/navbar/location.jsx
@@ -112,6 +112,7 @@ export class Location extends Component {
 
   componentDidMount() {
     this.props.ipcRenderer.on('focus-urlbar', () => this.refs.input.select());
+    this.setInputValue(this.getRenderURL());
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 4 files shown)


app/ui/browser/views/navbar/location.jsx
```diff
@@ -293,12 +298,12 @@ export class Location extends Component {
           </div>
           <input id="urlbar-input"
             className={INPUT_BAR_STYLE}
-            hidden={!this.state.showURLBar}
+            style={{
```
jsantell: Should add a comment explaining why we have to keep this displayed

app/ui/browser/views/navbar/location.jsx
```diff
@@ -140,6 +132,19 @@ export class Location extends Component {
...
+    // this happens when userTypedLocation changed outside of a 'normal' input
+    // change i.e. 'paste' context menu.  Need to make sure the input is actually set.
+    // No tests for this right now, since shallow rendering doesn't support 'refs'.
+    const nextLocation = this.getVisibleLocation();
+    if (this.refs.input && nextLocation && this.refs.input.value !== nextLocation) {
+      this.setInputValue(nextLocation);
+    }
+  }
+
+  getVisibleLocation() {
```
jsantell: both `visible` and `location` are vague in this context -- maybe something like `getRenderURL`, or atleast a comment

jsantell:
r+ with comments addressed

PR #417, opened 2016-05-12
Initial stab at fetching bookmarks within content.

rnewman:
This selectively adds a `_browser` object to `tofino` pages. We need to hook this up to actually fetch bookmarks, and then make the page look good.

Oh, and `<a>` inside these pages doesn't work with `target='_new'`…

### DIFF
--- a/app/content/preload/content.js
+++ b/app/content/preload/content.js
@@ -20,7 +20,6 @@ import { api } from '../../ui/browser/user-agent';
 window._readerify = readerify.bind(null, Readability);
 
 if (document.location.protocol === 'tofino:') {
-  console.log('Adding _browser.');
   window._browser = {
     async visited(limit) {
       const res = await api(`/visits?limit=${limit}`);
--- a/app/content/preload/content.js
+++ b/app/content/preload/content.js
@@ -22,6 +22,15 @@ window._readerify = readerify.bind(null, Readability);
 if (document.location.protocol === 'tofino:') {
   console.log('Adding _browser.');
   window._browser = {
+    async visited(limit) {
+      const res = await api(`/visits?limit=${limit}`);
+      if (res.ok) {
+        const body = await res.json();
+        return body.pages;
+      }
+      return [];
+    },
+
     // We use `recentStars` instead of `stars` because the latter only
     // gives us URLs, and we want much more.
     async recentStars(limit) {
--- a/app/content/preload/content.js
+++ b/app/content/preload/content.js
@@ -15,5 +15,22 @@ import './context-menu';
 import { readerify } from './reader';
 import Readability from './readability';
 import './scroll';
+import { api } from '../../ui/browser/user-agent';
 
 window._readerify = readerify.bind(null, Readability);
+
+if (document.location.protocol === 'tofino:') {
+  console.log('Adding _browser.');
+  window._browser = {
+    // We use `recentStars` instead of `stars` because the latter only
+    // gives us URLs, and we want much more.
+    async recentStars(limit) {
+      const res = await api(`/recentStars?limit=${limit}`);
+      if (res.ok) {
+        const body = await res.json();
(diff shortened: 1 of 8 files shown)


test/profile/test-storage.js
```diff
@@ -16,7 +16,7 @@ import { ProfileStorage, VisitType } from '../../app/services/storage';
 
 describe('Utility tests', () => {
   it('Compares sets correctly.', (done) => {
-    expect(Immutable.Set(['a', 'b']).equals(Immutable.Set(['b', 'a'])));
+    expect(Immutable.Set(['a', 'b']).equals(Immutable.Set(['b', 'a']))).toBe(true);
```
ncalexan: Is there a different between `toBe` and `toEqual`?  Is one `==` and the other `===`?

app/main/user-agent-service.js
```diff
@@ -25,13 +25,23 @@ import * as profileDiffs from '../shared/profile-diffs';
 
 const PORT = 9090;
 
+const allowCrossDomain = function(req, res, next) {
```
ncalexan: Heh, I wondered about this.  Interesting.

app/ui/about/stars.html
```diff
@@ -29,5 +29,22 @@
 
 <body>
 Hello, 🌟.
+
+<ul id="bookmarks"></ul>
```
ncalexan: I feel the front-end folks will want to set some patterns for this, but I'm cool with this approach.

test/profile/test-storage.js
rnewman: Correct. `be` is `===`. Note that we use `.equals(…).toBe(true)` here because I haven't yet sent in a patch for `expect.js` to do the right thing with sets. I could use `.toBeTruthy` instead of `toBe(true)`, but this is more precise.

app/ui/about/stars.html
rnewman: Yeah, this is super 1999-era JS and HTML.

rnewman:
This now does history, too.

@bgrins, take a look?

Note that Electron seems to remove `_browser` from `window` when navigating to a different page, so nothing leaks here.

bgrins:
Redirecting to @jsantell

app/ui/about/stars.html
jsantell: This is totally fine for now.

app/main/user-agent-service.js
```diff
@@ -134,6 +144,21 @@ function configure(app: any, storage: ProfileStorage) {
...
+    }
+
+    const limit = parseInt(req.query.limit, 10) || Number.MAX_SAFE_INTEGER;
+
+    // TODO: this should return full-fledged page objects.
+    const pages = await storage.visited(0, limit);
+    res.json({ pages });
+  }));
+
+  router.get('/awesome', wrap(async function(req, res) {
```
jsantell: I'm guessing this is for the awesomebar completion, but we may want completions elsewhere, maybe a better name?

app/content/preload/content.js
```diff
@@ -15,5 +15,31 @@ import './context-menu';
 import { readerify } from './reader';
 import Readability from './readability';
 import './scroll';
+import { api } from '../../ui/browser/user-agent';
 
 window._readerify = readerify.bind(null, Readability);
+
+if (document.location.protocol === 'tofino:') {
+  console.log('Adding _browser.');
```
jsantell: remove console logging

jsantell:
r+ with comments addressed; does target work if you did `anchorEl.setAttribute('target', '_new')`?

app/main/user-agent-service.js
rnewman: Yeah, I wanted to rethink this. Changed to `query` for now.

rnewman:
Landed.

PR #419, opened 2016-05-12
Improper use of `expect` in tests

rnewman:
Some tests seem to be using `expect` without assertions, e.g. `expect(db instanceof DB)`. This does nothing. If `db` is not an instanceof DB, nothing will fail. Using `expect` only by itself is useless.

I saw this in `test-storage.js`. There might be other tests that do this, but I haven't looked too closely. @rnewman ping.


rnewman:
Good catch!

r? @victorporof

victorporof:
r+

Issue #342, opened 2016-05-04
+l Location Focus


jsantell:
Tracking for MVP milestone.

Should not be considered complete without tests.

bgrins:
This is now working, but needs tests still (and possible cleanup so the location component doesn't hook directly into IPC) before closing it out

bgrins:
This has been fixed, and hotkey tests aren't working right now.  Closing out

PR #422, opened 2016-05-12
Focus url bar

ncalexan:
Here's some WIP on #342.  @bgrins, over to you.


app/ui/browser/views/navbar/location.jsx
```diff
@@ -111,7 +110,10 @@ export class Location extends Component {
   }
 
   componentDidMount() {
-    this.props.ipcRenderer.on('focus-urlbar', () => this.refs.input.select());
+    this.props.ipcRenderer.on('focus-url-bar', () => {
```
jsantell: I feel this should go through the normal channels as the other ipc messages and actions and showURLBar should be on the redux state

bgrins: Ah, already pushed this in https://github.com/mozilla/tofino/commit/1c24497ea6a1fbf08d1cf710635ab54a74230f63 and https://github.com/mozilla/tofino/commit/e0a064eaebeb2ce252e756e37997408b3037a433.  Let's move that into a follow up

bgrins:
This landed

PR #423, opened 2016-05-12
Split out the API request module into its own and expose functions fo…

jsantell:
…r making requests to the main process service. r=nalexander

### DIFF
--- a/app/ui/browser/actions/main-actions.js
+++ b/app/ui/browser/actions/main-actions.js
@@ -36,12 +36,8 @@ export function createTab(location: ?string = undefined,
     dispatch({ type: types.CREATE_TAB, id, ancestorId, location, instrument: true });
 
     // TODO: properly track window scope.
-    userAgent.createSession({ ancestor: ancestorId, reason }).then(res => {
-      if (!res.ok) {
-        return {};
-      }
-      return res.json();
-    }).then(({ session }) => dispatch(didStartSession(id, session, ancestorId)));
+    userAgent.createSession({ ancestor: ancestorId, reason }).then(({ session }) =>
+      dispatch(didStartSession(id, session, ancestorId)));
   };
 }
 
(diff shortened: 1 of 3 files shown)


app/ui/browser/actions/main-actions.js
```diff
@@ -137,17 +104,12 @@ export function setUserTypedLocation(pageId: string, payload: Object): Action {
...
-      })
-        .then(async function (response) {
-          if (!response.ok) {
-            return;
-          }
-          const body = await response.json();
-          dispatch(profileDiffs.completions(payload.text, body.results));
-        })
-        .catch(); // Right now, nothing to be done.  In the future, we could retry.
+      userAgent.query({ text: payload.text }).then(res => {
```
ncalexan: Why not bake this JSON parsing into the UA library?  We should always get some JSON back, even if it's just `{}`.  (We might not, yet.)

app/ui/browser/lib/user-agent.js
```diff
@@ -0,0 +1,80 @@
...
+      page,
+    }
+  });
+}
+
+export function createSession({ ancestor, reason }) {
+  return request('/session/start', {
+    method: 'POST',
+    json: {
+      scope: 0,
```
ncalexan: Think I'd rather have this from the caller (and defaulted).

app/ui/browser/lib/user-agent.js
```diff
@@ -0,0 +1,80 @@
...
+    json: {
+      session,
+      reason,
+    },
+  });
+}
+
+export function createStar({ url, session, title }) {
+  return request(`/stars/${encodeURIComponent(url)}`, {
+    // @TODO Shouldn't this be a POST, in terms of creating
```
ncalexan: I thought `PUT` / `DELETE` made more sense than `POST` / `DELETE`.  Not married to it.

ncalexan:
By dropping the `.catch()`, you will see "errors" in the console that you can't really do anything about.  If you're happy with that, roll on.

app/ui/browser/actions/main-actions.js
jsantell: Initial thought was that's extra processing if we don't want it, but that is trivial -- will add that

app/ui/browser/lib/user-agent.js
jsantell: `PUT` should be used for updating and `POST` creating, but not familiar with the backend how this is actually represented

jsantell:
w/r/t `catch()`, I'd rather them be visible so we're atleast aware of them -- we shouldn't really be having dropped requests here for the most part

PR #425, opened 2016-05-13
Show each BrowserWindow, even in the face of early UI errors.

ncalexan:
This allows to debug errors that happen after the UI content starts to
load but before the UI connects to the UA service and messages the main
process that the window is ready.  This could happen for any number of
reasons: compile errors, runtime issues, failed WS connections, etc.


jsantell:
Just tried it out because was running into an error again -- worked great! r+

jsantell:
This does something weird to the hotkey tests though

ncalexan:
I also have witnessed hotkey errors.  I can't explain them, and now I can't reproduce locally :(

PR #426, opened 2016-05-13
Wait for a page's session id on all user agent requests. r=nalexander

jsantell:
This does what we talked about. The inclusion of the deferred library is to prevent nested Promise constructors, and it still uses native promises and is rather short.

### DIFF
--- a/app/ui/browser/lib/user-agent.js
+++ b/app/ui/browser/lib/user-agent.js
@@ -14,28 +14,53 @@ import request from './request';
 import * as endpoints from '../../../shared/constants/endpoints';
 import { getPageById } from '../selectors/index';
 
-export function createPage({ url, session, page }) {
-  return request(`/pages/${encodeURIComponent(url)}`, {
-    method: 'POST',
-    json: {
-      session,
-      page,
-    }
-  });
+/**
+ * A mapping from a page id to a promise that
+ * will resolve to a session id. When we initiate the session,
+ * we immediately store a promise that will resolve to the session id
+ * so that future requests using the page id can hook into the initial promise
+ * even before we have the session.
+ */
+const PAGE_TO_SESSION = new Map();
+
+async function waitForSession(page) {
+  // If we have a session id, use it; this could have been generated
+  // this browsing session, or from a previous session, but either way
+  // this was an explicit choice.
+  if (page.sessionId) {
+    return page.sessionId;
+  }
+  // If we have a promise to the session for this page, and the page
+  // does not have its own sessionId, we probably hit a race condition,
+  // so wait until the page has its sessionId by listening for this promise.
+  else if (PAGE_TO_SESSION.has(page.id)) {
+    return PAGE_TO_SESSION.get(page.id);
+  }
+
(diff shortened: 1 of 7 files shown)


victorporof:
@ncalexan @jsantell should this be landed?

app/ui/browser/lib/user-agent.js
```diff
@@ -10,32 +10,59 @@
...
     json: {
       scope,
       ancestor,
       reason,
     },
-  })
+  }).then(res => {
+    sessionRequestPromise.resolve(res.session);
+    return res;
+  });
```
victorporof: Maybe something like
```
const promise = request();
map.set(id, promise.then(res => res.session))
return promise;
```
would be better.

app/ui/browser/lib/user-agent.js
```diff
@@ -14,28 +14,51 @@ import request from './request';
...
+ */
+const PAGE_TO_SESSION = new Map();
+
+async function waitForSession(page) {
+  // If we have a promise to the session for this page,
+  // that means we requested a new session, so use it.
+  if (PAGE_TO_SESSION.has(page.id)) {
+    return PAGE_TO_SESSION.get(page.id);
+  }
+  // If we don't have a session promise saved, we may have used a
```
ncalexan: I feel like this should be first, but they are exclusive, so okay.

app/ui/browser/lib/user-agent.js
```diff
@@ -78,3 +109,13 @@ export function createHistory({ url, title, session }) {
   });
 }
 
+/**
+ * Unauthenticated Requests
```
ncalexan: _Authenticated_ is the wrong term here.  Do you mean _ambient_ or something similar?

ncalexan:
So I'm fine with this approach, but I would like you to consider binding the `userAgent` client to the `page` directly.  That is, rather than `userAgent(page, { parameters })` consider `page.userAgent({ parameters })`.

However, if you prefer this expression, roll on.

app/ui/browser/lib/user-agent.js
jsantell: Moving this to first -- fixing up tests, and since we're not running a UA service in unit tests, this will actually work once flipping this order.

app/ui/browser/lib/user-agent.js
jsantell: *Unauthorized*? Pretty much requests that do not require a session id.

jsantell:
> 
> 
> So I'm fine with this approach, but I would like you to consider binding the userAgent client to the page directly. That is, rather than userAgent(page, { parameters }) consider page.userAgent({ parameters }).
> 
> However, if you prefer this expression, roll on.

We don't have methods on any other state objects, so I think this would be a bit strange (and not sure how functions are handled in a redux state)

app/ui/browser/lib/user-agent.js
ncalexan: I think I made an error returning 401.  That should have just been 400.  There's nothing about authorization or authentication here; it's just that we need a session for some things.

Issue #420, opened 2016-05-12
Add test that makes sure app/ doesn't use modules from devDependencies


victorporof:

victorporof:
@Mossop making sure you saw this.

PR #427, opened 2016-05-13
Don't import modules from devDependencies in app/ code

victorporof:
Fixes https://github.com/mozilla/tofino/issues/420


test/lint/test-devdeps.js
```diff
@@ -0,0 +1,46 @@
...
+
+// XXX: Which of these do we actually need to ignore?
+const ignored = [
+  'electron-ipc-mock',
+
+  // XXX: I'm not sure about these ones...
+  'express',
+  'react-addons-perf',
+  'source-map-support',
+  'uuid',
```
Mossop: react-addons-perf can be ignored. The others should be put into dependencies, no devDependencies (looks like express is in both)

Mossop:
LGTM

PR #430, opened 2016-05-13
All page navigation appears to be user-entered text

rnewman:
I noticed that we're hitting the query endpoint on every page load. That's because every page load sends `SET_USER_TYPED_LOCATION`, just as if the user typed the URL… and that triggers a search.


rnewman:
@ncalexan, @jsantell: take a look?

ncalexan:
lgtm.  I think @bgrins is the one to flag here.

rnewman:
Landed.

PR #433, opened 2016-05-13
Make the main menu open from the hamburger button too.

### LINKED ISSUE #305: The main menu isn't available on Windows
Either because it is OSX only or because we've hidden the titlebar or something but the menu that we're making changes to isn't visible to windows users. We must fix that or remove it and use something else that Windows users can see.

Mossop:
We cache the bookmarks data so we can rebuild the menu at any time and some
bits are moved around to allow for building menus for minimal OSX, regular
application and the hamburger menu. The hamburger menu has a slightly different
template allowing important options to be at the top level.

Fixes #305.

### DIFF
--- a/app/main/menu/index.js
+++ b/app/main/menu/index.js
@@ -11,22 +11,35 @@ specific language governing permissions and limitations under the License.
 */
 
 import electron from 'electron';
-import assert from 'assert';
-import BrowserMenuTemplate from './template';
-
-const DEFAULT_MENU_ITEMS = [
-  'file', 'edit', 'view', 'history', 'bookmarks', 'tools', 'window', 'help',
-];
-
-const OSX_MINIMAL_MENU_ITEMS = [
-  'file', 'edit', 'window', 'help',
-];
+import { MinimalAppMenuTemplate, AppMenuTemplate, WindowMenuTemplate } from './template';
+import menus from './menus';
 
 /**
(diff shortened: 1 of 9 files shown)


app/main/menu/index.js
```diff
@@ -11,58 +11,60 @@ specific language governing permissions and limitations under the License.
...
+    }
+
+    return itemTemplate;
+  };
+
+  return template.map(buildItem);
+}
+
+function buildMenu(template, data) {
+  return currentMenu = electron.Menu.buildFromTemplate(buildTemplate(template, data));
```
bgrins: Nit: please split this up into two lines

bgrins:
With this applied, the "File" menu is missing on the OSX app menu bar

bgrins:
screenshot 2016-05-19 09 30 45

Mossop:
@bgrins That should fix it but you should probably check

app/ui/browser/lib/webview-controller.js
```diff
@@ -71,6 +71,16 @@ class WebViewController extends EventEmitter {
     this.emit('navigate-refresh', pageId);
     this.emit(`navigate-refresh:${pageId}`, pageId);
   }
+
+  capturePage(pageId) {
+    console.log(3);
```
bgrins: Leftover console.log()

bgrins:
Nice

PR #443, opened 2016-05-17
Fix #376 - Immediately go to URL and close autocomplete widget

### LINKED ISSUE #376: hitting enter to select a url from history does not load the url
STR:

1. click on the location bar 
2. start typing a url form the borwser history, eg google news
3. arrow down to select a url from history then hit enter

ER: the browser should complete the url and load the page
AR: the browser only completes the url, you need to hit enter a second time to load the page

darkwing:
Needs a test or two.

### DIFF
--- a/app/ui/browser/views/navbar/location.jsx
+++ b/app/ui/browser/views/navbar/location.jsx
@@ -180,6 +180,12 @@ export class Location extends Component {
     this.refs.input.ownerDocument.execCommand('insertText', false, value);
   }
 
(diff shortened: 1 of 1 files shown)


bgrins:
Looks fine to me - @jsantell what do you think?  The only way to test this right now is via spectron, which is basically permafailing, so I'm OK with putting that off until we have a better solution for testing the location component.

darkwing:
Rebased!

PR #453, opened 2016-05-18
Make it possible to skip building in appveyor like travis

Mossop:
Travis already skips builds with "[ci skip]" in a commit message by default, this makes appveyor do the same.


bgrins:
Nice!

Issue #456, opened 2016-05-18
Eslint tests currently fail


victorporof:
app/content/preload/content.js
  32:47  error  Unexpected string concatenation  prefer-template
  35:47  error  Unexpected string concatenation  prefer-template
  38:53  error  Unexpected string concatenation  prefer-template

victorporof:
Blame https://github.com/victorporof/tofino/commit/205a4a703884ccda512078d21481930ce1480e2b

PR #457, opened 2016-05-18
Avoid string concatenation in preload/content.js

victorporof:

Fixes https://github.com/mozilla/tofino/issues/456


victorporof:
Fixes https://github.com/mozilla/tofino/issues/456

jsantell:
It looks like this was handling `undefined` values since `encodeURIComponent(undefined) === "undefined"` -- let's handle this via default args or something, because this looks like it was doing something useful

r+ with that in mind

victorporof:
The changes are exactly equivalent with what used to be there before. These changes just make the build pass again since eslint doesn't like string concatenation.

PR #459, opened 2016-05-18
Readd react-addons-test-utils that was removed in 82bc1bd2, since it'…

jsantell:
…s necessary for other React modules, and ignore it in the package linting. r=victorporof


victorporof:
nice catch r+

PR #461, opened 2016-05-18
Handle escape in the URL bar

bgrins:
@linclark can you have a look?


app/ui/browser/model/index.js
```diff
@@ -66,6 +66,9 @@ export const UIState = Immutable.Record({
 
   // What the user has typed into the location bar, stored by page id
   userTypedLocation: Immutable.Map(),
+
+  // Should the autocomplete popup be visible
+  showCompletions: true,
```
linclark: would we want this to be indexed by page? I'm not sure if we keep the completions box open when you switch tabs.

app/ui/browser/reducers/ui-state.js
```diff
@@ -23,10 +23,14 @@ export default function uiState(state = initialState, action) {
       return state.set('statusText', action.text);
 
     case types.LOCATION_CHANGED:
+      return state.setIn(['userTypedLocation', action.pageId], action.payload.text);
 
-      // Fallthrough.
     case types.SET_USER_TYPED_LOCATION:
-      return state.setIn(['userTypedLocation', action.pageId], action.payload.text);
+      return state.setIn(['userTypedLocation', action.pageId], action.payload.text)
```
linclark: not a huge deal, but might want to use withMutations to avoid creating the intermediate object that is returned from setIn. See an [example here](https://github.com/mozilla/tofino/blob/c959b42c374f97356e06b28dcd65fe492d9a5efa/app/ui/browser/reducers/pages.js#L82)

app/ui/browser/model/index.js
bgrins: We have some logic in the local state for the locations component right now that hides the popup if the url bar isn't visible / focused, which would be the case when switching tabs.  I think we should move these bits of state up to the UI state using the same pattern as here.  Not sure that indexing completion by page id is necessary right now although it might become eventually.

app/ui/browser/reducers/ui-state.js
bgrins: Thanks, I'll update it

app/ui/browser/views/browser.jsx
```diff
@@ -82,8 +82,9 @@ class BrowserWindow extends Component {
       }));
     };
     const onLocationContextMenu = e => menuLocationContext(e.target, currentPageId, dispatch);
+    const onClearCompletions = () => dispatch(actions.clearCompletions());
     const onLocationReset = () => {
-      dispatch(actions.setUserTypedLocation(currentPageId, { text: void 0 }));
+      dispatch(actions.locationChanged(currentPageId, { text: null }));
```
linclark: It looks like this is an existing action. Do we know if there was a reason this wasn't already being used here?

bgrins: I don't think there was a reason - I believe locationChanged was added after the fact and this never got changed b/c there was no difference between the two actions at the time

app/ui/browser/views/navbar/location.jsx
```diff
@@ -163,6 +163,17 @@ export class Location extends Component {
     return 'glyph-bookmark-hollow-16.svg';
   }
 
+  getVisibleCompletionsForURL() {
```
linclark: We use the word location in other places, like userTypedLocation. Should we be consistent?

linclark: Also, it feels like this might make sense as a selector rather than a function on the component.

app/ui/browser/views/navbar/location.jsx
```diff
@@ -367,11 +391,13 @@ Location.propTypes = {
   ipcRenderer: PropTypes.object.isRequired,
   navigateTo: PropTypes.func.isRequired,
   userTypedLocation: PropTypes.string.isRequired,
+  showCompletions: PropTypes.bool.isRequired,
 };
 
 function mapStateToProps(state, ownProps) {
   return {
     userTypedLocation: getUserTypedLocation(state, ownProps.page.id),
+    showCompletions: state.uiState.showCompletions,
```
linclark: We've been using selectors (like getUserTypedLocation above). This makes it easier to restructure the state shape if we need to.

app/ui/browser/views/navbar/location.jsx
bgrins: I'll rename to getVisibleCompletionsForLocation

app/ui/browser/views/navbar/location.jsx
```diff
@@ -306,7 +322,14 @@ export class Location extends Component {
           ref="input"
           onFocus={this.handleURLBarFocus}
           onBlur={this.handleURLBarBlur}
-          onChange={this.props.onLocationChange}
+          onChange={(e) => {
+            // There's a weird case here where the location can be reset from an action
```
linclark: we discussed this in person, the comment should be expanded to explain why there is a weird case

PR #462, opened 2016-05-19
Use default cursor for navbar and location bar

darkwing:
The navbar, tab content, and autocomplete items all show the text cursor upon hover which seems weird -- no one is looking to copy that text.  Default cursor seems much more logical.


victorporof:
r+

PR #468, opened 2016-05-19
Upload packages to a private s3 location.

Mossop:
Uploads our packages to a private location ready for distribution when we decided to do so.


victorporof:
(lgtm)

PR #470, opened 2016-05-19
Status component cleanups and tests

victorporof:
@bgrins 

Relevant changes start with commit ed327c3.

### DIFF
--- a/app/ui/browser/views/page/status.jsx
+++ b/app/ui/browser/views/page/status.jsx
@@ -34,27 +34,18 @@ const STATUS_STYLE = Style.registerStyle({
  * A status bar at the bottom of the browser UI.
  */
 
-const Status = ({ page, statusText }) => {
-  let contents;
-
-  if (statusText) {
-    contents = statusText;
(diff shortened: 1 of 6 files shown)


app/ui/browser/views/page/status.jsx
```diff
@@ -24,36 +23,29 @@ const STATUS_STYLE = Style.registerStyle({
...
+      {statusText}
     </div>
   );
-};
+}
 
 Status.displayName = 'Status';
 
 Status.propTypes = {
-  page: PropTypes.object.isRequired,
```
bgrins: Are we now actually not passing the 'page' in as a prop anymore?

bgrins:
r+, see line note

app/ui/browser/views/page/status.jsx
victorporof: We don't use it anymore. We still pass it in somewhere, but the only reason it was used was because we cared about the `isLoading` property on a page, which doesn't exist anymore. We use the `getStatusText` selector now for status texts.

PR #475, opened 2016-05-19
Use path module so windows doesn't attempt to run renderer tests with…

jsantell:
… plain mocha. r=vporof


jsantell:
Handled in #476

PR #478, opened 2016-05-19
Switch to electron 1.1 and require Node 6 by default.

Mossop:
Not ready for landing yet

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,10 @@
   ],
 
   "settings": {
+    "import/extensions": [
+      ".js",
+      ".jsx"
+    ],
     "import/resolver": {
       "node": {
         "extensions": [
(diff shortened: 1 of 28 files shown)


.eslintrc
```diff
@@ -63,6 +64,7 @@
     "react/jsx-indent-props": ["error", 2],
     "react/jsx-max-props-per-line": ["error", { "maximum": 1 }],
     "react/jsx-equals-spacing": ["error", "never"],
+    "react/jsx-first-prop-new-line": ["off"],
```
victorporof: Interesting that we need this

Mossop: upgrading babel and eslint which seems to be required to work with node 6
also meant upgrading the airbnb preset which seems to have changed a bunch
of stuff. I'm trying to figure out what is needed and what isn't right now.

victorporof: 👍

Mossop:
Looks like this is ready for review now @victorporof

bgrins:
r+

PR #494, opened 2016-05-22
Use Control+Tab and Control+Shift+Tab to move between tabs

darkwing:
I know this is an experimental browser but this shortcut is has been around forever sooooo....


victorporof:
neat!

PR #501, opened 2016-05-23
[POC] Using Relay

linclark:
bookmarks

## Proof-of-concept caveats
- I took the suggestion to mock out bookmarks (instead of the whole browser). This made it easier to build the proof of concept, but also means we don't see the full effect of using Relay. For example, I expect we'd have less full-app rerendering if we were using Relay to manage data higher in the tree.
- I didn't implement toggling on the bookmark. If you want to delete it, use the delete button in the bookmark pane
- The current implementation uses a service provider (Reindex) to persist the data. If we decide to go with Relay, we'd probably have the data persisted locally, or have some that is local and some in the cloud. The current schema is more complicated than it would actually be because of Reindex defaults.

## Should we explore further?

**Pros**
- I think this would make it easier to experiment with persisted data in the front-end.
- It's easy to de-optimize any performance improvements in the Redux architecture. I believe it would be easier to maintain Relay's performance.
- This architecture provides interesting options for persisting data, like having some in the cloud.

**Cons**
- There are a number of new concepts that component developers would need to learn.
- It can be frustrating to debug data issues since Relay's a bit of a black box.
- I don't know if Relay makes it harder to test components.

*Please comment with your own pros/cons*


victorporof:
Even though the benefits are very appealing, using relay looks like it might interfere with rapid prototyping, which we're currently focused on afaik. Is that true?

Would switching to relay at a later stage of the project development be more difficult? It's not clear to me what the migration cost is at different stages.

linclark:
> Even though the benefits are very appealing, using relay looks like it might interfere with rapid prototyping, which we're currently focused on afaik. Is that true?

I think in the long run, there's a good chance it would make prototyping faster. Front-end devs would be able to create their own weird connections between different parts of the data, rather than waiting for a backend service to be developed. Plus, it makes components more composable so they can be moved around more easily.

However, I'd say it would probably slow us down in the short run. We'd have to 1) train all the component developers on the conventions of Relay, which are kind of complex, 2) build out the backend, and 3) do the conversion.

> Would switching to relay at a later stage of the project development be more difficult? It's not clear to me what the migration cost is at different stages.

I think it would be more difficult at a later stage, but not impossible. The basic component structure will probably be the same when we do the conversion, so that wouldn't require too much work. What we would need to do is 1) figure out the query fragment for each component and 2) convert all actions to mutations.

rnewman:
There's an additional element to the calculation: if we switch later, and
the project continues, then we will have maintained specific backend APIs
(and their front end callers) for a longer duration. They're not super
expensive, but it's cheaper to front-load changes.

All of this is moot if we don't continue to drive this project forward for
a few more months, of course.

linclark:
As discussed in the meeting today, we do not plan to pursue this at this time. Introducing Relay would slow us down in the short term, and Relay itself introduces a lot of functionality that is unnecessary to our use case. However, it introduces interesting concepts (for example, expressing data dependencies for a component explicitly and in a composable way) which might be useful. 

Closing it now, but discussion may still continue on this PR to clarify concepts.

PR #506, opened 2016-05-23
Add `npm run clobber` command to clear out built folders. r=Mossop

bgrins:
@Mossop what do you think?  The sad part here is you can't run clobber twice in a row because the second time the required modules to run the command are gone.  So not sure if we should go ahead and add this or just include instructions in the docs to manually remove the three folders for people having build problems.

### DIFF
--- a/build/const.js
+++ b/build/const.js
@@ -2,3 +2,6 @@ import path from 'path';
 import * as BuildUtils from './utils';
 
 export const DIST_DIR = path.join(BuildUtils.getRoot(), 'dist');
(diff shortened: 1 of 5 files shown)


Mossop:
Ugh yeah it probably isn't great to require a module in order to remove it. Another option would be to do this as a python script instead since we require python to build anyway.

victorporof:
We already have the clean task, which does something similar. Should the clobber and clean tasks be merged? Is it useful to have them both?

bgrins:
OK, this probably makes more sense as a bit of instruction in the readme for now (I'm not sure if anyone else has been bit by the thing that made me want to add this so it's not worth too much time)

PR #516, opened 2016-05-24
Always launch a new process for the UA service.

Mossop:
This doesn't work yet (the UA service doesn't start up) but it should be pretty close. Basically we try to connect the websocket, if it fails we start a new service and connect again. The service closes itself once all clients have gone away.

Does this seem like it is along the right lines @ncalexan?

One issue is that there is a delay between starting the service and being able to connect. I almost want to make the main process listen for a connection from the service in that case.

### DIFF
--- a/app/main/browser.js
+++ b/app/main/browser.js
@@ -38,16 +38,14 @@ import * as menu from './menu/index';
 import * as instrument from '../services/instrument';
 import registerAboutPages from './about-pages';
 import * as BW from './browser-window';
-import { ProfileStorage } from '../services/storage';
-import * as userAgentService from './user-agent-service';
-const profileStoragePromise = ProfileStorage.open(path.join(__dirname, '..', '..'));
+import userAgentClient from '../shared/user-agent-client';
 
-import WebSocket from 'ws';
 import * as endpoints from '../shared/constants/endpoints';
 
 const app = electron.app; // control application life.
 const ipc = electron.ipcMain;
 const globalShortcut = electron.globalShortcut;
+const userAgentPromise = userAgentClient.connect();
 
 const appStartupTime = Date.now();
 instrument.event('app', 'STARTUP');
@@ -64,7 +62,7 @@ app.on('ready', async function() {
   // Register `about:*` protocols after app's 'ready' event
   registerAboutPages();
 
-  await BW.createBrowserWindow(profileStoragePromise, () => {
+  await BW.createBrowserWindow(userAgentPromise, () => {
     instrument.event('browser', 'READY', 'ms', Date.now() - browserStartTime);
   });
 });
@@ -89,12 +87,12 @@ app.on('activate', async function() {
   // On OS X it's common to re-create a window in the app when the
   // dock icon is clicked and there are no other windows open.
   if (electron.BrowserWindow.getAllWindows().length === 0) {
-    await BW.createBrowserWindow(profileStoragePromise, () => menu.buildAppMenu(menuData));
+    await BW.createBrowserWindow(userAgentPromise, () => menu.buildAppMenu(menuData));
   }
 });
 
 ipc.on('new-browser-window', async function() {
-  await BW.createBrowserWindow(profileStoragePromise, () => menu.buildAppMenu(menuData));
+  await BW.createBrowserWindow(userAgentPromise, () => menu.buildAppMenu(menuData));
 });
 
 ipc.on('close-browser-window', BW.onlyWhenFromBrowserWindow(async function(bw) {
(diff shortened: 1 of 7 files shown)


app/main/user-agent-client.js
```diff
@@ -0,0 +1,87 @@
...
+    this.on('message', (type, args) => {
+      this.emit(type, ...args);
+    });
+  }
+}
+
+const timeout = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
+function startService() {
+  // Runs Electron as a node process
```
ncalexan: Is there an advantage to launching Electron in this way?  Should we perhaps launch `argv[0]` or something to respawn?

app/main/user-agent-service.js
```diff
@@ -66,23 +66,30 @@ function configure(app: any, storage: ProfileStorage) {
...
 
     ws.on('close', () => {
-      const index = diffsClients.indexOf(ws);
+      const index = wsClients.indexOf(ws);
       if (index > -1) {
-        diffsClients.splice(index, 1);
+        wsClients.splice(index, 1);
+      }
+
+      if (wsClients.length === 0) {
```
ncalexan: I'm pretty sure we want the service to outlive clients and be tied to the both client counts and the launching process, but I'm fine with this for now.

app/main/user-agent-service.js
```diff
@@ -66,23 +66,30 @@ function configure(app: any, storage: ProfileStorage) {
...
+        // No more clients, close the service
```
ncalexan: Nit: full sentence comments, and logging.

app/main/user-agent-client.js
```diff
@@ -0,0 +1,87 @@
...
+
+export default {
+  async connect() {
+    console.log('Attempting to connect');
+    try {
+      return await attemptConnect();
+    } catch (e) {
+      if (e.code === 'ECONNREFUSED') {
+        startService();
+        await timeout(5000);
```
ncalexan: Exponential backoff?  This is pretty blunt.

app/shared/constants/endpoints.js
```diff
@@ -12,4 +12,4 @@
 
 export const UA_SERVICE_PORT = 9090;
 export const UA_SERVICE_HTTP = `http://localhost:${UA_SERVICE_PORT}/v1`;
-export const UA_SERVICE_WS = `ws://localhost:${UA_SERVICE_PORT}/v1`;
+export const UA_SERVICE_WS = `ws://localhost:${UA_SERVICE_PORT}/v1/ws`;
```
ncalexan: I would guess that this isn't doing the right thing in the UI client module, since `/diffs` would have had to get pushed in there somewhere.

app/main/browser.js
```diff
@@ -260,4 +245,6 @@ profileStoragePromise.then(async function(profileStorage) {
       menu.buildAppMenu(menuData);
     }
   });
+}, (err) => {
```
ncalexan: In Node.js, console methods are bound, so just `, console.error)` is good.

ncalexan:
This looks mostly okay to me.  I've left a few nits and comments throughout; interested to see the final form.

app/main/user-agent-client.js
Mossop: Right, see my note in the conversation.

app/main/user-agent-client.js
Mossop: Well, the one thing we know a tofino user has installed is the current electron build. They may not have node or anything else unless we add that as a dependency and ship it.

app/main/user-agent-service.js
Mossop: For debugging it is very helpful to have the service shut itself down, but yes we want something longer in the future.

Mossop:
I've updated this and it is ready for review though there are no tests right now. If you have suggestions beyond the webdriver tests verifying this as a side effect then let me know.

At some point it might make sense to merge shared/user-agent-client.js and ui/browser/lib/user-agent.js, but by the time I discovered that I was almost done so I'd rather get this landed first and see if any issues come up.

app/shared/user-agent-client.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    ws.on('error', (e) => {
+      reject(e);
+    });
+  });
+}
+
+async function backoffConnect() {
+  let timedout = false;
```
ncalexan: nit: `timedOut`.

ncalexan:
Nifty!  This works well locally, as long as I don't kill the initiating process.  (That's an intentional limitation.)

Throughout, I prefer full sentence comments, but whatever.

Mossop:
Really? I can kill the initiating process on windows and it seemed to work last I checked.

app/shared/user-agent-client.js
```diff
@@ -0,0 +1,134 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed
+ under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+
+import path from 'path';
+import EventEmitter from 'events';
+import WebSocket from 'ws';
+import request from 'request';
```
jsantell: We have isomorphic-fetch, do we also need request? We should have one or the other

app/shared/user-agent-client.js
```diff
@@ -0,0 +1,134 @@
...
+  startSession() {
+    return uaRequest('/session/start', {
+      method: 'POST',
+    });
+  }
+}
+
+const timeout = (ms) => new Promise(resolve => setTimeout(resolve, ms));
+
+function startService() {
```
jsantell: What/where is `server.js`? I don't think this client file (which is shared in the UI code) should do this, especially as it's using child process, and are trying to make the UI as node-less as possible (exceptions to unprivileged APIs that are browserifiable)

app/shared/user-agent-client.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+    });
+
+    ws.on('error', (e) => {
+      reject(e);
+    });
+  });
+}
+
+async function backoffConnect() {
```
jsantell: I'd just use a backoff package from npm (but just IMO here)

app/shared/user-agent-client.js
Mossop: Maybe, it would mean splitting this into two files, one that attempts to connect and starts the UA service if it fails and a second that just assumes it can always connect.

Issue #409, opened 2016-05-11
Allow console logging in mocha tests


victorporof:
I'm seeing a lot of /* eslint no-console: 0 */ and /* eslint-disable console */ in tests. We should probably just put those in the dedicated .eslintrc file.

PR #519, opened 2016-05-24
Make eslint more bearable

victorporof:
Fixes https://github.com/mozilla/tofino/issues/409 and a bunch of other things

@bgrins


victorporof:
See individual commits for what I propose changing/adding.

bgrins:
Nice, looks good to me - r+

Issue #341, opened 2016-05-04
Ctrl/Cmd+N: New Window, focus location bar


jsantell:
Tracking for MVP milestone.

I believe these are mostly working, but should not be considered complete without tests. I do not think the location bar is focusing at the moment.

jsantell:
Tests added in #403 -- still need focus location portion

victorporof:
Location is now focusing in the new frontend.

Issue #340, opened 2016-05-04
+t New Tab, focus location bar


jsantell:
Tracking for MVP milestone.

I believe these are mostly working, but should not be considered complete without tests. I do not think the location bar is focusing at the moment.

jsantell:
Tests added for new tab in https://github.com/mozilla/tofino/pull/370; still need to add focus location bar functionality

PR #521, opened 2016-05-24
Show and focus URL bar when creating a new tab.

ncalexan:
This also shows and focuses the URL bar when creating a new window,
since a new window immediately creates a new tab.

I elected to allow dispatching lists (Arrays) of actions to save renders
and set a pattern for avoiding races in the future.  In spirit, this is
similar to https://twitter.com/dan_abramov/status/656074974533459968 and
https://github.com/tshelburne/redux-batched-actions.  However, this
approach works with our thunk middleware (and those don't).

### DIFF
--- a/app/ui/browser/actions/main-actions.js
+++ b/app/ui/browser/actions/main-actions.js
@@ -33,8 +33,18 @@ export function createTab(location: ?string = undefined,
     // We use the Page ID to associate the User Agent allocated session ID with the Page.
     const id = uuid.v4();
 
-    // Start loading a tab while asking the User Agent to start its browsing session.
+    // Start loading a tab while asking the User Agent to start its browsing session.  It's safe to
+    // dispatch sequentially, since these are "plain actions".  It's unfortunate to render multiple
+    // times, but c'est la vie.  If the actions change to also dispatch on future ticks, we might
+    // race to set focus across user interactions.
     dispatch({ type: types.CREATE_TAB, id, ancestorId, location, options, instrument: true });
+    // New tabs are opened with the URL bar focused, ready to enter a new location.
+    dispatch(setShowURLBar(id, true));
+    dispatch(setFocusedURLBar(id, true));
+
+    if (options.selected) {
+      dispatch(resetUIState()); // Reset global UI state when displaying the new tab immediately.
+    }
 
     // TODO: properly track window scope.
     userAgent.createSession(id, { ancestor: ancestorId, reason }).then(({ session }) =>
@@ -101,21 +111,27 @@ export function locationChanged(pageId: string, payload: Object): Action {
   };
 }
 
+export function resetUIState(): Action {
+  return (dispatch) => {
+    dispatch({ type: types.RESET_UI_STATE, instrument: false });
+  };
+}
+
 export function clearCompletions(): Action {
   return (dispatch) => {
     dispatch({ type: types.CLEAR_COMPLETIONS, instrument: false });
   };
 }
 
-export function setShowURLBar(visible) {
+export function setShowURLBar(pageId: string, visible: boolean) {
   return (dispatch) => {
-    dispatch({ type: types.SET_URL_INPUT_VISIBLE, payload: { visible } });
+    dispatch({ type: types.SET_URL_INPUT_VISIBLE, pageId, payload: { visible } });
   };
 }
 
-export function setFocusedURLBar(focused) {
+export function setFocusedURLBar(pageId: string, focused) {
   return (dispatch) => {
-    dispatch({ type: types.SET_URL_INPUT_FOCUSED, payload: { focused } });
+    dispatch({ type: types.SET_URL_INPUT_FOCUSED, pageId, payload: { focused } });
   };
 }
 
--- a/app/ui/browser/reducers/ui-state.js
+++ b/app/ui/browser/reducers/ui-state.js
@@ -10,35 +10,48 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
+import assert from 'assert';
 import 'babel-polyfill';
 
 import * as types from '../constants/action-types';
 import { UIState } from '../model';
+import { isUUID } from '../browser-util';
 
 const initialState = new UIState();
 
 export default function uiState(state = initialState, action) {
   switch (action.type) {
-    case types.SET_STATUS_TEXT:
-      return state.set('statusText', action.text);
-
+    // Per-Page state modifications first.  Eventually, we'll separate another reducer here.
     case types.LOCATION_CHANGED:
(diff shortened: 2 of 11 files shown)


ncalexan:
@bgrins you've been mostly owning this.  Take a look?  This should address #340 and #341.  However, I don't know how to test that the focus has been set.  Actually, I now see `ui-state.js` handling these.  I'll add tests for these properties.

bgrins:
I was actually able to produce a crash with this applied by doing: Open New Window, Close the window, Add tab on the original window, close tab on the original window. Looks like it's referencing this call, although only on the generated source: `bw.webContents.send('focus-url-bar')`

    Error: Object has been destroyed
        at Error (native)
        at Function.<anonymous> (/tofino/lib/main/hotkeys.js:73:551)
    127.0.0.1 - - [24/May/2016:21:51:04 +0000] "POST /v1/session/end HTTP/1.1" - - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Electron/1.1.0 Safari/537.36"
    Build failed.
    Error: Exited with exit code 1
        at ChildProcess.<anonymous> (/tofino/build/utils.js:95:18)
        at emitTwo (events.js:106:13)
        at ChildProcess.emit (events.js:191:7)
        at Process.ChildProcess._handle.onexit (internal/child_process.js:204:12)

bgrins:
Oh wait, that crash seems to happen even without this commit applied

bgrins:
If I open up the autocomplete box on the first tab by starting to edit it, then close it, then open a new tab, then the autocomplete box pops back up.  Off hand, I think if we did a clearCompletions() action when focusing and showing the URL bar that would fix this issue

ncalexan:
@bgrins -- my guess is you're seeing something like #527.  That commit may address your crash.

Is it reasonable to `clearCompletions` during `CREATE_TAB`?  That is, should creating a new (selected) tab also dispatch a "reset UI state" command?

bgrins:
> Is it reasonable to clearCompletions during CREATE_TAB? That is, should creating a new (selected) tab also dispatch a "reset UI state" command?

If the tab is going to become selected, I think so.  Same as how awesomebar becomes closed when opening a new tab in Firefox.

ncalexan:
@bgrins, @linkclark -- sorry for the force-push; hopefully earlier comments are still present.  I built on top to ease review.  How do you feel about this implementation?

app/ui/browser/actions/main-actions.js
```diff
@@ -34,7 +34,16 @@ export function createTab(location: ?string = undefined,
     const id = uuid.v4();
 
     // Start loading a tab while asking the User Agent to start its browsing session.
-    dispatch({ type: types.CREATE_TAB, id, ancestorId, location, options, instrument: true });
+    // Dispatching a list ensures only one render and prevents racing to set focus across user
```
linclark: We may want to rethink the underlying approach (existing before this PR) to clearing the UI state. My thinking is that if we need to ensure only one render happens, we've lost idempotence.

app/ui/browser/reducers/ui-state.js
```diff
@@ -23,22 +25,31 @@ export default function uiState(state = initialState, action) {
...
     case types.SET_USER_TYPED_LOCATION:
+      assert(isUUID(action.pageId), 'SET_USER_TYPED_LOCATION requires a page id.');
       return state.withMutations(mut => {
         mut.setIn(['userTypedLocation', action.pageId], action.payload.text);
         mut.set('showCompletions', true);
       });
 
+    case types.RESET_UI_STATE:
+      return state.withMutations((mut) => {
+        mut.set('showCompletions', false);
```
linclark: I don't think we need withMutations here since we're only setting one thing.

app/ui/browser/reducers/ui-state.js
```diff
@@ -23,22 +25,31 @@ export default function uiState(state = initialState, action) {
       return state.set('statusText', action.text);
 
     case types.LOCATION_CHANGED:
+      assert(isUUID(action.pageId), 'LOCATION_CHANGED requires a page id.');
```
linclark: I had been meaning to ask about our use of assert, so created an issue for discussion of it, #550

app/ui/browser/actions/main-actions.js
ncalexan: @linclark so the re-render is purely an implementation detail; it's not *required*.  In general, I don't understand perfectly what our race story is across multiple action dispatches (assuming the actions are not thunks).  If we dispatched two actions, is there a guarantee that another action (switching tabs!) don't interleave?  I know things are synchronous mostly but that type of thing scares me.

app/ui/browser/reducers/ui-state.js
ncalexan: I'll drop this.

app/ui/browser/actions/main-actions.js
linclark: Here's my understanding:

Given our current configuration, using the thunk middleware we have... If two plain object actions are dispatched in a row, then no other action should be able to get in between them. A plain object action gets routed through our thunk middleware without dispatching anything else. So it get passed into Redux's dispatch, which [passes the state and action](https://github.com/reactjs/redux/blob/master/src/createStore.js#L170) to the root reducer synchronously and doesn't allow any actions to be dispatched in the reducers.

The one caveat is that if we introduce new custom middleware, it could theoretically do wild and unforseen things.

app/ui/browser/views/navbar/location.jsx
```diff
@@ -406,7 +406,10 @@ Location.propTypes = {
   bookmark: PropTypes.func.isRequired,
   unbookmark: PropTypes.func.isRequired,
   navigateTo: PropTypes.func.isRequired,
-  userTypedLocation: PropTypes.string.isRequired,
+
+  // null is a valid uTL, which warns when `isRequired`.
```
victorporof: Don't think we need those comments here.

app/ui/browser/views/page/status.jsx
```diff
@@ -46,7 +46,8 @@ export function Status({ statusText }) {
 Status.displayName = 'Status';
 
 Status.propTypes = {
-  statusText: PropTypes.string.isRequired,
+  // null is a valid statusText, which warns if `isRequired`.
```
victorporof: Ditto.

app/ui/browser/actions/main-actions.js
jsantell: This seems a bit strange, but if it's necessary to have all these mutations in one tick/render, then it's a way to go. I just don't think it's necessary that all these mutations happen in one render (they could already get batched, and the views should handle it in a performance-friendly way, rather than in middleware, but I don't feel too strongly.

bgrins:
Just tested this out locally, LGTM

bgrins:
This landed

Issue #357, opened 2016-05-04
Sourcemaps are only built for files during development builds


Mossop:
Right now we only build sourcemaps when transpiling files for development builds but we also skip transpiling files if they haven't changed since the last time. To make matters worse, test runs do a non-development build.

So a file will only have a valid sourcemap if you do a development run immediately after changing it.

We should probably just always do source maps, since they aren't embedded I don't think it buys us much.

Mossop:
Fixed by #525

victorporof:
@Mossop Actually fixed by https://github.com/mozilla/tofino/pull/522

PR #522, opened 2016-05-24
Use react optimize, fixes #255

### LINKED ISSUE #255: Consider using babel-react-optimize
With #248, we'll be updating to `react@15.0.1`. This means we should be able to take advantage of perf optimizations for free by using [babel-react-optimize](https://github.com/thejameskyle/babel-react-optimize).

victorporof:
https://github.com/mozilla/tofino/issues/255

@Mossop, this involves a few build changes, mind having a look?


victorporof:
See individual commits for a step by step explanation of what should change.

build/tasks.js
```diff
@@ -30,45 +30,45 @@ export default {
   },
 
   async run(args = []) {
-    await this.build();
+    await this.build([...args, '--force']);
```
Mossop: This is fixing #357 and I think it's fine to do it like this for now. What might be nice in the future is to load build-config first, then diff it against the new build-config and in that case turn force on.

Mossop:
LGTM

build/tasks.js
victorporof: Sounds good!

victorporof: Maybe re-purpose that issue to follow this, or close that and file a new one instead?

Mossop: I've opened #528

Issue #350, opened 2016-05-04
Expose/separate browser devtools from web devtools


darkwing:
I went to "Tools > Toggle Developer Tools" and somehow Web Inspector loses context and I end up with a representation of the Tofino internals and not the page/site itself:

<img width="1501" alt="webview" src="https://cloud.githubusercontent.com/assets/46655/15003486/f223c640-1171-11e6-9754-9e800117df17.png">

jsantell:
We should probably be explicit that this is for browser development, not web development and expose a traditional web dev tools option here as well (and rename the current one)

bgrins:
I think we should use a different shortcut for the 'browser toolbox' version (maybe cmd/ctrl+alt+shift+i), and possibly have it auto-open when running via `npm run dev`.  Then keep the current shortcut reserved for the web page inspector.

victorporof:
Right-click and 'Inspect element' already works. We might want to have better UI for this though at some point.

PR #530, opened 2016-05-24
Separate Browser Toolbox from Developer Tools.

ncalexan:
@victorporof just the last commit.  Feel free to suggest better hotkeys, too -- no reason to vary from Firefox, I guess.

This should address the messaging around #350, although I agree that `Right Click > Inspect Element` is faster way to open the content devtools than the menu item.


victorporof:
🚢 shipit

PR #536, opened 2016-05-24
Add newtab page. Fixes #533. r=vp

### LINKED ISSUE #533: Implement a tofino://newtab page that links to the history, starred and saved pages

bgrins:


bgrins:
screenshot 2016-05-24 16 36 30

app/ui/about/newtab.html
```diff
@@ -0,0 +1,21 @@
...
+<body>
+
+Tofino: <a href="https://github.com/mozilla/tofino">https://github.com/mozilla/tofino</a>
+<ul>
+  <li><a href="tofino://history">History</a></li>
+  <li><a href="tofino://stars">Stars</a></li>
+  <li><a href="tofino://saves">Saves</a></li>
+</ul>
+
+</body>
```
victorporof: Wonder if it's worth having these pages in React. I'd suggest it is.

bgrins: Not sure I follow, what do you mean?

victorporof: Instead of having plain HTML files for `about:` pages, we should write them in React. That means the newtab page as well, since we're likely to make this much richer in the future.

victorporof: Currently all `tofino://foobar` pages are plain HTML. We should stop doing this and be consistent across our application code and use React everywhere. I'm working on converting the old pages at the moment, so any new page should follow this.

bgrins: OK, we can wait to land this until a pattern is in place for the history page

victorporof:
OK, closing this for now.

PR #541, opened 2016-05-25
Remove selector cruft from pages reducer. r=bgrins

linclark:
These selectors were moved in c2150a2, but were reintroduced in a rebase.


bgrins:
r+

PR #542, opened 2016-05-25
Use babel transforms in renderer tests. r=vporof

jsantell:


jsantell:
Seeing if this still fails on Linux/Windows.

victorporof:
r+ if passes.

PR #546, opened 2016-05-25
Separate out BrowserWindow logic and duplicated code from browser.js …

jsantell:
…into its own module. r=ncalexander

This cleans up browser.js a bit, which should just be plumbing the main components, and pushes BW logic into a module.

### DIFF
--- a/app/main/browser-window.js
+++ b/app/main/browser-window.js
@@ -1,0 +1,121 @@
+/*
+Copyright 2016 Mozilla
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of the
+License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed
+under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+*/
+
+import path from 'path';
+import electron from 'electron';
+
+import * as hotkeys from './hotkeys';
+import { UI_DIR, fileUrl } from './util';
+import BUILD_CONFIG from '../../build-config';
+
+const BROWSER_CHROME_URL = fileUrl(path.join(UI_DIR, 'browser', 'browser.html'));
+const BrowserWindow = electron.BrowserWindow;  // create native browser window.
+
+/**
+ * Top-level list of Browser Windows.
+ * @TODO Is this still needed? Electron can return all of its browser windows.
+ */
+
+const browserWindows = [];
+
+/**
+ * Take a promise to `ProfileStorage`. Remove this once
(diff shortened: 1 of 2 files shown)


app/main/browser-window.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+export function filterInvalidBrowserWindows(handler) {
+  return function(event, ...args) {
+    if (!event || !event.sender) {
+      return;
+    }
+
+    const bw = BrowserWindow.fromWebContents(event.sender);
+
+    if (!bw) {
```
ncalexan: Consider also `&& !bw.isDestroyed()`.

app/main/browser-window.js
```diff
@@ -0,0 +1,121 @@
...
+  browserWindows.splice(index, 1);
+
+  bw.close();
+}
+
+/**
+ * Takes a function to be used as a handler for an ipc event,
+ * and discards the event if we cannot identify a browser window
+ * from which the event came. If successful, the handler is called with
+ * the `BrowserWindow` instance rather than the `event` as the first argument.
```
ncalexan: I would prefer to include `event`, just so that adding/removing `filterInvalid...` is easier.

app/main/browser-window.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+const browserWindows = [];
+
+/**
+ * Take a promise to `ProfileStorage`. Remove this once
+ * UA service is in its own process. Also takes an onload callback --
+ * currently used for the first window created to record load times.
+ */
+export async function createBrowserWindow(profileStoragePromise, onload) {
```
ncalexan: It appears we're always invoking `createBW` in an async function.  If so, just pass `profileStorage` -- or consider passing the window scope directly, moving the profile usage out to the caller, which is probably closer to what we want over time.

app/main/browser-window.js
ncalexan: In addition, can we make this more descriptive of the outcome?  Something like `onlyWhenFromBrowserWindow` or something?  The plural windows really throws me off.

ncalexan:
I have some suggestions; feel free to ignore or apply as you see fit.

I expect this will clash with #516, but I'll let others work out the landing order.

app/main/browser-window.js
jsantell: I figure this'll be a moot point with #516 landed, although I didn't look at the implementation, it should be straight forward and remove the coupling all together

app/main/browser-window.js
jsantell: :+1:

app/main/browser-window.js
jsantell: `onlyWhenFromBrowserWindow` sounds good.

We'd still need to reapply the boilerplate of getting the BW from BrowserWindow.fromWebContents(event.sender), and it doesn't seem like something we'll be toggling often -- either it's a BW-based event, or not, yeah?

PR #552, opened 2016-05-26
Add a credits page listing the licenses for Tofino, Electron, ffmpeg and all the libraries that Chromium depends on.

### LINKED ISSUE #293: Have an about:license page

Mossop:
Fixes #293


Mossop:
See the credits page this adds here: https://rawgit.com/Mossop/tofino/5a07c7a90ac9e60e65d163f45ce529bb4f56a9ae/app/ui/about/credits.html

victorporof:
I believe this needs a rebase now that we're serving about pages over HTTP.

Mossop:
Legal have approved the credits page so we can land this if the code looks good.

app/main/about-pages.js
```diff
@@ -0,0 +1,43 @@
...
+ * app's 'ready' event.
+ *
+ * @TODO considerations:
+ * @TODO Use `protocol.registerHttpProtocol` so we can redirect invalid requests
+ * to default search?
+ * @TODO Proxy valid `chrome://*` URIs to `tofino://*`?
+ *   https://github.com/electron/electron/issues/2618
+ */
+export default function() {
+  protocol.registerFileProtocol(ABOUT_PROTOCOL_PREFIX, (request, callback) => {
```
victorporof: Should either register a http protocol now that that works, or make use of the fact that we're serving static content already via express, so no need to register a file protocol, right?

Mossop: Oh I think this ended up in here because of a weird rebase experience. I'll get rid of it.

victorporof: Feel free to land afterwards.

Issue #540, opened 2016-05-25
Ensure that render functions do not write to the DOM, etc


linclark:
We currently have some code interacting with the DOM in render functions, eg the [Page](https://github.com/mozilla/tofino/blob/master/app/ui/browser/views/page/page.jsx#L91) component. This work should happen elsewhere, as described in the [docs](https://facebook.github.io/react/docs/component-specs.html).

> The render() function should be pure, meaning that it does not modify component state, it returns the same result each time it's invoked, and it does not read from or write to the DOM or otherwise interact with the browser (e.g., by using setTimeout). If you need to interact with the browser, perform your work in componentDidMount() or the other lifecycle methods instead. Keeping render() pure makes server rendering more practical and makes components easier to think about.

linclark:
I was wrong about what is going on inside Page.jsx. It's actually creating a handler in `render()`.

Even though it's not the problem I thought it was, this is still a problem since the function is recreated on each render. There's a good explanation of why this is a problem in [React.js pure render performance anti-pattern](https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f#.qkwomzbnz) (see the *Functions create identities too* section).

jsantell:
No longer an issue with new backend?

victorporof:
Not sure what the backend has to do with this, I assume you mean frontend?
This isn't an issue anymore, but we still don't have any way to enforce this.

PR #553, opened 2016-05-26
[WIP] Best practice fixes to render functions

linclark:
WIP for #540


victorporof:
I think we should try enforcing a rule that disallows creating function props in components' render methods. In some near future when we'll start embracing `shouldComponentUpdate`, both automatic shallow and deep equality checks would usually improperly cause components to rerender just because some functions were recreated.

victorporof:
@linclark are you still working on this or should I close?

linclark:
We can close it out.

PR #554, opened 2016-05-26
Use Babel's transform-class-properties rule. r=vporof

linclark:
This enables us to declare and bind handlers in one go.

For example, the following code in [browser.jsx](https://github.com/mozilla/tofino/blob/master/app/ui/browser/views/browser.jsx#L35):

```
class BrowserWindow extends Component {
  constructor(props) {
    super(props);
    this.handleKeyDown = this.handleKeyDown.bind(this);
  }
  ...
  handleKeyDown(ev) {
    // function body here
  }
}
```

becomes:

```
class BrowserWindow extends Component {
  ...
  handleKeyDown = (ev) => {
    // function body here
  }
}
```

How would folks feel about that?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [ "react" ],
-  "plugins": [ "transform-es2015-destructuring", "transform-es2015-parameters", "transform-es2015-modules-commonjs", "transform-async-to-generator", "transform-object-rest-spread" ],
+  "plugins": [ "transform-es2015-destructuring", "transform-es2015-parameters", "transform-es2015-modules-commonjs", "transform-async-to-generator", "transform-object-rest-spread", "transform-class-properties" ],
   "sourceMaps": "inline",
   "retainLines": true
 }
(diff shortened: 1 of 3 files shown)


victorporof:
`handleKeyDown = (ev) => {` looks (and it is) magical, but I'm wary of it being too similar to something like `handleKeyDown: (ev) => {` which does means completely different.

I'd be much more in favor of using the `core-decorators` package and decorate event handlers with `@autobind`, which does the same thing but looks more explicit.

victorporof:
AFAIK, currently we'd have to use the `"transform-decorators-legacy"` plugins, with the `"core- decorators"` package from npm to use `@autobind`.

linclark:
We discussed this in Slack and I *think* we settled on going this way (at least for now since the decorators stuff is in flux). @victorporof lmk if I misunderstood (and also r? if I did not)

victorporof:
r+ :ship

victorporof:
Please file a bug about the eslint rule too.

linclark:
FYI, the eslint rule issue that I think you were talking about is #561

victorporof:
@linclark I was referring to an eslint rule forbidding `foo: () => {}` declarations on classes, since they're too similar to  `foo = () => {}` which can lead to bugs. #561 is also a good idea.

PR #558, opened 2016-05-26
UserServiceClient

jsantell:
This separates out the UAClient starting the server if not found (which means renderer process no longer uses child_process), and wraps it up into a standalone class.

We still need to consider the plan where it cannot connect (like a remote service that's down), etc.


app/shared/user-agent-client.js
```diff
@@ -37,98 +35,81 @@ function uaRequest(service, options) {
...
-      delete data.message;
-      this.emit(message, data);
-    });
+    this.url = url;
   }
 
-  startSession() {
-    return uaRequest('/session/start', {
+  startSession(url = endpoints.UA_SERVICE_HTTP) {
+    return uaRequest(url, '/session/start', {
```
Mossop: Random thought, rather than needing all callers to await connect() first we could just do it in here.

app/main/browser.js
```diff
@@ -36,19 +36,21 @@ import * as hotkeys from './hotkeys';
...
 const ipc = electron.ipcMain;
 const globalShortcut = electron.globalShortcut;
-const userAgentPromise = userAgentClient.connect();
+const userAgentClient = new UserAgentClient();
 
 const appStartupTime = Date.now();
 instrument.event('app', 'STARTUP');
 
+// Start the UA service running on a different process
+spawn.userAgentService();
```
Mossop: So this always attempts to start up the new process. What happens the second time that happens?

jsantell: Meaning another instance of Tofino starting up?

app/shared/user-agent-client.js
jsantell: Good call, even easier since this is the only exposed method really

app/main/browser.js
jsantell: In that case it'd fail but additional clients would connect to the original one. If that's the case, we should rethink a bit how additional Tofinos will find and connect to the local UA service (we don't have any port cycling, but we will need to, and the additional tofinos need to know where to find it)

Mossop: That is what we want, additional clients to connect to the existing service. What makes them fail, just the lack of available port? That seems fine for now.

jsantell: Yeah they'll probably just fail since they can't use the port.

Tried it out with two Tofinos (or Twofinos), and It Just Works:tm:

jsantell:
Closed via 47c6c1df99d16e099183699f57a6c4d134ff45f0

PR #560, opened 2016-05-26
Remove all trace of flow from the source.

### LINKED ISSUE #559: Remove flow type annotations
Since flow doesn't work on windows and we aren't getting much value from it we've decided to remove the type annotations and no longer rely on this.

Mossop:
Fixes #559.


ncalexan:
I didn't look too closely, but if it passes tests, it's fine by me!

PR #563, opened 2016-05-27
Start a REPL after launching the UA service.

rnewman:
After this you can run

```
./node_modules/.bin/babel-node lib/main/server.js
```

to get a REPL running the UA service.

### DIFF
--- a/app/services/user-agent-service/bin/user-agent-service
+++ b/app/services/user-agent-service/bin/user-agent-service
@@ -1,9 +1,16 @@
 #!/usr/bin/env node
 
 var UserAgentService = require('../').UserAgentService;
+var repl = require('repl');
 
 var argv = require('yargs')
   .usage('Usage: $0 [options]')
+  .option('r', {
+    alias: 'repl',
+    require: false,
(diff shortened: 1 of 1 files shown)


rnewman:
Something about the build config check has stopped working after your latest changes landed, @jsantell. Is the spawning stuff chewing up the environment?

Mossop:
Can we add an "npm run ua" to start this?

jsantell:
build-config shouldn't change based on process, what are you seeing?
If #565 looks good, maybe we can just roll this into another option (and then can expose either via `npm run ua` like @Mossop says)

rnewman:
Perhaps it's down to how I'm invoking the process? The development check wasn't true.

And yes, @mossop, that'd be great.

I'll take a look at your PR after this fx-team call, @jsantell.

rnewman:
Ah, we don't default to `development`. That'll do it. This'll rebase nicely on top of #565.

jsantell:
Closed via 01d195dc140a9927dfc1d2fa669a283cab8c79aa

PR #565, opened 2016-05-27
Wrap UserAgentService into a file being spawned by an executable scri…

jsantell:
…pt that can be called independently of the main process. r=rnewman

Now we can call `./lib/services/user-agent-service/bin/user-agent-service -p 9090 -d .` directly -- another step into making this process more independent.

### DIFF
--- a/app/main/spawn.js
+++ b/app/main/spawn.js
@@ -12,20 +12,27 @@ specific language governing permissions and limitations under the License.
 
 import { spawn } from 'child_process';
 import path from 'path';
+import * as endpoints from '../shared/constants/endpoints';
 
-const UA_SERVICE_PATH = path.join(__dirname, '..', 'services', 'user-agent-service', 'index');
-const CONTENT_SERVICE_PATH = path.join(__dirname, '..', 'services', 'content-service', 'index');
+const DB_PATH = path.join(__dirname, '..', '..');
+const SERVICES_PATH = path.join(__dirname, '..', 'services');
+const UA_SERVICE_BIN = path.join(SERVICES_PATH, 'user-agent-service', 'bin', 'user-agent-service');
+const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.js');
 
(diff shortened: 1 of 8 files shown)


app/main/spawn.js
```diff
@@ -12,13 +12,21 @@ specific language governing permissions and limitations under the License.
...
+                       'user-agent-service', 'bin', 'user-agent-service');
+const DB_PATH = path.join(__dirname, '..', '..');
 
 // @TODO Use something like `forever` or `pm2` to ensure that this
 // process remains running
+// @TODO Will `node` be accessible via path on users machines that don't
+// have node installed as an engineer?
 export function userAgentService() {
-  spawn('node', [UA_SERVICE_PATH], {
+  spawn('node', [UA_SERVICE_BIN,
```
rnewman: In order to get an ES6 repl here, we need to use `babel-node`.

On the plus side, `babel-node` will be in `node_modules`, and takes care of finding the environment `node` for us. Perhaps use that here?

app/services/user-agent-service/README.md
```diff
@@ -0,0 +1,7 @@
+```
```
rnewman: <3 <3

app/services/user-agent-service/index.js
```diff
@@ -10,18 +10,26 @@
...
-import path from 'path';
 import * as userAgentService from './server';
 import { ProfileStorage } from './storage';
-import * as endpoints from '../../shared/constants/endpoints';
 
 process.on('uncaughtException', console.error);
 process.on('unhandledRejection', console.error);
 
-ProfileStorage.open(path.join(__dirname, '..', '..')).then(async function (profileStorage) {
+export async function UserAgentService(options = {}) {
```
rnewman: `app/main/` has `userAgentService()`, and we have `UserAgentService` here. Can we reduce the collision by renaming one of them?

rnewman:
r+ from me, @jsantell.

app/services/user-agent-service/index.js
jsantell: Ah yes, good call

PR #569, opened 2016-05-27
Further decouple user agent service by only depending on profile diff…

jsantell:
… actions, and add version and content service origin options to the CLI. r=rnewman

### DIFF
--- a/app/services/user-agent-service/index.js
+++ b/app/services/user-agent-service/index.js
@@ -12,6 +12,7 @@
 
 import * as userAgentService from './server';
 import { ProfileStorage } from './storage';
+import meta from './meta.json';
 
 process.on('uncaughtException', console.error);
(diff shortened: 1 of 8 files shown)


app/services/user-agent-service/index.js
```diff
@@ -23,16 +24,26 @@ export async function UserAgentService(options = {}) {
   if (typeof options.db !== 'string') {
     throw new Error('UserAgentService requires a `db` string.');
   }
+  if (meta.validVersions.indexOf(options.version) === -1) {
```
rnewman: Use `Array.prototype.includes`?

rnewman:
LGTM.

app/services/user-agent-service/index.js
jsantell: good call, fancy new methods

Issue #555, opened 2016-05-26
Use webpack to allow importing modules in the about pages


victorporof:
Let's land webpack again 💯 

This blocks rewriting the about pages in React, which blocks https://github.com/mozilla/tofino/issues/518

PR #573, opened 2016-05-27
Webpack content pages and serve them from the lib directory

victorporof:
@Mossop you wanted to have a look at this.

Fixes https://github.com/mozilla/tofino/issues/555


build/main.js
```diff
@@ -53,7 +53,6 @@ const handleDepsCheckSucceeded = () => {
   const argv = process.argv;
   const tasks = require('./tasks').default;
   const handlers = {
-    '--build': args => tasks.build(args),
```
Mossop: We use this for our postinstall step so after download/install users mostly have everything pre-compiled already. Not against dropping that if that is complex to do now though.

build/tasks.js
```diff
@@ -1,43 +1,40 @@
...
+
+const Lazy = {
+  buildDeps: () => require('./task-build-deps').default(),
+  config: options => require('./task-config-builder').default(options),
+  build: args => require('./task-build').default(args),
+  run: args => require('./task-run').default(args),
+  test: args => require('./task-test').default(args),
+  package: () => require('./task-package').default(),
+  clean: () => require('./task-clean-package').default(),
+};
```
Mossop: Can you elaborate on why this is helpful, it isn't clear to me.

build/utils.js
```diff
@@ -66,9 +69,7 @@ export function getElectronVersion() {
   return version.trim().substring(1);
 }
 
-/**
- * Use a widnows `.cmd` command if available.
- */
+// Use a widnows `.cmd` command if available.
```
Mossop: Let's correct the spelling of windows here while you're changing blame already.

build/webpack.config.content.prod.js
```diff
@@ -0,0 +1,24 @@
...
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+import webpack from 'webpack';
+import contentConfig from './webpack.config.default.content';
+
+export default {
+  ...contentConfig,
+  plugins: [
+    ...contentConfig.plugins,
+    new webpack.NoErrorsPlugin(),
```
Mossop: Why wouldn't we use this all the time?

build/webpack.config.content.prod.js
```diff
@@ -0,0 +1,24 @@
...
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify('production'),
+      },
+    }),
+  ],
+  debug: true,
```
Mossop: It's unclear from webpack's docs what this does and why we'd want it in prod but not dev, can you elaborate?

build/webpack.config.content.prod.js
```diff
@@ -0,0 +1,24 @@
...
+  devtool: 'source-map',
+  output: {
+    ...contentConfig.output,
+    pathinfo: true,
```
Mossop: Webpack's docs say not to use this in production?

app/ui/content/index.js
```diff
@@ -0,0 +1,23 @@
...
+import { fetchRecentBookmarks } from './stars';
+
+if (~document.location.pathname.indexOf('history')) {
+  bindHistorySearch();
+  fetchRecentHistory();
+}
+
+if (~document.location.pathname.indexOf('stars')) {
+  fetchRecentBookmarks();
+}
```
Mossop: Even aside from these tests being susceptible to accidental breakage (if we add a "recent_history.html" for example) it doesn't seem right to have to have a single entry point for every content page we add.

build/task-build-browser.js
```diff
@@ -89,11 +89,18 @@ export async function buildFile(sourceFile, sourceStats, args = []) {
 
 export default async function babelBuild(args = []) {
   const source = path.resolve(path.join(__dirname, '..', 'app'));
-  const paths = await fs.walk(source);
+  let paths = await fs.walk(source);
+
+  // Exclude the `ui/content` dir. This is super ugly here, but we won't
+  // need any of this after we switch our entire build system to webpack.
```
Mossop: We also have glob as a dependency already, something like glob("app/**" { ignore: "app/ui/content/**"}) would probably do the trick.

build/main.js
victorporof: Ah! I somehow missed that. Will keep this.

build/tasks.js
victorporof: It looked pretty complicated to properly maintain the arguments/options and their order passed into require(foo).default(...) calls. Making sure there's only one way to invoke those default imports should prevent bugs where we don't pass in the correct args.

build/utils.js
victorporof: I thought I already did that! Oops.

build/webpack.config.content.prod.js
victorporof: Made it so that we do.

build/webpack.config.content.prod.js
victorporof: This is a flag that makes webpack and (more importantly) plugins spew stuff to stdout when building when "bad things might be happening". I found a template webpackrc file that had this, so I turned it on too. But since we're not really sure we need this, I'll remove it altogether.

build/webpack.config.content.prod.js
victorporof: Removed it. Also enabled source maps for both production and development builds.

app/ui/content/index.js
victorporof: As discussed on Slack, this is _very_ temporary.

build/task-build-browser.js
victorporof: Ah nice!

PR #587, opened 2016-05-31
Make tofino work on windows again

### LINKED ISSUE #578: Tofino doesn't launch on Windows
Looks like since this patch, Appveyor has been permared -- could be a coincidence, but should investigate:
https://github.com/mozilla/tofino/commit/e8535dc6ef3fdbdd66590cfba194fd4089668ab7

@victorporof

Mossop:
For the UA service to be completely detached and outlive the parents on all platforms it can't share any stdio with the parent.

Apparently ignoring just stdin breaks process spawning on Windows.

Fixes #578

### DIFF
--- a/app/main/spawn.js
+++ b/app/main/spawn.js
@@ -31,15 +31,12 @@ export function startUserAgentService() {
     '--content-service', endpoints.CONTENT_SERVER_ORIGIN,
   ], {
     detached: true,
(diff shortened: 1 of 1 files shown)


Mossop:
I haven't testing this on OSX but this makes things work right on Windows.

jsantell:
Looks good -- confirmed it works in OSX with Twofinos

PR #589, opened 2016-05-31
Use a configurable port/version/host for connecting to UserAgentServi…

jsantell:
…ce to prepare it for having variable port numbers. r=nalexander

### DIFF
--- a/app/main/browser-window.js
+++ b/app/main/browser-window.js
@@ -32,9 +32,18 @@ const browserWindows = [];
  * currently used for the first window created to record load times.
  */
 export async function createBrowserWindow(userAgentClient, onload) {
-  await userAgentClient.connect();
   // TODO: don't abuse the storage layer's session ID generation to produce scopes.
+  // Await for `startSession()` here since that ensures we have a connection to
+  // the UA service at this point.
   const scope = await userAgentClient.startSession();
+
+  // Get the UA service address information from the user agent client
+  // after it has negotiated a connection so we can send it to the client.
+  const { port, host, version } = userAgentClient;
+
+  if (!port || !host || !version) {
+    throw new Error('The host, port, and version must be defined after connecting.');
+  }
 
   // Create the browser window.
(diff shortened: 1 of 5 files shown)


jsantell:
This is getting messy, fast. There are a lot of scenarios where we could lose connection or the service goes down and having 2+ Tofinos negotiate that is rough.

app/ui/browser/index.jsx
```diff
@@ -73,7 +73,6 @@ store.dispatch(actions.createTab());
 // that if an error occurs while we connect, we at least have some UI in place.
 ReactDOM.render(chrome, container);
 
-userAgentClient.connect().then(() => onWindowReady(true));
```
ncalexan: My guess is that this will not work in production mode: I don't see where you're calling `onWindowReady` now.  Development mode always shows the window, so you probably haven't witnessed this.

app/main/browser-window.js
```diff
@@ -52,8 +61,13 @@ export async function createBrowserWindow(userAgentClient, onload) {
     if (onload) {
       onload();
     }
+    // The client needs to know where the UA service is in order to connect
```
ncalexan: We might be able to avoid the round trip by having the UI fish host and service from it's initial URL, or from query parameters.  (I did this once before; it worked fine.)  Your choice.

ncalexan:
I see one serious issue, but otherwise looks okay.

app/main/browser-window.js
jsantell: Wanted to separate that out so we load the browser chrome first ASAP, incase we needed to wait for the address for the UA service. Probably a moot point if we don't intend on having that being too variable.

app/ui/browser/index.jsx
jsantell: The `onWindowReady` is still inside the first diff handler -- just tested in production mode, and we still have syncing bookmarks!

PR #600, opened 2016-06-02
Turn off arrow-body-style so arrow functions can be wrapped in blocks if chosen

bgrins:
@jsantell, we talked about this a while ago.  Here's a PR


jsantell:
beautiful, r+

PR #601, opened 2016-06-02
Looking for feedback: Disable logging for SET_STATUS_TEXT actions since they happen on every mousemove

bgrins:
Looking for feedback about this.  The console becomes very hard for me to scan because of the constant logging of the SET_STATUS_TEXT action (which happens whenever a link is hovered and then again when it's un-hovered).

I'd personally be fine with just not showing this action ever, even if that makes it a less correct action log and harder to debug SET_STATUS_TEXT calls.  That's what this PR does, but if others would rather keep it there I'm fine with closing.

### DIFF
--- a/app/ui/shared/configure-store.js
+++ b/app/ui/shared/configure-store.js
@@ -32,7 +32,11 @@ export default function(rootReducer, initialState) {
       duration: true,
       collapsed: true,
       predicate: (getState, action) => {
-        return typeof action !== 'function' && action.type !== types.SET_STATUS_TEXT;
+        if (action.type === types.SET_STATUS_TEXT) {
+          console.log(types.SET_STATUS_TEXT + " " + action.text);
+          return false;
+        }
+        return typeof action !== 'function';
       },
       stateTransformer: (state) => state.toJS(),
     }));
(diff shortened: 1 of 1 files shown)


bgrins:
Interested in any opinions, but specifically frontend-related ping to @jsantell @linclark @victorporof

victorporof:
How about a 'logging' build config option that can set verbosity?

jsantell:
Another solution would be just dump the action type as a string for this action, so we'd still get (x) bubbles next to dupes in devtools, so it's less noisy, but we still get an alert that the action did occur

bgrins:
> Another solution would be just dump the action type as a string for this action, so we'd still get (x) bubbles next to dupes in devtools, so it's less noisy, but we still get an alert that the action did occur

I like the idea so tried it locally and pushed it up to the PR.  It makes them look a little different than the other actions, although it definitely does collapse a lot of dupes in the output when mousing around the new tab page.  Not sure if this would be better or a build config option that makes them all disappear. 

We could also keep some state in this logger function so it allows the normal rendering, but only passes the predicate if the text is different than last time.

victorporof:
Wonder if `console.groupCollapsed` and `console.groupEnd` could also help. We could keep things in a collapsed section if things keep coming as the same action.

victorporof:
@bgrins any consensus on this? Should this PR land or be closed?

victorporof:
I'm closing this due to lack of activity. Feel free to reopen.

PR #606, opened 2016-06-03
Use class property initialization for binding. r=vporof

### LINKED ISSUE #605: Use class property initialization instead of binding in constructor
We can do this now that #554 has landed. I'm going to transition all of the existing instances of this so that its not inconsistent when I add new ones.

linclark:
Fixes #605


linclark:
Didn't think about this before, but we could move propTypes and displayName definitions into the class with this too...

anyone have an opinion about that?

victorporof:
🚢

victorporof:
@linclark I have no strong preference either way.

linclark:
ok, I'll just ship it as is. we can always switch that up later.

Issue #607, opened 2016-06-03
Killing off the UAS server once the last webscoket client disconnects is incorrect


victorporof:
We tie the UAS lifetime to how many tofinos are currently running. So if we run our first tofino, a UAS starts. If we start our second tofino, there’s a silent EADDRINUSE, but we still have the first server running. Once we close all tofinos, the server itself shuts itself down, because it keeps track of how many websockets connections are open. Once it goes back to 0, the server kills itself.

Tying the UAS lifetime to how many websocket connections are open at a given time seems incorrect. If i start the UAS myself in a terminal somewhere, i don’t expect my server to be shut down at some point simply because there’s no more clients remaining. If i start the server myself, i should control the lifetime.

e.g. We start our UAS standalone, outside of the main process, the server keeps running. Once we open a tofino somewhere else, there’s the silent EADDRINUSE, so we use that server that we opened. When we close our tofino, we also kill the server opened via the external command, which we don't own.

The core of the problem is that right now we don’t have an elegant way to kill servers, so we baked that logic into websocket connections lifetimes, which is ugly but works. Having a proper (correct) way to kill/restart servers seems like a better approach.

PR #608, opened 2016-06-03
Create an npm script for starting app services standalone, and make their lifetime configurable

victorporof:
Fixes https://github.com/mozilla/tofino/issues/607

@Mossop

### DIFF
--- a/app/main/spawn.js
+++ b/app/main/spawn.js
@@ -14,8 +14,9 @@ import { spawn } from 'child_process';
 import path from 'path';
 import * as endpoints from '../shared/constants/endpoints';
 
-const DB_PATH = path.join(__dirname, '..', '..');
-const SERVICES_PATH = path.join(__dirname, '..', 'services');
+const ROOT = path.join(__dirname, '..', '..');
+const DB_PATH = ROOT;
+const SERVICES_PATH = path.join(ROOT, 'lib', 'services');
 const UA_SERVICE_BIN = path.join(SERVICES_PATH, 'user-agent-service', 'bin', 'user-agent-service');
 const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.js');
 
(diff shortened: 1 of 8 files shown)


victorporof:
@jsantell you've been working on UAS stuff recently too, might want to be aware of this change.

app/main/spawn.js
```diff
@@ -13,38 +13,60 @@ specific language governing permissions and limitations under the License.
...
     '--db', DB_PATH,
     '--version', version,
     '--content-service', endpoints.CONTENT_SERVER_ORIGIN,
   ], {
     detached: true,
-    stdio: ['ignore'],
-  });
+    // Careful passing in options.verbose as true, as this stops the services from
+    // outliving the their original parents, which we want to allow in the UAS case.
+    stdio: options.verbose ? ['ignore', process.stdout, process.stderr] : ['ignore'],
```
jsantell: [This breaks Windows](e8047ffe146e8cd9bc586b4cefc11ad546b91ca2) (should add a comment w/r/t this)

jsantell:
What's the use-case for this? Couldn't we just use the [services CLI directly](https://github.com/mozilla/tofino/blob/master/app/services/user-agent-service/README.md)?

app/main/spawn.js
```diff
@@ -13,38 +13,60 @@ specific language governing permissions and limitations under the License.
...
+if (!BUILD_CONFIG.keepAliveAppServices) {
+  process.on('exit', () => killChildren('SIGTERM'));
+  process.on('SIGINT', () => killChildren('SIGTERM'));
+  process.on('SIGTERM', () => killChildren('SIGTERM'));
+}
 
-  process.on('exit', () => {
-    child.kill('SIGKILL');
-  });
+function killChildren(code) {
```
jsantell: We can probably have a better function name for this.

app/main/spawn.js
victorporof: You mean in addition to the comment above?

app/main/spawn.js
victorporof: See #607. The CLI doesn't start all services, and doesn't address the lifetime issues.

app/main/spawn.js
jsantell: The stdio options breaks Windows according to @mossop which was already reverted in the linked commit, so we should comment on that, and these changes would break Windows

app/main/spawn.js
jsantell: I just mean a better name than `killChildren`

victorporof: This is in reply to your "What's the use-case for this?" question.

victorporof:
Addressed comments and changed wording to make it clear that we do want to run the servers attached to their parent process on npm run serve. @jsantell, this doesn't break anything on windows, and I don't see any further reason for your concerns.

Issue #35, opened 2016-04-07
Tests should throw when PropType validation fails for a react component


jsantell:

victorporof:
@linclark do you know of any way react could notify us when a proptype validation fails, so that we may fail the tests automatically?

victorporof:
Also, note that now that we're using babel-preset-react-optimize in production mode, these prop types are deleted from production builds, which we're actually using for tests.

Should we test builds in development mode instead? Or have yet another build configuration which is for tests only, mostly the same as production but without babel-preset-react-optimize?

victorporof:
Found a middleground approach, see the PR.

PR #617, opened 2016-06-05
Use webpack for the main and renderer processes

victorporof:


victorporof:
@Mossop please take a look at this.

This PR simplifies our build scripts quite a bit and ends up using webpack for both the main process and the renderer process. It cleanups the directory structure as well (and it has to, because with the new architecture it's impossible to not put stuff in `app/shared` for example, if certain modules are imported both in main and renderer modules). Now that we're using webpack in these places, we can make use of its own watcher and caching mechanisms, so incremental builds are now faster.

I'm keeping straight babel without webpack for services, since those are not technically part of the main process nor renderer process modules. If we decide to use webpack for those too, we can do it in a followup.

I'm also fixing `nyc` from (redundantly) babeling the lib dir, this took a very long time. As a result, tests are now faster to start. We can further improve test startup times by building everything in 'development', since that emits dumber sourcemaps and takes advantage of other little things in our webpack configuration. I'm not sure if that's desirable though. Not doing this now, I'll leave that up for discussion.

As a further followup, we might consider using webpack-dev-server instead of the compiler's watch mechanism which writes files to disk and does an incremental build when something changes. Instead, webpack-dev-server would serve everything from memory, which would make our workflow even faster. It's pretty fast enough though already (incremental builds being about 100ms), so I'm not sure it's worth investigating at this moment.

victorporof:
One more theoretical advantage of doing this is that we'll be able to run renderer code in any browser (without the IPC functionality). We don't need this now, but it's useful to have it for free if we'll ever do.

victorporof:
Regarding startup times for tests, see also https://github.com/mozilla/tofino/issues/35#issuecomment-223889141 for another reason to use development builds instead of production for tests (or something similar).

package.json
```diff
@@ -44,7 +44,7 @@
     "redux": "3.5.1",
     "redux-immutable": "3.0.6",
     "redux-logger": "2.6.1",
-    "request": "2.71.0",
+    "request": "2.65.0",
```
Mossop: Why the downgrade?

build/webpack.config.browser.default.js
```diff
@@ -0,0 +1,50 @@
...
+    new CopyWebpackPlugin([{
+      from: path.join(shared, 'assets'),
+      to: path.join(dest, 'assets'),
+      flatten: true,
+    }]),
+    new CopyWebpackPlugin([{
+      from: path.join(shared, 'fonts'),
+      to: path.join(dest, 'fonts'),
+      flatten: true,
+    }]),
```
Mossop: Can we share the shared pieces? Is it worth it for two uses or will there be more?

package.json
victorporof: `request` > 2.65 fails with https://github.com/request/request/issues/1920 in webpack because it depends on json-schema which uses a non-standard (incorrect) amd export which apparently didn't get fixed yet.

victorporof: See also https://github.com/webpack/webpack/issues/138 and https://stackoverflow.com/questions/35236986/config-webpack-with-request

build/webpack.config.browser.default.js
victorporof: Don't know if it's really worth it, being straightforward code and all. The browser and content don't necessarily have to have the same rules either.

Mossop:
LGTM

victorporof:
Thanks!

Issue #618, opened 2016-06-05
Many promise rejections are unhandled in tests, and we're not actually logging them


victorporof:
Some of them are about refused express server connections, while others are in fact more important errors that we should fix. Take a look at the PR.

victorporof:
Here's an output: https://ci.appveyor.com/project/Mozilla/tofino-u1hv8/build/1.0.519-hssexhts/job/alplxntj1g0a2k8c

It's grim. 😨

victorporof:
Here's a PR that doesn't use thenify: https://github.com/mozilla/tofino/pull/656

ncalexan:
I've investigated this, and I concur with https://github.com/mozilla/tofino/issues/618#issuecomment-223832937: it's grim.

I see two things at work here.  The first is the interaction from the UI to the UA service (as I expected and suggested to @victorporof).  The UI expects the UA service to be available (reasonable!) but doesn't handle the situation when it's not.  That's fine at runtime, but fails during tests, when the UA service is explicitly not running.  The solution to this is mostly routine: we chase all methods in `app/ui/browser/lib/user-agent.js` and handle promise rejections.  Some rejections can be safely ignored (`createHistory`); some will lead to bad UI state (`createStar` will show an optimistic star in the UI that may never be retracted); and some should probably be explicitly rescheduled (`createSession`).

The second is that the tests invoke actions that are inherently asynchronous, so we witness callbacks on ticks _after_ the invoking test has completed.  This manifests in unhandled `AssertionError` rejections.  What is happening is that an action (`createTab`) is invoked, which updates the Redux state and then requests a new session to be started.  The test verifies that the Redux state is updated and then moves on.  Afterwards, the session request succeeds (or fails!) and the action code tries to update the Redux state, which is no longer valid.

I think the UI team needs to own the asynchronous, possibly fallible, interaction with the UA service.  In extremis, this might require us to declare the UA service is always available and to run it (or mock it) during tests.

As for the asynchronous tests, the best approach I've seen is https://github.com/yelouafi/redux-saga.  Others will have much more experience with this than me, however.

@jsantell, @victorporof -- your thoughts?

ncalexan:
I started tracing locations that would need workarounds, and added some code that let me see rejections in my IDE (which doesn't invoke using `build/main.js`).  Commit is https://github.com/ncalexan/tofino/commit/4ebd9fa05a06e41026441f2a81782266a5558093.

jsantell:
[...]

PR #619, opened 2016-06-05
Looking for feedback: Log promise rejections in build scripts

victorporof:

https://github.com/mozilla/tofino/issues/618

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -21,22 +21,36 @@ const semver = require('semver');
 const checkDependencies = require('check-dependencies');
 const VALID_NODE_VERSION_RANGE = require('../package.json').engines.node;
 
+process.on('uncaughtException', (err) => {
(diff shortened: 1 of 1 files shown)


victorporof:
Worth noting is that the only reason the tests don't fail now with logging enabled is because we buffer the console.log output via `process.stdout.once('drain', () => process.exit(1));`, so mocha manages to kill our process before ourselves exiting with an error, so the tests look like they pass.

If we're not draining stdout, the output gets trimmed after one unhandled rejection being logged, since the process is killed immediately and the tests fail instantly.

victorporof:
Output: https://ci.appveyor.com/project/Mozilla/tofino-u1hv8/build/1.0.519-hssexhts/job/alplxntj1g0a2k8c
It's grim. 😨

PR #620, opened 2016-06-06
Fail tests when prop type validations fail

victorporof:
Fixes https://github.com/mozilla/tofino/issues/35

This also makes test builds still compile in production mode but without babel-react-optimize.
Depends on https://github.com/mozilla/tofino/pull/617

Only commit a198e9d is relevant.

@jsantell

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,6 +5,9 @@
     },
     "development": {
       "presets": ["react"]
+    },
+    "test": {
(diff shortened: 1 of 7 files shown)


victorporof:
To make sure this works, just remove `onClick` handlers from `test-btn.js`.

jsantell:
r+

PR #626, opened 2016-06-06
Use electron as a node process.

### LINKED ISSUE #610: Tofino can't run unless node is installed on the system
Because the UA and content services are both spawned node instances they won't work unless node is installed and in the path for the system. We should make them use electron as a node instance instead.

Mossop:
Fixes #610.

This is not yet ready for review as it doesn't work!

### DIFF
--- a/app/main/spawn.js
+++ b/app/main/spawn.js
@@ -24,8 +24,6 @@ const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.
 
 // @TODO Use something like `forever` or `pm2` to ensure that this
 // process remains running
-// @TODO Will `node` be accessible via path on users machines that don't
-// have node installed as an engineer?
 
 const children = new Set();
 function spawnProcess(name, command, args, options) {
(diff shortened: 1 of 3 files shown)


app/main/spawn.js
```diff
@@ -23,8 +24,15 @@ const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.
 
 // @TODO Use something like `forever` or `pm2` to ensure that this
 // process remains running
-// @TODO Will `node` be accessible via path on users machines that don't
-// have node installed as an engineer?
+
+export function getElectronExecutable() {
```
Mossop: This should be smarter and use process.execPath when being run from electron already. Arguably we could always use process.execPath but I'd prefer to run from electron even when manually starting the services for consistent testing.

Mossop:
Ugly, but working

jsantell:
Great! If the maintainers have any interest in supporting this, we could maybe expose the build config in a constructor to make this easier (but if they don't accept, could make the fork being up to date harder) via something like `require('sqlite3').verbose({ target: ELECTRON_VERSION, runtime: 'electron' })`

PR #627, opened 2016-06-06
Fix opening of a bookmark via menu. r=vporof

jsantell:
When clicking on a bookmark in the menu, we get an error, expecting immutable objects.

```
TypeError: bookmark.toJS is not a function
    at click (webpack:///./app/main/menu/submenu-items.js?:133:64)
    at MenuItem.click (/Users/jsantell/Dev/tofino/.electron/Electron.app/Contents/Resources/electron.asar/browser/api/menu-item.js:81:16)
    at Function.executeCommand (/Users/jsantell/Dev/tofino/.electron/Electron.app/Contents/Resources/electron.asar/browser/api/menu.js:119:40)
```


victorporof:
r+

PR #631, opened 2016-06-06
Ipc renderer tests

jsantell:


jsantell:
Start of tests that map ipc-renderer events to the underlying actions they create -- the unit action tests test the outcome of the action, this only the actions are fired.

victorporof:
Very nice r+

PR #636, opened 2016-06-07
Remove scroll listener preload script that is no longer used. r=vporof

jsantell:


package.json
```diff
@@ -33,7 +33,6 @@
     "express-ws": "2.0.0-rc.1",
     "fs-promise": "0.5.0",
     "immutable": "3.7.6",
-    "lodash": "4.13.1",
```
victorporof: This is surprising. We still use lodash in tests though.

app/shared/preload/scroll.js
```diff
@@ -1,24 +0,0 @@
...
-import { ipcRenderer as ipc } from 'electron';
-import throttle from 'lodash/throttle';
-
-// How often the scroll handler may be fired at most
-const THROTTLE_TIMER_IN_MS = 50;
-
-window.addEventListener('scroll', throttle(() => {
-  ipc.sendToHost('scroll', {
-    x: window.scrollX,
-    y: window.scrollY,
```
victorporof: I can imagine us wanting to use this again at some point. Would personally like to keep it, but it's your choice.

package.json
victorporof: I wonder why removing this line didn't make the tests fail. Linting tests use `lodash/without` and `lodash/difference`

jsantell: Huh. Weird. I'll add it back as a devDependency atleast for now

victorporof: See https://github.com/mozilla/tofino/issues/647 
I think node_modules should be clobbered.

PR #637, opened 2016-06-07
Do not lint dist/ directory, and move node_module ignoring to the .es…

jsantell:
…lintignore file for consistency. r=vporof


victorporof:
r+

PR #645, opened 2016-06-07
Prevent status text from overflowing, and keep it just one line high.…

### LINKED ISSUE #199: Link status should be trimmed when the URL is wider than the browser window
See screenshot

wrapping text

jsantell:
… Fixes #199. r=vporof


victorporof:
r+

PR #646, opened 2016-06-07
Remove 'thenify' dependency, since 'thenify-all' also includes and ex…

### LINKED ISSUE #503: We're depending on both `thenify` and `thenify-all`
Is there a reason for this? AFAIK `thenify-all` can also be used on single functions, or at least exports a `thenify` method that works on a single function.

jsantell:
…poses it. Fixes #503. r=vporof

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -16,7 +16,7 @@
 require('babel-polyfill');
 require('babel-register')();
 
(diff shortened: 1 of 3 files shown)


victorporof:
r+

PR #648, opened 2016-06-08
Split completion row into own component. r=bgrins

### LINKED ISSUE #604: Split completion row into its own component
This is required to avoid binding in the render function. See #602

linclark:
Fixes #604

### DIFF
--- a/app/ui/browser/views/navbar/location-completion-row.jsx
+++ b/app/ui/browser/views/navbar/location-completion-row.jsx
@@ -1,0 +1,72 @@
+/*
+Copyright 2016 Mozilla
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of the
+License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed
+under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+*/
+
+import React, { PropTypes } from 'react';
+
+import Style from '../../../shared/style';
+
+import * as actions from '../../actions/main-actions';
+
+const LOCATION_COMPLETION_ROW_STYLE = Style.registerStyle({
+  color: 'rgba(100, 100, 100, 1)',
+  padding: '0.5em',
+  whiteSpace: 'pre-wrap',           // So that spaces show up in snippets.
+  zIndex: 2,
+});
+
+/**
+ * A completion row in the address bar.
+ */
+export function LocationCompletionRow({ completion, focusedResultIndex, index }) {
+  // We get safe, decorated (<b>foo</b>) HTML from the database.
+  const snippet = completion.snippet ? (
+    <div
+      className={LOCATION_COMPLETION_ROW_STYLE}
+      dangerouslySetInnerHTML={{          // eslint-disable-line react/no-danger
+        __html: completion.snippet,
+      }}>
+    </div>
+  ) : null;
+
+  return (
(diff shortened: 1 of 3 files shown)


app/ui/browser/views/navbar/location-completion-row.jsx
```diff
@@ -0,0 +1,72 @@
...
+      }}>
+    </div>
+  ) : null;
+
+  return (
+    <div className={"completion-row"}
+    style={{
+      // We're nesting multi-line items inside a flexbox, so
+      // we need to mark these children as columnar.
+      flexDirection: 'column',
```
victorporof: Since you're splitting this component out anyway, I suggest moving this to our style definition above to keep things consistent with using freestyle everywhere. No reason to have this inline afaict.

app/ui/browser/views/navbar/location-completion-row.jsx
```diff
@@ -0,0 +1,72 @@
...
+    }}>
+      <div
+        onMouseDown={(ev) => { ev.preventDefault(); }}
+        onMouseOver={() => { this.props.dispatch(actions.setFocusedResultIndex(index)); }}
+        onClick={() => {
+          this.selectAutocompleteItem(completion.uri);
+        }}
+        style={focusedResultIndex === index ? {
+          background: 'var(--theme-selection-color)',
+          color: 'var(--theme-content-selected-color)',
```
victorporof: Ditto for this maybe: set a classname instead of inline styles.

victorporof:
Really happy to see this PR @linclark

linclark:
Good call, I extracted the styles out and that makes things much clearer. If anyone has suggestions for other const names for them, lmk.

@bgrins, I want to make sure that you have a chance to review this before we land it. No rush, I'll be off doing this for a while 🚴🌅

app/ui/browser/views/navbar/location-completion-row.jsx
```diff
@@ -0,0 +1,75 @@
...
+    </div>
+  ) : null;
+
+  return (
+    <div className={`completion-row ${ROW_STYLE}`}>
+      <div
+        onMouseDown={(ev) => { ev.preventDefault(); }}
+        onMouseOver={() => { this.props.dispatch(actions.setFocusedResultIndex(index)); }}
+        onClick={() => {
+          this.selectAutocompleteItem(completion.uri);
```
bgrins: This won't work - selectAutocompleteItem isn't a property on this component so it should probably be passed in as a prop

app/ui/browser/views/navbar/location-completion-row.jsx
```diff
@@ -0,0 +1,75 @@
...
+        }}
+        className={focusedResultIndex === index ? FOCUSED_RESULT_STYLE : null}>
```
bgrins: Instead of passing in focusedResultIndex and comparing with index, seems like we could pass in `isSelected` or similar as a prop to make the LocationCompletionRow know a little less about it's surroundings.  What do you think?

bgrins:
I've left a few notes, clearing review flag (not sure if that's the correct process here)

app/ui/browser/views/navbar/location-completion-row.jsx
linclark: Ah, good catch. We should probably add a test for this (and other clicky things). I'm thinking using Enzyme's `mount` would work, but I haven't worked heavily on our tests. I'll check with @jsantell about this.

app/ui/browser/views/navbar/location-completion-row.jsx
linclark: Good idea. I made this change.

linclark:
Hey hey, look at that, it even works now. Marking for review again.

bgrins:
LGTM, thanks!  Please squash when landing

Issue #531, opened 2016-05-24
Use React for the tofino://stars page, replacing the current prototype


victorporof:

victorporof:
Part of https://github.com/mozilla/tofino/issues/518

bgrins:
FYI - there is already a tofino://stars page.  Not sure if you want to use this issue to improve it or if you'd rather close this out

victorporof:
Should use React in there. Rewording this issue title.

ncalexan:
There's an interesting thing happening here, which I don't think we've really explored.  The `tofino://stars` page is _content_, rendered in a `<webview>` with an injected API.  To produce that with React is possible, and possibly beneficial -- but there's no particular tie to our existing React application.  I'd really like to try hosting these content bits (including `tofino://newtab`!) remotely, changing the React calculus yet further.

victorporof:
[Posting our conversation here]

Definitely, tying it to our current application in any way is a clear no-no. We want this to be completely separate from the browser, and we might want to explore the possibility of serving these pages as regular privileged web content in the future ("privileged" in the sense of secure, not like how we defined "privileged" in Firefox where pages could interact with the browser). One immediate clear benefit would be ease of A/B testing for example, but this would also allow any client to consume 'user agent data' through regular https, allowing much more diverse interaction possibly with/through third parties.

Issue #493, opened 2016-05-21
Use React for the tofino://history page, replacing the current prototype


victorporof:
We should also make it follow the blueprint theme.

victorporof:
Part of https://github.com/mozilla/tofino/issues/518

PR #653, opened 2016-06-08
Rewrite content pages to use React

victorporof:
Fixes https://github.com/mozilla/tofino/issues/531 and https://github.com/mozilla/tofino/issues/493


victorporof:
@bgrins

app/ui/content/index.jsx
```diff
@@ -10,14 +10,20 @@
...
-}
+import App from './views/app';
+import { createContentStore } from './store';
 
-if (~document.location.pathname.indexOf('stars')) {
-  fetchRecentBookmarks();
-}
+const store = createContentStore();
+
+const chrome = (
```
bgrins: 'chrome' name seems overloaded here, probably shouldn't show up at all in this app

victorporof: Good point

app/ui/content/views/stars/star-item.jsx
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import React from 'react';
+
+import Style from '../../../shared/style';
+
+import * as ContentPropTypes from '../../model/content-prop-types';
+import ListItem from '../../../shared/widgets/list-item';
+
+const STAR_ITEM_STYLE = Style.registerStyle({
```
bgrins: This is duplicated between stars and history view.  I'd suggest move it to the shared ListItem component in LIST_ITEM_STYLE, until we have a third consumer that doesn't need this style at which point we could break it out.

bgrins:
Looks good, thanks

Issue #319, opened 2016-05-02
Add logging service for both main and renderer processes


jsantell:
Like winston or bunyan

PR #654, opened 2016-06-08
Add bunyan as a logging framework and use it in all non-content processes.

Mossop:
Bunyan was mainly chosen in preference to winston because winston doesn't work in a webpacked environment.

The idea is that the first thing a process should do (before it has chance to log anything) is to configure the logging for the process and after that most code can jsut use the default logger. This gives a slightly weird indirection between the default logger and the actual real logger than isn't created until after startup.

The logger is designed to be mostly swappable for console.

UA and content services get logging to file which helps debugging issues on windows.


Mossop:
This fixes #319.

.eslintrc
```diff
@@ -48,6 +48,7 @@
     "arrow-body-style": ["off"],
     "func-names": ["off"],
     "space-before-function-paren": ["error", "never"],
+    "no-console": ["error"],
```
victorporof: happy happy joy joy

app/ui/browser/index.jsx
```diff
@@ -34,6 +34,9 @@ import * as actions from './actions/main-actions';
 import * as profileDiffs from '../../shared/profile-diffs';
 import BUILD_CONFIG from '../../../build-config';
 import UserAgentClient from '../../shared/user-agent-client';
+import { configureLogging } from '../../shared/logging';
+
+configureLogging('renderer');
```
victorporof: I think this should go before every other import, yes?

build/main.js
```diff
@@ -20,16 +20,19 @@ const thenify = require('thenify-all').thenify;
 const semver = require('semver');
 const checkDependencies = require('check-dependencies');
 const VALID_NODE_VERSION_RANGE = require('../package.json').engines.node;
+const { configureLogging, logger } = require('../app/shared/logging');
+
+configureLogging('build');
```
victorporof: Ditto here?

app/main/browser.js
```diff
@@ -36,6 +36,9 @@ import * as instrument from '../services/instrument';
 import * as spawn from './spawn';
 import * as BW from './browser-window';
 import UserAgentClient from '../shared/user-agent-client';
+import { configureLogging, logger } from '../shared/logging';
+
+configureLogging('main');
```
victorporof: Init this before everything else? Not sure, just asking.

app/shared/logging.js
```diff
@@ -0,0 +1,76 @@
...
+// The entry point to the process should configure this logging service
+let rootLogger = null;
+export function configureLogging(processType) {
+  const streams = [{
+    stream: consolePipe,
+    level: 'debug',
+  }];
+
+  if (processType.endsWith('-service')) {
+    const filestream = fs.createWriteStream(`${processType}.log`);
```
jsantell: Where does this get dumped? Directory root?

jsantell:
Maybe we should turn this off (or push to browser console or something) when in production/packaged build?

app/shared/logging.js
Mossop: Right now CWD. Suggestions for alternatives welcome

Mossop:
I'm reworking this a little.

Mossop:
Ok, @victorporof or @jsantell whoever wants to look. This adds a logging file per process to set up the default logger. It's a tiny bit of boilerplate but it means you don't have to be careful to init before anything else could potentially do logging. Unfortunately that makes logging in shared code harder, I'm passing the logger through here but I'd love to hear an alternative.

The logging to files only happens in development builds. Still to current working directory, do we have a profile directory yet?

app/main/spawn.js
```diff
@@ -51,7 +52,7 @@ export function startUserAgentService(userAgentClient, options = {}) {
   // Connecting to a client immediately is sometimes not necessary,
   // e.g. when starting this service standalone.
   if (userAgentClient) {
-    userAgentClient.connect({ port, host, version });
+    userAgentClient.connect({ port, host, version, logger });
```
victorporof: Why is the `logger` here passed as a param, instead of importing it where required?

app/services/content-service/index.js
```diff
@@ -12,14 +12,15 @@
 
 import * as contentService from './server';
 import * as endpoints from '../../shared/constants/endpoints';
+import { logger } from './logging';
 
-process.on('uncaughtException', console.error);
-process.on('unhandledRejection', console.error);
+process.on('uncaughtException', (...args) => logger.error(...args));
+process.on('unhandledRejection', (...args) => logger.error(...args));
```
victorporof: Why these changes? Do we have to bind `logger` to `error` and other such functions?

app/services/content-service/logging.js
```diff
@@ -0,0 +1,29 @@
+/*
```
victorporof: It's a bit strange having a different "logging" file in each part of the app. Looks like most of this code can be reused. Why not do that?

app/services/content-service/server.js
```diff
@@ -34,7 +35,8 @@ function configure(app, router) {
 export async function start() {
   const { setup } = await makeServer(
     endpoints.CONTENT_SERVER_VERSION,
-    endpoints.CONTENT_SERVER_PORT);
+    endpoints.CONTENT_SERVER_PORT,
+    logger);
```
victorporof: Ditto here: why pass `logger` in when it can be imported?

app/services/user-agent-service/index.js
```diff
@@ -13,9 +13,10 @@
 import * as userAgentService from './server';
 import { ProfileStorage } from './storage';
 import meta from './meta.json';
+import { logger } from './logging';
 
-process.on('uncaughtException', console.error);
-process.on('unhandledRejection', console.error);
+process.on('uncaughtException', (...args) => logger.error(...args));
+process.on('unhandledRejection', (...args) => logger.error(...args));
```
victorporof: Ditto here for the change. Why not bind in the logging module itself, if necessary?

app/services/user-agent-service/logging.js
```diff
@@ -0,0 +1,29 @@
+/*
```
victorporof: Code reuse maybe? :)

app/shared/logging.js
```diff
@@ -0,0 +1,76 @@
...
+    stream: serializer,
+    level,
+  };
+}
+
+/**
+ * Sends log messages to the `console` object
+ */
+export function pipeToConsole() {
+  /* eslint no-console: off */
```
victorporof: Either make this file-level explicitly, or turn on "eslint no-console" at the end of this function.

app/ui/content/logging.js
```diff
@@ -0,0 +1,22 @@
...
+ specific language governing permissions and limitations under the License.
+ */
+
+import bunyan from 'bunyan';
+import { pipeToConsole } from '../../shared/logging';
+
+export const logger = bunyan.createLogger({
+  name: 'content',
+  streams: [
+    // TODO pipe this to the main process stdout
```
victorporof: Does this happen without the PR? If so, it might be a bit sad to have it gone, but not that big of a deal i think, since we'll see log calls in devtools.

victorporof:
@Mossop how about relying on webpack to include the boilerplate, or at least import the right thing for us in every file, instead of having to import it ourselves?

app/main/spawn.js
Mossop: user-agent-client is in shared code, it doesn't know which logger to use.

app/services/content-service/index.js
Mossop: I should double check this, it might have leaked through from another change. If so I guess I can manually bind those.

app/services/content-service/logging.js
Mossop: Of the six different process types we have roughly 3 or 4 different logging configurations. Sharing some of the code is possible but most of it already is. I don't see this changing much except in ways that makes the processes differ more.

bunyan also doesn't seem to allow configuring a logger after creation.

app/services/content-service/server.js
Mossop: Again, makeServer is in shared code.

app/ui/content/logging.js
Mossop: I have no idea, it doesn't on Windows I think but stdout is almost completely busted on windows, partly why I'm doing this to make it actually possible to fix.

Mossop:
@victorporof I'd love to use webpack if I could figure out how. Unfortunately some of the services use babel and not webpack right now and even then I couldn't figure out how to make webpack do what I wanted. Let's talk in London!

Mossop:
@victorporof I failed to notice the r+ and ended up reworking this so you always import the same module regardless of process (build scripts aside) and a webpack define chooses what streams to use. This means we're now webpacking everything that ships in the app!

https://cdn.meme.am/instances/500x/68853970.jpg

app/main/spawn.js
```diff
@@ -26,7 +26,29 @@ const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.
 // @TODO Will `node` be accessible via path on users machines that don't
 // have node installed as an engineer?
 
-const children = [];
+const children = new Set();
+function spawnProcess(name, command, args, options) {
```
victorporof: Should maybe make this shared, possibly augmenting the existing spawn function we have in build/utils?

app/main/spawn.js
```diff
@@ -26,7 +26,29 @@ const CONTENT_SERVICE_PATH = path.join(SERVICES_PATH, 'content-service', 'index.
 // @TODO Will `node` be accessible via path on users machines that don't
 // have node installed as an engineer?
 
-const children = [];
+const children = new Set();
```
victorporof: Uber nit: newline.

app/main/spawn.js
```diff
@@ -18,7 +18,7 @@ import BUILD_CONFIG from '../../build-config';
 const ROOT = path.join(__dirname, '..', '..');
 const DB_PATH = ROOT;
 const SERVICES_PATH = path.join(ROOT, 'lib', 'services');
-const UA_SERVICE_BIN = path.join(SERVICES_PATH, 'user-agent-service', 'bin', 'user-agent-service');
+const UA_SERVICE_BIN = path.join(SERVICES_PATH, 'user-agent-service', 'user-agent-service');
```
victorporof: Why this change?

build/task-build-shared.js
```diff
@@ -1,14 +0,0 @@
...
-import { SRC_DIR } from './const';
-import { babelBuild, babelFile, watchSource } from './utils';
-
-const source = path.join(SRC_DIR, 'shared');
-
-export default async function() {
-  console.log('Building shared...');
-  await babelBuild(source);
-  return watchSource(source, babelFile);
-}
```
victorporof: We'll see who lands our PR first. Losers rebase.

build/webpack.config.content-service.default.js
```diff
@@ -0,0 +1,29 @@
...
+  output: {
+    path: dest,
+    filename: 'index.js',
+    sourceMapFilename: 'index.map',
+  },
+  target: 'node',
+  plugins: [
+    ...defaultConfig.plugins,
+    new webpack.DefinePlugin({
+      LIBDIR: 'require("path").join(__dirname, "..", "..")',
```
victorporof: Instead of __dirname, can we use a predefined const from const.js?

app/main/spawn.js
Mossop: That's the target file for webpack, I could still keep the bin directory around but it seems unnecessary once everything else is gone.

Issue #647, opened 2016-06-08
CI is reporting success when tests fail


Mossop:
See the results at https://travis-ci.org/mozilla/tofino/jobs/135662381 for example. Even though a test timed out travis says everything worked. Appveyor is doing the same so this isn't a travis issue.

victorporof:
I wonder if https://github.com/mozilla/tofino/pull/649 helps at all?

victorporof:
@Mossop Found another problem: https://github.com/mozilla/tofino/pull/636#discussion_r66229583
Removing packages from package.json doesn't make the tests fail, even though packages continue to be used.

Mossop:
@victorporof I'm pretty sure the problem is that `npm test` isn't returning a non-zero exit code in case of error, at least it seems that way when testing locally. You worked on the code that runs tests most recently, could you take a look here?

victorporof:
@Mossop I'll have a look, but I'm not sure if any change I made could have made mocha failures not actually exit with a failure.

Regarding https://github.com/mozilla/tofino/issues/647#issuecomment-224548472 I believe it's because the node_modules folder isn't clobbered? A package was removed from package.json, but was still used somewhere else, and no test failed.

victorporof:
Here's another example: https://github.com/mozilla/tofino/pull/652#event-686758380
CI reports everything as passing, but we were getting 403s https://ci.appveyor.com/project/Mozilla/tofino-u1hv8/build/1.0.538-svnbmaus/job/n4ycvuh03qehdbnf

victorporof:
So https://github.com/mozilla/tofino/pull/660 should fix the originally reported problem (https://github.com/mozilla/tofino/issues/647#issue-159068329), we still have the issue of 
* node_modules folder isn't clobbered, resulting in no failures when modules are removed from package.json, but still being used in app code or tests.
* `npm install` failures (e.g. by getting 403s) aren't reported as CI failures.

Mossop:
The node_modules thing is intentional caching to save time. We can npm
prune at the start of things though.

victorporof:
@Mossop: so how do you feel about pruning? Expectig to take that soon?

Mossop:
Sorry, was preoccupied with other stuff. Happy to take pruning, I can take it tomorrow if you don't want to write it.

Mossop:
Hopefully this is fixed by #695.

PR #660, opened 2016-06-09
Don't wait for stdout/stderr before exiting

victorporof:
There is a race condition between when the process exits and when the stdout and stderr are drained. As a rule, process.exit() will always exit the process immediately, regardless of whether or not there's still something to be done, even like a console.log; furthermore, logging operations themselves aren't atomic, so performing multiple error logging operations consecutively might result in garbled output if they're not drained in succession. This patch only circumvents the issue, because since we drain the output in build/util.js/webpackBuild anyway, and that's the only place we expect rich long output, we can probably safely refrain from draining stdout and stderr at the top level as well.

EDIT:
So what ended up happening was that we were exiting normally, because a forced process.exit(1) wasn't being called since we were waiting for the output to drain.

Should fix *some* of the issues in https://github.com/mozilla/tofino/issues/647

### DIFF
--- a/build/main.js
+++ b/build/main.js
@@ -16,7 +16,6 @@
 require('babel-polyfill');
 require('babel-register')();
 
-const thenify = require('thenify-all').thenify;
 const semver = require('semver');
 const checkDependencies = require('check-dependencies');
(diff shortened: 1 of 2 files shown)


Mossop:
LGTM

PR #664, opened 2016-06-09
Correctly use nowrap property on status hover div, and set max width …

jsantell:
…so we don't have empty space. r=vporof

Not sure how this one slipped by.


victorporof:
Please use "review?" labels in the future.

PR #677, opened 2016-06-10
Add datascript support and tidy up storage implementations.

rnewman:


package.json
```diff
@@ -28,6 +28,7 @@
     "backoff": "2.5.0",
     "body-parser": "1.15.1",
     "datascript": "0.15.0",
+    "datascript-mori": "0.16.0",
```
ncalexan: I think you only need one of these, since `datascript-mori` provides both.  Drop the former.

app/services/user-agent-service/datomstore.js
```diff
@@ -0,0 +1,239 @@
...
+      'page/title': {
+        ':db/ident': ':page/title',
+        ':db/cardinality': ':db.cardinality/one',      // We supersede as we see new titles.
+        ':db/doc': 'A page\'s title.',
+        ':db.install/_attribute': ':db.part/db',
+      },
+
+      'event/visit': {
+        ':db/valueType': ':db.type/ref',
+        ':db/cardinality': 'db.cardinality/many',
```
ncalexan: This doesn't actual work yet, so don't land it.  Specifically, you have `'db.cardinality/many'` when you need `':db.cardinality/many'` (observe colon).

app/services/user-agent-service/datomstore.js
```diff
@@ -0,0 +1,239 @@
...
+
+  // Chains through place ID.
+  starPage(url, session, action, now = microtime.now()) {
+    const result = datascript.core.transact_BANG_(this.conn, vector(
+      vector(DB_ADD, -1, 'page/url', url),
+    ));
+
+    // TODO
+
+    return datascript.core.resolve_tempid(result.tempids, -1);
```
ncalexan: I'm not sure this `result.tempids` does what you think it does.  If you used `datascript.js`, maybe -- but I think this is a mori `map`, not a JS object.

ncalexan:
I am generally okay with this split, although I think the DS code is not quite functional yet.  (I'm working on some fixes, and you can too as you write tests.)

PR #678, opened 2016-06-10
Import from Places into DataScript.

ncalexan:


app/services/places.js
```diff
@@ -0,0 +1,95 @@
...
+        title: row.title,
+        visit_count: row.visit_count,
+        last_visit_date: row.last_visit_date,
+        guid: row.guid,
+        visits: [],
+      });
+    });
+
+    const visitQuery = `SELECT id, from_visit, place_id, visit_date, visit_type, session
+    FROM moz_historyvisits`;
```
rnewman: `ORDER BY place_id, visit_date`? Then you can make the place lookup cheaper, and indeed just assign straight to `place.visits` by partitioning on `place_id` and not have to sort on `visit_date`.

app/services/places.js
```diff
@@ -0,0 +1,95 @@
...
+      let txn = vector(vector(DB_ADD, -1, 'page/url', place.url));
+      place.visits.forEach((visit, i) => {
+        const vid = -2 - i;
+        txn = conj(txn,
+                   vector(DB_ADD, -1, 'event/visit', vid),
+                   vector(DB_ADD, vid, 'visit/instant', visit));
+      });
+
+      const result = datascript.core.transact_BANG_(conn, txn);
+      placeEntities.push(datascript.core.resolve_tempid(result.tempids, -1));
```
rnewman: Do we need `placeEntities`?

test/unit/user-agent-service/test-datomstorage.js
```diff
@@ -0,0 +1,68 @@
...
+
+        const db_after = djs.db(datomStorage.conn);
+        // To inspect the DS store:
+        console.log('count', count(db_after));
+
+        // To see some datoms:
+        // console.log('', toJs(vector(mori.take(10, d.datoms(djs.db(datomStorage.conn),
+        // keyword('eavt'))))));
+
+        const query = parse(`
```
rnewman: I'll move back to fleshing this out this afternoon.

app/services/places.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    // TODO: sort places; prune places with no visits.
+  }
+
+  writePlacesTo(conn) {
+    const placeEntities = [];
+
+    this.places.forEach((place) => {
+      let txn = vector(vector(DB_ADD, -1, 'page/url', place.url));
```
rnewman: Also `page/title`, `place/guid` while we're here — might as well get us most of the way there.

test/unit/user-agent-service/test-datomstorage.js
```diff
@@ -0,0 +1,68 @@
...
+describe('PlacesImporter', function() {
+  this.timeout(30000); // Importing can be slow.
+
+  it('Should import places from disk.', function(done) {
+    (async function () {
+      try {
+        const datomStorage = await ProfileDatomStorage.open('.');
+        const conn = datomStorage.conn;
+
+        const pathToPlaces = '/Users/nalexander/Mozilla/tofino/places.sqlite';
```
rnewman: Can we ship a testing DB instead?

app/services/places.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  static async importFromPlaces(pathToPlaces, datomStorage, limit = undefined) {
+    const placesDB = await DB.open(pathToPlaces);
+    try {
+      const importer = new PlacesImporter(placesDB);
+      await importer.buildPlaceMap(limit);
+      return importer.writePlacesTo(datomStorage.conn);
+    } finally {
+      // TODO: close placesDB.
```
rnewman: Should be pretty easy to add a `close` method to `DB`.

app/services/places.js
```diff
@@ -0,0 +1,95 @@
...
+
+      const result = datascript.core.transact_BANG_(conn, txn);
+      placeEntities.push(datascript.core.resolve_tempid(result.tempids, -1));
+    });
+
+    return placeEntities;
+  }
+
+  static async importFromPlaces(pathToPlaces, datomStorage, limit = undefined) {
+    const placesDB = await DB.open(pathToPlaces);
```
rnewman: Open in read-only mode.

app/services/places-import.js
```diff
@@ -0,0 +1,93 @@
...
+  writePlacesTo(conn) {
+    this.places.forEach((place) => {
+      let assertions = vector(
+        vector(DB_ADD, -1, 'page/url', place.url),
+        vector(DB_ADD, -1, 'page/guid', place.guid));
+
+      if (place.title) {
+        assertions = conj(assertions,
+                   vector(DB_ADD, -1, 'event/title', -2),
+                   vector(DB_ADD, -2, 'event/instant', place.last_visit_date),
```
rnewman: `visit/instant`

rnewman: Err, wait. This isn't the current model. Pages have titles, they're cardinality 1, and there's no timestamp for the event. We use transactionality to manage superseding.

app/services/places-import.js
```diff
@@ -0,0 +1,93 @@
...
+                   vector(DB_ADD, -2, 'title/text', place.title));
+      }
+
+      place.visits.forEach((visit, i) => {
+        const vid = -3 - i;
+        assertions = conj(assertions,
+                   vector(DB_ADD, -1, 'event/visit', vid),
+                   vector(DB_ADD, vid, 'event/instant', visit));
```
rnewman: `event/visit`, `visit/instant`.

app/services/places-import.js
```diff
@@ -0,0 +1,93 @@
...
+          visit_count: row.visit_count,
+          last_visit_date: row.last_visit_date,
+          guid: row.guid,
+          visits: [],
+        };
+        this.places.set(row.id, place);
+      }
+
+      // For now, build entire places tree in memory.  TODO: stream to the DS store.
+      place.visits.push(row.visit_date);
```
rnewman: We'll want to capture visit types, too. File an issue?

app/services/places-import.js
```diff
@@ -0,0 +1,93 @@
...
+    this.places = new Map();
+  }
+
+  async buildPlaceMap(limit = undefined) {
+    const placeQuery = `SELECT p.id, p.url, p.title, p.visit_count, p.last_visit_date, p.guid,
+    hv.visit_date
+    FROM moz_places AS p LEFT JOIN moz_historyvisits AS hv
+    WHERE p.hidden = 0 AND p.id = hv.place_id
+    ORDER BY p.id, hv.visit_date
+    ${limit ? ' LIMIT ${limit}' : ''}
```
rnewman: This'll cause SQL injection checkers to scream. Might as well use a bound var.

test/unit/user-agent-service/test-places-import.js
```diff
@@ -0,0 +1,67 @@
...
+
+        expect(mori.count(dbAfter)).toBe(2409);
+
+        let query = mori.parse(`
+          [:find [?t ...]
+           :in $ ?url
+           :where
+           [?pe "page/url" ?url]
+           [?pe "event/visit" ?ve]
+           [?ve "event/instant" ?t]
```
rnewman: This'll need changing to match.

rnewman:
Consider adding a post-import test that uses the real API to fetch visited pages.

app/services/places-import.js
ncalexan: I think this is short-sighted, for the record.  Will we grow title events later, and then perhaps migrate away from the materialized value?

rnewman: This is the idiomatic way to do this, at least for page titles as we experience them now: a page has one title at a time, and it might change over time.

If you retract the most recent title — as you might when clearing recent history — the older title should transparently take over.

Doing it this way allows you to simply query for the page title, rather than having to manually find the most recent one. That's much more efficient, and the underlying data is still there if you need it (modulo DataScript limitations) — you can find all historical titles by querying across transactions.

If we need to manage multiple titles at the same time, then either (a) they'll be a different kind of thing — e.g., a title associated with a downloaded page summary record, or (b) we can alter our schema; e.g., associating a title with every individual visit, so we can show the right title in a history timeline.

Does that make sense?

PR #696, opened 2016-06-15
Add a clobber command. r=mossop

jsantell:


Mossop:
It won't work on windows but might as well take it anyway.

victorporof:
Should maybe also remove the lib and build-config.json files?

victorporof:
I also suggest having this written in python or node so that we may use it on windows. If written in js, can't import modules.

package.json
```diff
@@ -10,6 +10,7 @@
     "revision": "5cd1dc12f984"
   },
   "scripts": {
+    "clobber": "rm -r .electron; rm build-config.json; rm -r node_modules",
```
pdehaan: Nit: use something like [**rimraf**](https://www.npmjs.com/package/rimraf) for cross platform greatness.

victorporof: If you delete node_modules you can't use rimraf again.

victorporof:
Ok, landing this, will add the `rm -rf lib`.

victorporof:
https://github.com/mozilla/tofino/commit/12425675f8474763ff24da538fce06b2a2f74625

PR #697, opened 2016-06-15
Change colors of build invalidations to yellow, as these are not erro…

jsantell:
…rs and are normal processes to be expected. r=vporof


victorporof:
Was just about to do that.

Issue #655, opened 2016-06-08
Run tests both on development and production builds


victorporof:
Right now we’re using a hybrid: a development build but without react-babel-optimize. Not a good idea. Should have them run both on dev and prod.

But that can get annoying locally. Maybe have that only on CI? @Mossop?

Mossop:
I lean towards always running on production since that is the target and we should at least always do that in CI. Likewise #622.

victorporof:
FWIW we discussed this in the meeting and the consensus was to have them both.
The current approach to have them on 'something that is like production but not quite' seems like could be an open door for problems in the future.

Mossop:
SGTM

PR #698, opened 2016-06-15
Reuse existing builds when running tests

victorporof:
This PR does many things.

1. It tries to nuke as many things as possible out of build-config.json
2. It uses shared files for webpack development and production configurations
3. It relies on an environment variable for determining whether or not we're in a test environment
4. It makes `npm test` reuse whatever build was present, instead of enforcing a production build
5. It makes the CI run tests on both production and development builds

Reason why the `development` property is still in build-config.json, is that in tests we're sometimes importing app/ modules directly (and not running the tests against the js bundle created by webpack), a potential `GLOBAL_CONFIG` variable set by webpack would not exist.

@jsantell

### DIFF
--- a/app/shared/constants/endpoints.js
+++ b/app/shared/constants/endpoints.js
@@ -26,3 +26,6 @@ export const CONTENT_SERVER_VERSION = 'v1';
 
 export const CONTENT_SERVER_ORIGIN = `http://${CONTENT_SERVER_ADDR}:${CONTENT_SERVER_PORT}`;
 export const CONTENT_SERVER_HTTP = `${CONTENT_SERVER_ORIGIN}/${CONTENT_SERVER_VERSION}`;
+
+// The Google Analytics tracking identifier to use when instrumenting.
+export const GOOGLE_ANALYTICS_TRACKING_ID = 'UA-76122102-1';
--- a/build/task-test.js
+++ b/build/task-test.js
@@ -16,8 +16,6 @@ const ELECTRON_MOCHA = path.join(__dirname, '..', 'node_modules', '.bin', 'elect
 const PATH_TO_ELECTRON_MOCHA_OPTS = path.join(__dirname, '..', 'test', 'renderer', 'mocha.opts');
 
 export default async function(args) {
-  logger.info(colors.green('Running tests...'));
-
   // If no arguments passed in, we must run all tests; both mocha
   // and electron-mocha.
   if (!args.length) {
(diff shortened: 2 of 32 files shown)


victorporof:
Also fixes https://github.com/mozilla/tofino/issues/655

app/shared/constants/endpoints.js
```diff
@@ -26,3 +26,6 @@ export const CONTENT_SERVER_VERSION = 'v1';
 
 export const CONTENT_SERVER_ORIGIN = `http://${CONTENT_SERVER_ADDR}:${CONTENT_SERVER_PORT}`;
 export const CONTENT_SERVER_HTTP = `${CONTENT_SERVER_ORIGIN}/${CONTENT_SERVER_VERSION}`;
+
+// The Google Analytics tracking identifier to use when instrumenting.
+export const GOOGLE_ANALYTICS_TRACKING_ID = 'UA-76122102-1';
```
jsantell: I don't know if it makes sense for this constant to be in the endpoints file

build/task-test.js
```diff
@@ -38,8 +39,10 @@ export default async function(args) {
     // globbing, but this is more than fine for now.
     const pathToTests = args[0];
     if (pathToTests.indexOf(path.join('test', 'renderer')) !== -1) {
+      logger.info(colors.cyan('Running renderer tests...'));
       await runRendererTests(pathToTests);
     } else {
+      logger.info(colors.cyan('Running unit tests...'));
```
jsantell: This could also be webdriver tests

build/tasks.js
```diff
@@ -85,7 +86,22 @@ export default {
...
+  },
+
+  async testCI(args = []) {
+    // These tests run on the CI server.
+    logger.info('Making a production build...');
+    await this.build();
+    await Lazy.test(args);
+
+    logger.info('Making a development build...');
+    await this.build({ development: true });
```
jsantell: I think testing both dev and production builds is unnecessary, and increases our chances of webdriver intermittents as well, but whatever

build/utils.js
```diff
@@ -46,6 +46,14 @@ export function getRoot() {
   return path.dirname(__dirname);
 }
 
+export function safeGetBuildConfig() {
```
jsantell: Can't `getBuildConfig` just do this?

test/webdriver/test-basic.js
```diff
@@ -4,8 +4,16 @@
 import expect from 'expect';
 import Driver from '../utils/driver';
 import { HOME_PAGE } from '../../app/ui/browser/constants/ui';
+import BUILD_CONFIG from '../../build-config';
 
 describe('application launch', function() {
+  if (BUILD_CONFIG.development) {
```
jsantell: Ah so we only test this once then even if running both dev and prod tests

app/shared/constants/endpoints.js
victorporof: Eh, I don't have opinions on where to put it. Just not build-config.

build/task-test.js
victorporof: Right.

build/tasks.js
victorporof: https://github.com/mozilla/tofino/issues/655#issuecomment-224663610

build/utils.js
victorporof: Other parts of the codebase need to handle the case where getBuildConfig fails IIRC, so I opted for having a new method.

test/webdriver/test-basic.js
victorporof: Yup.

PR #720, opened 2016-06-19
Allow for separate frontends

bgrins:
This is big, so let me explain and solicit feedback before continuing on working on it.

Goal:
Support more than one frontend for browser chrome.

Non goal:
Support more than one set of content / about pages.

I had two initial ideas:

1) Build it as a single react app that allows switching out one of the higher level container components depending on some sort of frontend pref.  That pref would be loaded / chaged in the rendering process.

Pros:
* Could switch browser chrome immediately in a window without messing up webview state
* Easy, limited build system changes, no need to communicate back to main process

Cons:
* Ties any frontends into the single application

2) Change the URL that the main process uses to load and build two or more completely separate apps that are loaded in a different HTML pages.  Pull out some of the core chrome components (location bar, page, etc) into the shared directory and then webpack two separate bundles.

Pros:
* Additional frontends could use whatever tech they wanted
* Better potential for loading frontends from an external URL if we want to do that in the future.

Cons:
* Extra build system work
* Can only really switch browser chrome when a new window is opened (requires a top-level navigation which would lose track of all webviews). I think this is an OK restriction though, since it seems likely we'd be shipping a different frontend based on the build mostly, not so much focused on runtime switching.
* Can be tricky to figure out what components are 'shared chrome' and what is specific to a frontend.  I think long term depending on frontend designs, things like a Location bar is shared, but NavBar isn't.  We might also need to be able to inject styles and possibly extra options into the shared things.

@victorporof @linclark @jsantell any thoughts?  Am I missing another approach?  If there ends up being a consensus around this, there's still more to do:
* Fix up tests
* Refactor duplication in the build system
* Figure out the UI and IPC mechanism for switching between the frontends.  Some kind of preference that applies to all new windows?


bgrins:
This PR is implementing idea number 2 - mostly shuffling files around and build system changes to make two separate bundles.

victorporof:
I am very much in favor of idea 2, with the caveat that we should try to reuse as much code as possible.

app/ui/browser2/index.jsx
```diff
@@ -0,0 +1,99 @@
+/*
```
victorporof: A lot of stuff in index.jsx looks like it could be shared between browser frontends. More meta, communication with the user agent service and handling of messages/diffs etc. should be dried up.

app/ui/browser2/views/browser-content.jsx
```diff
@@ -0,0 +1,45 @@
+/*
```
victorporof: This looks like it should be shared? I'd expect the content area to be a set of Pages, and that is the same in every frontend I can imagine.

app/ui/browser2/views/developerbar.jsx
```diff
@@ -0,0 +1,94 @@
+/*
```
victorporof: This should also definitely be shared.

bgrins: Agreed, I can move it

app/ui/browser2/views/browser-content.jsx
bgrins: Yeah, I guess this should pretty much be used for any kind of traditional browser chrome

app/ui/browser2/index.jsx
bgrins: Yeah probably, although I'd prefer to do that in a follow up since it'd be easy to do later

bgrins:
Just pushed a rebase / update based on the initial comments.  Only thing I didn't address is getting rid of duplication in build system

jsantell:
Also in favor of method 2; I don't think there's any value in being able to 'hot swap' out UIs while saving state. While cool, reminds me of this gif titled, "[How responsive web designers think users use their site](http://i.imgur.com/YkbaV.gif)".

Also, there's not going to be unified components and style and state right off the bat, and we still may end up with unshareable components due to goals (which, while we have some better ideas, far from concrete), and method 2 atleast can be eased in over time

linclark:
Agreed, method 2 seems like it gives us flexibility that we'll potentially need to use different tech across frontends. If the build system work is manageable, then that sounds like the better solution.

victorporof:
Really nice work here @brian. This is going to help us a lot in the near future.

bgrins:
Thanks for feedback all.  Closing this in favor of #734

PR #734, opened 2016-06-22
Add a stubbed out second frontend

bgrins:
This replaces #720.  We discussed landing a much smaller second frontend initially until there was a clearer decision on what we'd want to share.

@victorporof this is in much less risk of bitrotting, so lmk if you want to reduce duplication in build system here before landing, or if we should do it later.  I think one piece of low hanging fruit would be to have each webpack config file export dev and prod configs (so 1 file instead of 3).

### DIFF
--- a/build/config/webpack.browseralt.default.js
+++ b/build/config/webpack.browseralt.default.js
@@ -1,0 +1,55 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+import path from 'path';
+import webpack from 'webpack';
+import CopyWebpackPlugin from 'copy-webpack-plugin';
+import defaultConfig from './webpack.base';
+import * as Const from '../utils/const';
+
+export const SHARED_DIR = path.join(Const.SRC_DIR, 'ui', 'shared');
+export const SRC_DIR = path.join(Const.SRC_DIR, 'ui', 'browser-alt');
+export const DST_DIR = path.join(Const.LIB_DIR, 'ui', 'browser-alt');
+
+export default {
+  ...defaultConfig,
+  entry: path.join(SRC_DIR, 'index.jsx'),
+  output: {
+    ...defaultConfig.output,
+    path: DST_DIR,
+    filename: 'index.js',
+    sourceMapFilename: 'index.map',
+  },
+  plugins: [
+    ...defaultConfig.plugins,
+    new webpack.DefinePlugin({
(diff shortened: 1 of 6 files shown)


build/config/webpack.browseralt.default.js
```diff
@@ -0,0 +1,55 @@
...
+    new CopyWebpackPlugin([{
+      from: path.join(SHARED_DIR, 'assets'),
+      to: path.join(DST_DIR, 'assets'),
+      flatten: true,
+    }]),
+    new CopyWebpackPlugin([{
+      from: path.join(SHARED_DIR, 'fonts'),
+      to: path.join(DST_DIR, 'fonts'),
+      flatten: true,
+    }]),
```
victorporof: None of these copy rules are actually necessary for now.

build/config/webpack.browseralt.prod.js
```diff
@@ -0,0 +1,14 @@
...
+import browserConfig from './webpack.browseralt.default';
+
+export default {
+  ...prodConfig,
+  ...browserConfig,
+  plugins: [
+    ...prodConfig.plugins,
+    ...browserConfig.plugins,
+  ],
+};
```
victorporof: Should have a transformer function that creates these dev and prod configurations, instead of having to spread things ourselves into object literals. Dev and prod look the same for every single configuration we've written so far, so it's probably safe to assume a generic transformer would suffice.

build/task-build-browser.js
```diff
@@ -35,6 +41,20 @@ async function buildBrowser() {
...
+
+  if (!(await shouldRebuild(id, [SRC_DIR, id], [SHARED_DIR, 'ui/shared']))) {
+    logger.info(colors.green(`No changes in ${id}.`));
+    return { close: () => {} };
+  }
+
+  logger.info(colors.cyan(`Building ${id}...`));
+  const { development } = getBuildConfig();
+  return await webpackBuild(development ? browserAltDevConfig : browserAltProdConfig);
+}
```
victorporof: On the other hand, all of this is yucky to have to write for every single frontend.

victorporof:
Can fix anything whenever, I'm ok with followups.

build/task-build-browser.js
bgrins: Yeah, although until we have a clear idea of how many frontends we want to support I'm OK with having some copy/paste in the build stuff (and it's actually surprisingly little copy/paste so kudos on that)

PR #737, opened 2016-06-23
Shared build config

bgrins:
@victorporof what do you think about this?  The idea is there'll be less places to look for config files (1 config file per bundle), which will also make refactoring multiple configs at the same time a bit easier.  Doesn't solve the general problem of getting rid of duplicated code / making it more declarative, but it's not bad for a limited number of bundles.  And until we know exactly the shape of some more frontends I didn't want to make it too generalized.

I'm sure it was split up originally though for a reason, so let me know if you like this approach.  If so I can finish this up with the rest of the bundles.


victorporof:
Good start. Let's land this.

victorporof:
@bgrins are you still working on this?

bgrins:
> @bgrins are you still working on this?

I think it's good to land, will update commit message

PR #746, opened 2016-06-28
First cut at persisting datoms to SQLite.

ncalexan:
This adds the concept of a snapshot (the `datoms` table) and transaction
log (the `transactions` table).  Updating the snapshot is potentially
expensive; writing to the log should not be expensive.  (All for
SQLite.)  Therefore I'm going to separate the two activities and have
the application layer determine when to update the snapshot.

Next steps are to incrementally update the snapshot, to do some stress
testing, and to consider building Datomic API-alikes exposing the
transaction history, etc.


ncalexan:
@rnewman -- first look?

app/services/user-agent-service/datomstorage-sqlite.js
```diff
@@ -0,0 +1,161 @@
...
+
+      // ps is a lazy sequence.  intoArray forces, like `seq`.
+      ps = mori.intoArray(ps);
+      return Promise.all(mori.intoArray(ps));
+    });
+  }
+
+  async getLastTxInSnapshot() {
+    const [{ lastTx }] =
+      await this.db.all('SELECT COALESCE(MAX(tx), -1) AS lastTx FROM datoms');
```
rnewman: I don't believe there's an index that will apply to make this efficient, so this will be a full table walk.

You might consider putting the current max into a separate table, or adding an index, if this degrades open perf.

app/services/user-agent-service/datomstorage-sqlite.js
```diff
@@ -0,0 +1,161 @@
...
+   */
+  async replayTransactions(conn, fromTx = 0, toTx = 1 << 30) {
+    // TODO: larger upper bound?
+    const rows = await this.db.all(`
+      SELECT e, a, v, tx, added FROM transactions
+      WHERE tx BETWEEN ? AND ?
+      ORDER BY tx ASC`, [fromTx, toTx]);
+    mori.each(mori.partitionBy((row) => row.tx, rows), (txRows) => {
+      const assertions = mori.apply(vector, mori.map(
+        (txRow) => vector(txRow.added ? DB_ADD : DB_RETRACT, txRow.e, txRow.a, txRow.v, txRow.tx),
```
rnewman: I have a sneaking suspicion that this won't handle `:db/txInstant` correctly — it'll probably persist and replay the meta-addition itself, but perhaps it'll apply to a transaction that's not this one.

rnewman: That's true for all meta-statements, I think, not just `txInstant`.

victorporof:
@ncalexan @rnewman what's the status of this PR?

rnewman:
I think this can be closed in favor of Datomish, but I'll let @ncalexan clean up this and #751 and #752.

PR #748, opened 2016-06-28
Test switching to the previous VS environment

Mossop:


Mossop:
@jsantell Apparently we were using an unstable build environment, this looks to fix things.

jsantell:
Ah, nice r+

PR #755, opened 2016-06-29
Allow building packages from any branch

Mossop:
Now we don't use branches in the main repository for personal work it makes sense to do CI builds and package uploads in all branches. This will move the location of the current builds.


bgrins:
Looks good to me, thanks.  Let's confirm builds appear in the new folder after landing this (since IIRC it won't run this on the PR builds)

Mossop:
Builds are showing up where expected (OSX is broken on travis right now)

Issue #771, opened 2016-07-03
[Mac] Click on icon in dock when all windows are closed is not opening new window


vsethuooo:
Open browser in OSX.
Close current window (this does not quit the application in mac; icon still exists in dock)
Click on Electron icon to open new window.

Expected: Clicking on icon should open no window when 0 windows are open
Actual: Nothing happens

PR #772, opened 2016-07-03
Fixing activate for Mac OSX. Issue https://github.com/mozilla/tofino/issues/771

vsethuooo:
[771](https://github.com/mozilla/tofino/issues/771)
https://github.com/mozilla/tofino/issues/771

Issue was that, during activation, the user-agent client was trying to connect, although there was an existing connection. Setting a connected flag to true, fixed this issue.


jsantell:
Just tried this patch out -- good find and fix! Thank you!

PR #776, opened 2016-07-05
Block greenkeeper branches from building in CI.

Mossop:
This appears to work on both CI systems. I don't think it will block PRs from building since we used to restrict builds to only master and PRs still built then too.


Mossop:
And that CI is building this says that that is true!

bgrins:
Works for me

PR #782, opened 2016-07-05
Extract sqlite.js from Tofino

rnewman:
This will allow us to use the promisified sqlite interface from Datomish.


package.json
```diff
@@ -39,9 +39,9 @@
     "immutable": "3.7.6",
     "microtime-fast": "1.0.2",
     "morgan": "1.7.0",
+    "promise-sqlite": "1.1.0",
     "request": "2.65.0",
     "source-map-support": "ncalexan/node-source-map-support#fileUrls-plus",
-    "sqlite3": "mossop/node-sqlite3#v3.1.4.1",
```
ncalexan: @Mossop suggests this may cause problems with `electron-rebuild`, but I don't know the details.

package.json
```diff
@@ -39,9 +39,9 @@
     "immutable": "3.7.6",
     "microtime-fast": "1.0.2",
     "morgan": "1.7.0",
+    "promise-sqlite": "1.1.0",
```
ncalexan: Not sure why we'd version this library something so late, but zfg.

app/services/user-agent-service/sqlite.js
```diff
@@ -163,15 +168,17 @@ export class DB {
    * @param mode any mode flags to pass to `sqlite3.Database`.
    *             Defaults to OPEN_CREATE | OPEN_READWRITE.  May be
    *             OPEN_READONLY, but be aware of https://www.sqlite.org/wal.html#readonly.
+   * @param logger an object with 'info' and 'error' methods, each of which takes a string argument.
+   *               Defaults to a placeholder that logs nothing.
    * @returns Promise that resolves to the opened database.
    */
-  static open(filename, mode = OPEN_CREATE | OPEN_READWRITE) {
+  static open(filename, mode = OPEN_CREATE | OPEN_READWRITE, logger = defaultLogger) {
```
ncalexan: Consider making `{mode, logger}` an argument map.

app/services/user-agent-service/sqlite.js
```diff
@@ -46,9 +45,15 @@ function thenifyMethods(source, dest, methods) {
   }
 }
 
+const defaultLogger = {
```
ncalexan: Prefer something descriptive -- `nullLogger`?

ncalexan:
In general, I'm fine with this.

Right now, my thinking is that Clojure(Script)'s `core.async` is interesting and strictly more powerful than promises, and therefore worth investigating for our internal usage.  There are channels with promise semantics `(promise-chan)` and there are promise-to-channel wrappers (https://github.com/jamesmacaulay/cljs-promises).  I'm not fluent with channels quite yet but they me be better than committing to promises immediately.

That discussion impacts this patch to say: I'm not sure we want a `promise-sqlite`; we might want a `channel-sqlite`.

app/services/places-import.js
```diff
@@ -80,7 +82,7 @@ export class PlacesImporter {
 
   static async importFromPlaces(pathToPlaces, datomStorage, limit = undefined) {
     // The WAL is enabled, so opening read-only leaves orphaned .shm/.wal files.
-    const placesDB = await sqlite.DB.open(pathToPlaces);
+    const placesDB = await sqlite.DB.open(pathToPlaces, undefined, logger);
```
bgrins: Would prefer to have an "options" object that takes in `{ mode, logger }` instead of the spread out params with undefined in there

rnewman:
@ncalexan: indeed. Either way that means breaking this stuff out — firstly so that Datomish can start using it as well as Tofino (e.g., for Places import), and secondly so that Datomish can stop using it later if we choose.

ncalexan:
@rnewman ah, I'd forgotten about Tofino's non-UAS SQLite desires.  Right-o.

rnewman:
64e33ebc62bf14c5f57993c9a5f70528eceb717e

Issue #788, opened 2016-07-07
Rename artifacts published to s3 with a non-version-specific name


bgrins:
Something like `tofino-latest-win32-x64.zip` instead of the current `tofino-0.1.0-win32-x64.zip`. This way there will be a stable URL for the latest master builds.

Issue #112, opened 2016-04-13
Distribute application as an installer


havenchyk:
Hey, thanks for your initiative, it's really nice to see new project based on electron.

I guess the next step will be to distribute the app, so, what do you think about [electron-builder](https://github.com/electron-userland/electron-builder)? It's a complete solution that ready to create installers for different platforms (win, osx, even linux). electron-packager is used inside as well, so you don't need to use so low-level tool.

I'll be glad to hear any feedback from core team.

rnewman:
Quick update: we'll certainly need to do this at some point, but not just yet, which explains the lack of discussion on this issue.

Mossop:
Right now we're using electron-packager for builds but I think I'm going to switch us to electron-builder.

PR #809, opened 2016-07-12
Electron builder

Mossop:
I think this is ready for review, do you want to take it Brian or maybe Jordan should do it.

This changes to using electron-builder. We also upload builds to two different places. S3 gets zip builds as before but additionally OSX gets a DMG build and Windows gets an installer (but we probably shouldn't recommend people use it yet, see below).

Also when doing a tagged build we upload dmg+zip (OSX), exe+nupkg+RELEASES (windows) and zip (linux) to github releases. These are installers and the files needed to enable auto updates. The update server detects them but I haven't verified that it works yet.

Before we can really use windows installers and verify updates we need to do more work in the app itself with the autoUpdater module and other things but I think that can happen outside this branch.

We downgrade eslint to a version that meets all peer dependency constraints (required by electron-builder unfortunately).


Mossop:
It should go without saying that we'll squash the heck out of this before landing ;)

bgrins:
I will do a review but it'd be nice if @jsantell could also look over it, since he will have fresher eyes on this

Mossop:
See #810 for the stuff we should do before people can use windows installers without confusion.

Mossop:
This fixes #112 and fixes #788.

build/task-package.js
```diff
@@ -39,43 +40,115 @@ const IGNORE = [
...
+       path.join(Const.PACKAGED_DIST_DIR, 'RELEASES')],
+    ],
+    zips: [
+      [path.join(Const.PACKAGED_DIST_DIR, `win${archIfNotx64}-unpacked`),
+       path.join(Const.PACKAGED_DIST_DIR, `tofino-${PLATFORM}-${ARCH}.zip`)],
+    ],
+  },
+  darwin: {
+    copies: [
+      [path.join(Const.PACKAGED_DIST_DIR, 'mac', `tofino-${getManifest().version}-mac.zip`),
```
jsantell: So it already zips for OSX?

build/task-package.js
```diff
@@ -39,43 +40,115 @@ const IGNORE = [
...
+            {
+              x: 130,
+              y: 150,
+              type: 'file',
+            },
+          ],
+        },
+        win: {
+          icon: path.join(Const.ROOT, 'branding', 'app-icon.ico'),
+          iconURL: 'https://raw.githubusercontent.com/mozilla/tofino/master/branding/app-icon.ico',
```
jsantell: Hum, I guess this is for a windows store or something?

package.json
```diff
@@ -68,13 +73,13 @@
...
     "dirsum": "0.1.1",
+    "electron-builder": "5.11.2",
     "electron-download": "2.1.2",
     "electron-ipc-mock": "0.0.2",
     "electron-mocha": "2.3.0",
-    "electron-packager": "7.2.0",
     "electron-rebuild": "1.1.5",
     "enzyme": "2.4.1",
-    "eslint": "3.0.0",
+    "eslint": "2.10.2",
```
jsantell: Is this eslint downgrade intentional? I thought the weird deps issue was fixed elsewhere?

package.json
```diff
@@ -83,6 +88,7 @@
     "extract-zip": "1.5.0",
     "fetch-mock": "4.6.0",
     "glob": "7.0.5",
+    "isomorphic-fetch": "2.2.1",
```
jsantell: How were we not using this previously? It's in our mocha tests, but not sure if that was being used. Was this an implicit dep?

appveyor.yml
```diff
@@ -1,6 +1,5 @@
 platform:
   - x86
-  - x64
```
jsantell: Are we intentionally only testing on 32bit?

jsantell:
Looks good, just some comments

build/task-package.js
Mossop: Yes, the "default" electron-builder target builds dmg and zip on OSX (the zip is needed for app updates).

build/task-package.js
Mossop: It is used in the windows control panel. I have no idea why a url is needed but there you go.

package.json
Mossop: Nope, this fixes peer dependency issues where modules we depend on peer depend on 2.9 and above.

package.json
Mossop: One of the other modules (I think fetch-mock) depended on it and npm's deduping meant it appeared at the top-level and so was usable by our code by magic.

appveyor.yml
Mossop: The current plan is to only ship 32-bit on Windows because we can only manage updates for one architecture. I guess we could keep testing on 64-bit and not push builds there but I don't think it likely we'll hit cross-architecture issues.

bgrins: I'm fine with only testing on 32 if that's the current deployment target

.travis.yml
```diff
@@ -39,10 +52,11 @@ install:
   - npm --version
   - npm install --progress false --depth 0
   - npm prune
+  - cd node_modules/mv
+  - npm install rimraf@~2.4.0
```
bgrins: This part is gross and likely to fall out of sync if/when the other packages that require it change dependencies. We still may as well land it in the name of progress, but can you think of a test case we could make to submit a bug to npm and/or the libraries that pull in rimraf to get rid of the need for this?  Then once we have a clean `npm ls` we should enforce that as part of the test suite to make sure we don't check in / update any packages that will re-introduce this issue.

Mossop: https://github.com/npm/npm/issues/13373

Issue #688, opened 2016-06-14
Error: Cannot find module "./build" after starting the browser


victorporof:
Started happening I believe after https://github.com/mozilla/tofino/pull/654

Errors show up in the browser console.

@Mossop ping

victorporof:
I'm also getting 
```
/Users/victorporof/Work/Mozilla/tofino/.electron/Electron.app/Contents/Resources/electron.asar/rend…:130 Unable to load preload script: /Users/victorporof/Work/Mozilla/tofino/lib/shared/preload/content.js
```

We're injecting shared/preload/content.js in page.jsx. This also stopped working after #654.

@Mossop try deleting your lib folder and rebuilding.

Mossop:
Ah I see it now. But it's in the developer tools not the browser toolbox.

jsantell:
We talked about this yesterday, but this is from `bunyan` including dtrace, which is an optional dependency that gets installed since it's used on our main process, but in the webpacked UI version, it can't be 'browserified'. There doesn't appear to be any actual breakage from this, however

Mossop:
This might be fixable with some additions to the webpack config: https://github.com/trentm/node-bunyan#webpack

jsantell:
The above fix doesn't seem to work, as the dtrace provider is still installed, so bunyan thinks it exists, and was installed optionally, getting an error like:

`webpack:///./~/bunyan/lib/bunyan.js?:443 Uncaught TypeError: dtrace.createDTraceProvider is not a function`

We could stub out all methods, but that won't trick Bunyan into not calling it at all

PR #817, opened 2016-07-13
Use a default profile directory and allow passing one on the command line.

### LINKED ISSUE #564: Define a profile directory, use it for storage and for tracking port numbers

### LINKED ISSUE #814: Allow the app to run immediately after launching the osx disk image
If you download a packaged OSX build that opens a folder showing both the app and the Applications directory ([example](https://github.com/mozilla/tofino/releases/download/v0.1.0/tofino-darwin-x64.dmg)), double clicking the app results in the process starting up, but no browser windows open.

I believe this is because the folder / dmg it's trying to write to for the profile database / logs isn't writeable.  A couple of options:

1) We could write to a temp directory just in this case.  That's easy, but your profile data would get lost when you drag / drop the app into the Applications directory.
2) We could switch to using a shared appdata folder across all tofinos regardless of if the app directory is writeable. Maybe this is what we want to ultimately get to, but I have a couple of concerns with it now.  First, it could potentially mess up usertesting.com builds if the same user is doing two different tests and the latter is polluted with the profile data from the former.  Second, now we get to do profile migrations across versions.

Mossop:
Fixes #564.

Brian, can you see if this fixes #814?

### DIFF
--- a/app/shared/environment.js
+++ b/app/shared/environment.js
@@ -1,0 +1,59 @@
+/*
+ Copyright 2016 Mozilla
+
+ Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ this file except in compliance with the License. You may obtain a copy of the
+ License at http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software distributed
+ under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+/* global LIBDIR */
+
+import yargs from 'yargs';
+import path from 'path';
+import AppDirectory from 'appdirectory';
+import fs from 'fs-extra';
+import manifest from '../../package.json';
+import BUILD_CONFIG from '../../build-config.json';
+
+export const argParser = yargs.usage('Usage: $0 [options]').option('P', {
+  alias: 'profile',
+  default: undefined,
+  describe: 'The user profile directory.',
+  type: 'string',
+});
+
+let parsedArgs = undefined;
+
+/**
+ * Note that it is important to not call this before all of the necessary
+ * options have been added to the above argument parser. Normally you should
+ * add options in the top level of modules that are imported at startup then
+ * call this some time later.
+ */
+export function parseArgs() {
+  if (parsedArgs !== undefined) {
+    return parsedArgs;
+  }
+
+  parsedArgs = argParser.argv;
+
+  if (parsedArgs.profile === undefined) {
+    if (!BUILD_CONFIG.development) {
+      const directories = new AppDirectory({
+        appName: manifest.name,
+        appAuthor: manifest.author.name,
+      });
+
+      parsedArgs.profile = directories.userData();
+    } else {
+      parsedArgs.profile = path.join(LIBDIR, '..', 'profile');
+    }
+  }
+
+  fs.mkdirsSync(parsedArgs.profile);
+
+  return parsedArgs;
+}
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ dist/**
 node_modules/**
 app/ui/preload/readability.js
 build/task-package.js
+profile/**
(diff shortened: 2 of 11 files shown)


app/shared/storage.js
```diff
@@ -0,0 +1,21 @@
+/*
```
bgrins: We should be careful about using node dependencies in `shared/` as the UI can pull these in which is breaking the assumption that the frontend can be loaded in a web context.  Some of this should be fixed through our build tools (i.e. the frontend should be packaged up as a web site).  But we should also think about the directory structure and how it related to this (i.e. should ui/ be in a different top-level directory, or should main/services/shared all be moved down one directory from here?).  Bigger than this issue, but we should discuss (cc: @jsantell)

app/main/arguments.js
```diff
@@ -0,0 +1,23 @@
...
+CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+*/
+
+import yargs from 'yargs';
+
+export default yargs
+  .usage('Usage: $0 [options]')
+  .option('p', {
+    alias: 'profile',
```
bgrins: How could we make this match current behavior for usertesting packages?  i.e. in that case we want a unique profile, preferably matching the current ROOT behavior, and it won't be run via command line but by double clicking a binary

bgrins:
> Brian, can you see if this fixes #814?

Yes, that fixes it!  I think we should make `npm run dev` match current behavior (i.e. profile based on working dir), and also make usertesting builds use current behavior.

Do we also need to clean up these directories when the app is uninstalled / deleted?

app/shared/storage.js
jsantell: The logging we use on the UI side of things already breaks this assumption (hence #688), but yeah maybe there's a directory change we could do for 'main process shared' and 'ui shared' and 'both'. Maybe app/main and app/renderer for the main distinction. Also, should make sure we're not putting UI deps in 'dependencies' unless the main process is using which just bloats space (as UI all gets webpacked). Maybe one of the few application structures out there with multiple package.json is the way to go in the future

Mossop:
I've completely re-worked this to address most of the concerns. The default profile directory is based on the build config. For now development builds use a local folder and production builds use something from the user's home directory but we can easily extend that to be controlled by the build branch or anything else once we've decided how we're making usertesting builds.

bgrins:
Looks like there's some bustage in CI: https://travis-ci.org/mozilla/tofino/jobs/145089461 `Error: Cannot find module '../../build-config.json'`

app/shared/environment.js
```diff
@@ -0,0 +1,49 @@
...
+/* global LIBDIR */
+
+import yargs from 'yargs';
+import path from 'path';
+import AppDirectory from 'appdirectory';
+import fs from 'fs-extra';
+import manifest from '../../package.json';
+import BUILD_CONFIG from '../../build-config.json';
+
+export const argparser = yargs.usage('Usage: $0 [options]')
```
bgrins: I believe usages of this module are currently split across process boundaries, but having them modify the state of an exported variable from a shared module seems like a possibility for a bug in the future (i.e. the next time this module is required it's going to have the options from the previous one included).

I'm not sure the best API to avoid that here, but maybe something like:

    export function getArgParser()
      return yargs.usage(...).option(...);
    }

    export function parseArgs(argparser) {
      const options = argparser.argv;
      ...
    }

Mossop: The goal I was going for was that multiple modules could add their arguments to the parser then get the results so I was intentionally having a singleton argparser for the process. I guess now I think of it there is no real reason that is needed, each module could just have its own argparser if it needs one, yargs doesn't modify process.argv I guess.

app/shared/environment.js
```diff
@@ -0,0 +1,53 @@
...
+  alias: 'profile',
+  default: undefined,
+  describe: 'The user profile directory.',
+  type: 'string',
+});
+
+let parsedArgs = undefined;
+
+export function parseArgs() {
+  if (parsedArgs !== undefined) {
```
bgrins: Let's document the decision and limitations here about only evaluating argv once

Issue #747, opened 2016-06-28
Implement tab overview with page summaries


phlsa:
Since a tab overview with page summaries did very well in user testing and also provides us with a basis for further development, we'd like to test it in a live product.

A (clickable) wireframe can be found here:
https://invis.io/2R7DXRM5E#/158003931_Overview

Here's a rough description of the feature:
- The user gets an overview button on the leftmost side of the tab strip
- Clicking that button reveals a full-window overview of all open tabs. Each tab is represented as a card
- Clicking anywhere on the card brings the user to the tab
- (not critical now, but we'll want that at some point: tab cards should be rearrangeable)

The wireframe shows a very ideal case where a lot of data is being scraped from the websites. To be a bit more pragmatic, here's what we should definitely have in the card:
- Key image
- Key color
- Favicon (or better, touch icon)
- Title (this doesn't need to be the `<title>` tag from the website. For example, Amazon prepends »Amazon.com« to the actual title. In that case, we should strip the »Amazon.com« out)
- Excerpt of body text (for articles in particular)
- Star rating (for anything where star ratings exist). We can be pragmatic here and only implement that for Amazon for now

I will follow up here with a higher fidelity mockup of the screen and the various kinds of cards.

One thing that »different kinds of cards for different kinds of websites« implies is that we have some rough categorization of websites (e.g. being able to tell a product page from an article from a recipe). @rnewman do you happen to know if that's something we can do yet?

rnewman:
Existing Mozilla tools can currently do:

* OG type — e.g., article, website. This is unlikely to be sufficient categorization for you.
* Title, URL, description.
* Image and icon(s).

We can special-case title and star rating extraction for Amazon using rules for Fathom.

The page summary folks are aware of, and are working towards, the full range of things you mention.

Categorization by domain or URL match is a tractible problem.

Categorization by available data — e.g., price + star rating + image = rich product card; price + star rating = slim product card — was the topic of my PhD, and reasonable heuristics exist.

phlsa:
Is there a list of metadata we can easily grab (through FB metadata, Twitter cards etc.)? I'm currently designing summary layouts for a bunch of page types and it would be very helpful to know what is out there.

phlsa:
This is what I have so far. The browser layout should be pretty stable. The layout of the summaries themselves less so (see last comment)

![page](https://cloud.githubusercontent.com/assets/514315/16498113/82b31796-3efa-11e6-8bba-f112192c28b2.png)


[...]

PR #821, opened 2016-07-14
Implement page summaries (#747)

victorporof:
Fixes https://github.com/mozilla/tofino/issues/747

Looking to start landing this incrementally. Here's a first couple of commits.
A few more coming soon after I finish cleaning up.
@bgrins


bgrins:
Pulled this down.  A UI note I noticed first is the extra spacing and separators between the active tab and inactive: extra separators and spacing

bgrins:
Is this ultimately planning to land in a branch for a particular experiment or on master?

bgrins:
> Is this ultimately planning to land in a branch for a particular experiment or on master?

Discussed this in the meeting today, and sounds like it's fine to land this on master with the knowledge that we might want to ultimately hide the 'overview' button or other UI elements while testing other features.  IOW we can consider the top bar / nav bar changes present here the new 'default' UI.  @phlsa please correct me if that's wrong.

victorporof:
Landed some of the changes in https://github.com/mozilla/tofino/commit/9770df4939873b10b98822ee4109659506869025

Further commits will follow in this PR.

victorporof:
@bgrins @phlsa This is ready to review.

victorporof:
Main UI:
screen shot 2016-07-14 at 6 03 32 pm

Overview UI:
screen shot 2016-07-19 at 2 30 18 pm

victorporof:
Further steps now revolve around actually creating complex summaries for these cards.

victorporof:
Alternatively, if preferred, I could land this with a subtle checkerboard pattern to be more inline with the rest of the UI.
screen shot 2016-07-19 at 2 52 02 pm

victorporof:
Rebased.

app/ui/browser/reducers/ui-state.js
```diff
@@ -55,6 +55,17 @@ export default function uiState(state = initialState, action) {
     case types.SET_URL_INPUT_AUTOCOMPLETE_INDEX:
       return state.set('focusedResultIndex', action.payload.index);
 
+    case types.TOGGLE_PAGE_SUMMARIES: {
```
bgrins: We've decided in the past to not use TOGGLE_* as an action, instead preferring SET_* so that the action log / frontend is more explicit about what it's requesting.  See something like SET_URL_INPUT_VISIBLE as an example

bgrins:
Should there be a separator here?

screenshot 2016-07-20 10 07 08

bgrins:
extra separator when the last tab is selected tab: screenshot 2016-07-20 11 40 15

jsantell:
This looks overall good to me -- some visual bugs that Brian pointed out, but easier to fix it with it landed, and can begin tackling the harder issue of custom cards

app/ui/browser/reducers/ui-state.js
jsantell: Agreed on this

jsantell:
@bgrins passed this over to me -- after looking through the code, I don't have really any comments other than the naming of the action and maybe the granularity of some of the views, like Thumbnail and FittedImage, especially as we don't have other consumers of those (yet -- maybe you're anticipating some things). Or maybe just a preference. Either way, some of the lower level components could benefit from brief comment descriptions on how/when to use them (if I'm building something else looking at components, I should be able to quickly know when I'd use thumbnail vs fitted image).

Looks great! Once landed we can file follow up bugs and tackle the custom cards.

victorporof:
@bgrins 
Mockups seem to disagree with you: https://cloud.githubusercontent.com/assets/514315/16498113/82b31796-3efa-11e6-8bba-f112192c28b2.png

But since I'm not really sure, I'll land as is and we can get @phlsa's feedback later.

victorporof:
@jsantell FittedImage and Thumbnail both have comments that describe what they're intended for.

Granular components are always a good idea, and incredibly useful once we properly start using `shouldComponentUpdate`. I'd suggest moving forward as we decided in Vancouver: very granular components for everything: better separation of concerns, better performance, better composability, cleaner code.

phlsa:
I'm not entirely sure what you are referring to @victorporof.
If it's about the separators, the idea is that they disappear when they would be next to a selected tab, much like they do in Firefox today.

Issue #804, opened 2016-07-11
Change favicon to magnifying glass when typing in URL bar


jsantell:
Via #740

PR #824, opened 2016-07-14
Change the favicon/info icon to a magnifying glass when typing a quer…

jsantell:
…y into the location bar. r=bgrins


jsantell:
Fixes #804

bgrins:
Works for me

PR #829, opened 2016-07-18
Update README.md with download links

Mossop:
We should maintain these in the readme rather than a pinned item in slack.


bgrins:
Works for me -- @canuckistani can you confirm we should add these download links to the readme?

canuckistani:
👍

PR #830, opened 2016-07-18
Use a less hacky hack for install npm dependencies

Mossop:
Running a second npm install fixes any dependencies missed from the first install attempt.


bgrins:
Why not!

PR #837, opened 2016-07-19
Mark dtrace-provider as an external that shouldn't be compiled into the UI package.

### LINKED ISSUE #688: Error: Cannot find module "./build" after starting the browser

Mossop:
Fixes #688


jsantell:
Nice, didn't know webpack could do this -- tried it out and indeed no more console.error spam r+

PR #844, opened 2016-07-20
Exit immediately if the app has been run by the squirrel installer.

Mossop:
Most of this patch is indentation changes. I can probably get around that if you're worried about blame but it will mean moving a few things around.

### DIFF
--- a/app/main/browser.js
+++ b/app/main/browser.js
@@ -15,19 +15,7 @@ specific language governing permissions and limitations under the License.
 // Must go before any require statements.
 const browserStartTime = Date.now();
 
-import 'source-map-support/register';
 import { logger } from '../shared/logging';
-
-process.on('uncaughtException', (err) => {
-  logger.error(err.stack);
-  process.exit(1);
-});
-
-process.on('unhandledRejection', (reason, p) => {
-  logger.error(`Unhandled Rejection at: Promise ${JSON.stringify(p)}`);
-  logger.error(reason.stack);
-  process.exit(2);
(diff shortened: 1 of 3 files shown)


app/main/browser.js
```diff
@@ -38,100 +38,128 @@ import * as instrument from '../services/instrument';
...
 import UserAgentClient from '../shared/user-agent-client';
+import { argParser, parseArgs } from '../shared/environment';
+
+const SQUIRREL_EVENTS = [
+  'squirrel-install',
+  'squirrel-updated',
+  'squirrel-uninstall',
+  'squirrel-obsolete',
+];
+argParser.boolean(SQUIRREL_EVENTS);
```
bgrins: Please add a comment here to the effect of 'any additional args for the main process need to be added here, before the first call to parseArgs()'

app/main/browser.js
```diff
@@ -38,100 +38,128 @@ import * as instrument from '../services/instrument';
...
 const app = electron.app; // control application life.
-const ipc = electron.ipcMain;
-const globalShortcut = electron.globalShortcut;
-const userAgentClient = new UserAgentClient();
 
-const appStartupTime = Date.now();
-instrument.event('app', 'STARTUP');
+if (handledSquirrelEvent()) {
+  app.quit();
+} else {
```
bgrins: AFAICT everything below here is just whitespace changes, correct me if that's wrong

bgrins:
works for me

PR #846, opened 2016-07-20
Use yargs for handling cli arguments in the build system

victorporof:
Also and trim down the main entry point into smaller modules.


build/main.js
```diff
@@ -16,84 +16,29 @@
 require('babel-polyfill');
 require('babel-register')();
 
+const Tasks = require('./tasks');
+const Node = require('./utils/node');
```
jsantell: Maybe `requirements` or `runtime` would be a better name for this module -- its unclear what it does by the name

PR #854, opened 2016-07-22
Add fathom's meta parser to preload script and store metadata on the …

jsantell:
…page model. r=vporof


app/ui/browser/model/index.js
```diff
@@ -25,6 +25,21 @@ export const Pages = Immutable.Record({
...
+export const PageMeta = Immutable.Record({
+  description: undefined,
+  icon_url: undefined,
+  image_url: undefined,
+  title: undefined,
+  type: undefined,
+  url: undefined,
+  player_url: undefined,
+  player_stream_url: undefined,
+  player_stream_content_type: undefined,
```
victorporof: I think these should be part of some sort of a "super record", not directly part of a PageMeta, which I see more of a base class for such metadata. Everything is fine except the `player_*` rules.

app/ui/preload/metadata-parsing.js
```diff
@@ -0,0 +1,21 @@
...
+ under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ specific language governing permissions and limitations under the License.
+ */
+
+import { metadataRules, getMetadata } from 'page-metadata-parser';
+import tofinoRules from './metadata-rules';
+
+// Use our own rules, as well as those provided from `page-metadata-parser`.
+const RULES = Object.assign({}, metadataRules, tofinoRules);
```
victorporof:
```js
const RULES = {
  ...metadataRules,
  ...tofinoRules,
};
```
is nicer.

build/config/webpack.base.snippets.dev.js
```diff
@@ -7,7 +7,9 @@ export default {
   output: {
     pathinfo: true,
   },
-  devtool: 'eval',
+  // Something in `fathom-web/utils.js` breaks when eval'd,
+  // so comment out 'eval' tool for now???
+  // devtool: 'eval',
```
victorporof: Use a different sourcemap generator instead, like "cheap-source-map".

jsantell: That works. I tried some others and it didn't, can't remember which one it was though

app/ui/browser/model/index.js
jsantell: Not quite sure what you're thinking of how this would work with Immutable and how we manage state. I'll just remove the player_* fields from now here

PR #856, opened 2016-07-22
Use a more fitting backdrop for page summary cards when a background image is supplied

victorporof:
Instead of using random colors:

screen shot 2016-07-22 at 10 38 18 am


jsantell:
Merging this in, working around this area

PR #862, opened 2016-07-22
Fix squirrel event detection in packaged builds.

### LINKED ISSUE #860: yargs can't handle the way process.argv looks in some electron contexts.
yargs assumes that process.argv is [node_path, initial_js, ...rest] so it strips the first two parts of the array and processes the rest (https://github.com/yargs/yargs/blob/master/index.js#L6).

But in electron, particularly when we run as a packaged app process.argv is actually [tofino.exe, ...rest] so we lose the first argument and this breaks events coming from squirrel's installer.

Mossop:
Fixes #860 and cleans up the code for quitting the app when we see squirrel events.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -30,32 +30,49 @@ process.on('unhandledRejection', (reason, p) => {
 import { app } from 'electron';
 import { argParser, parseArgs } from '../shared/environment';
 
-const SQUIRREL_EVENTS = [
+const SQUIRREL_VERSION_EVENTS = [
   'squirrel-install',
   'squirrel-updated',
   'squirrel-uninstall',
   'squirrel-obsolete',
 ];
-argParser.boolean(SQUIRREL_EVENTS);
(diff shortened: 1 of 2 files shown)


jsantell:
Can we move all the squirrel events to its own file in app/main? app/main/index is easy to bloat the top level scope, and this seems pretty contained to handling arguments and quitting the app

Mossop:
> Can we move all the squirrel events to its own file in app/main? app/main/index is easy to bloat the top level scope, and this seems pretty contained to handling arguments and quitting the app

I'm not sure what you're asking for here, index.js is basically 95% handling the squirrel events because it has to happen first and then exit quickly. I don't see the benefit of moving that into another file.

jsantell:
Fair -- I haven't looked at app/main/index in awhile and it's way smaller than last I looked. I think it's fine to land for now, maybe reconsider if app/main/index starts growing again

PR #876, opened 2016-07-25
Add additional metadata parsing explicitly for Amazon. r=vporof

jsantell:


jsantell:
Yes this is ugly, but Amazon explicitly makes it hard to parse, and I imagine this will not work in the future and maybe in different locales, but...

victorporof:
bleh

PR #877, opened 2016-07-26
Tweak the css in the overview

victorporof:
Before:
screen shot 2016-07-26 at 8 10 21 am

After:
screen shot 2016-07-26 at 8 10 13 am

All together:
screen shot 2016-07-26 at 8 16 04 am

### DIFF
--- a/app/ui/browser/constants/ui.js
+++ b/app/ui/browser/constants/ui.js
@@ -30,6 +30,8 @@ export const CARD_HEIGHT = 175; // px
 // Overview (page summaries) favicon/badge dimensions.
 export const CARD_BADGE_WIDTH = 24; // px
 export const CARD_BADGE_HEIGHT = 24; // px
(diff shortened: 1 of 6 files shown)


victorporof:
Additionally, sometimes background images don't fill out the available space horizontally. We do want the background to be contained proportionally horizontally, but we don't care about whether or not it is contained vertically. Because of that, a background size of '100% auto' works better than 'contain'.

Before:
screen shot 2016-07-26 at 9 22 15 am

After:
screen shot 2016-07-26 at 9 27 58 am

(look at the left and right edges)

victorporof:
Lastly, the case where no background images are available is better handled:
Before:
screen shot 2016-07-26 at 5 43 14 pm

screen shot 2016-07-26 at 5 40 13 pm
After:
screen shot 2016-07-26 at 5 40 13 pm

jsantell:
Looks great!

PR #894, opened 2016-07-28
Fix type error in the node version warning and format it more nicely

victorporof:


victorporof:
This also renames 'node.js' to `runtime.js`, as you suggested in a different PR but I forgot to address.

PR #925, opened 2016-08-03
Update to eslint 3.

Mossop:
This finally updates us to eslint 3 with all the related package updates. There are some new rules, most of which are fixed. mocha-eslint uses eslint 2 and so breaks so I have replaced it with something that basically does the same thing only a little faster.

Two rules are still catching problems:

react/no-string-refs objects to us using string refs. @linclark has volunteered to fix these cases.
no-void objects to us using void instead of undefined. @jsantell most of those are in your code, why are we using void? Should we disable the rule or switch to undefined (I don't care which really)?


Mossop:
This is now passing and ready for review

jsantell:
lgtm

Issue #931, opened 2016-08-05
Update mocha to version 3.0+


victorporof:
Current greenkeeper PRs to update mocha fail with a few bad assertions, e.g. https://travis-ci.org/mozilla/tofino/jobs/150041783

It's likely there have been recent breaking API changes in mocha. We should update nevertheless.

victorporof:
This landed I think.

PR #938, opened 2016-08-08
Update to mocha 3.0.2.

Mossop:


Mossop:
Fixes #931

test/utils/async.js
```diff
@@ -1,12 +1,6 @@
 export function autoFailingAsyncTest(runner) {
```
victorporof: Should just completely remove this function now?

Mossop: Maybe? As I recall you added it because people were saying that rejected promises weren't failing tests. If that is still the case then we still need the function. But I could never reproduce that problem so I can't tell if we still need this or not.

victorporof: Let's just remove it.

PR #944, opened 2016-08-10
Remove app icon path in main configuration for build due to change in…

jsantell:
… electron-builder; linux now uses OSX icns anyway. r=mossop

Looks like this [patch changed](https://github.com/electron-userland/electron-builder/commit/ec0bda51732bc2331b257f1965bec009a85bb52b) it in electron-builder, and Linux will now use the `mac` object's icns now.


Mossop:
lgtm

PR #961, opened 2016-08-15
Disable the webdriver test on OSX.

Mossop:
This is basically permafailing so we might as well disable it on OSX.


jsantell:
lgtm

PR #965, opened 2016-08-15
Make the additional processes run with stock node.

Mossop:
There are a number of parts to this:

We split into two package.json files. The package.json in the app directory
contains only dependencies that can't be webpacked either because of native
code or because webpack isn't clever enough. The top-level package.json
dependencies key contains the other dependencies that are needed by the app code
and can be webpacked. The top-level package.json devDependencies contains any
other dependencies needed by build/test code.

This structure means that app/node_modules contains only the dependencies that
actually ship as-is in the final package. Part of webpack now copies all of that
into the lib directory to be used by the app.

The test test-dependencies replaces test-packages and test-devdeps and does a
thorough verification that other than a few exceptions the above rules are met.

Eslints import plugin isn't smart enough for a two package.json structures so
one of its rules had to be replaced.

All of this means we can get rid of module rebuilding for electron since the
native code now runs in pure node.


bgrins:
Something I'm seeing in local testing:

Run `npm run dev`
Kill the process
Run `npm run dev`
See this output:

    main ERROR: Content service exited with exit code 0
    main ERROR: UA Service exited with exit code 0

Everything still works in the second run, but are the content and UA processes not getting killed with the app?

Mossop:
How do you kill the process? When exiting normally the content and UA processes were correctly going away for me.

bgrins:
Looks like they were Electron instances running on 9090 and 9191, so these must be orphaned from an older version (since now they are node instances)

build/config/webpack.base.js
```diff
@@ -30,5 +30,7 @@ export default {
     new webpack.OldWatchingPlugin(),
     new webpack.NoErrorsPlugin(),
   ],
-  externals: [],
+  externals: [
+    'dtrace-provider',
```
bgrins: Why was this added?

build/task-package.js
```diff
@@ -91,17 +72,16 @@ export default async function() {
...
-        app: Const.ROOT,
+        app: Const.LIB_DIR,
         output: Const.PACKAGED_DIST_DIR,
       },
       build: {
         appId: 'tofino',
         'app-category-type': 'web',
-        asar: true,
-        ignore: IGNORE,
+        asar: false,
```
bgrins: Why asar false?

bgrins:
This is a big change, but it makes sense to me.

I'm a little worried about perf on Windows since we can't use asar in stock node (even with the smaller # of node_modules that are shipped in lib).  Have you tested this to make sure that the perf for services is OK?

build/config/webpack.base.js
Mossop: Bunyan uses dtrace-provider to do some special logging on OSX. By marking this as external and then not including it in the app node_modules it fails to resolve when bunyan requires it, which bunyan silently ignores. The dtrace-provider module is broken in recent electron versions due to a complex ABI issue.

build/task-package.js
Mossop: Stock node can't handle asar packages so the services and all the node modules they depend on can't be in asar. The UI itself just webpacks down to a single file anyway so there isn't anything to gain from asar there.

Mossop:
I've been testing this on windows and haven't seen any perf problems.

PR #980, opened 2016-08-18
Ignore ERR_ABORTED/-3 page failures, as this is expected when a page …

jsantell:
…load is overridden by user command (back, reload, etc.). r=vporof

### DIFF
--- a/app/ui/browser/views/page/index.jsx
+++ b/app/ui/browser/views/page/index.jsx
@@ -175,6 +175,12 @@ function addListenersToWebView(webview, pageAccessor, dispatch) {
       return;
     }
 
+    // If a page is aborted (like hitting BACK/RELOAD while a page is loading), we'll get a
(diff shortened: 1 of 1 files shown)


app/ui/browser/views/page/index.jsx
```diff
@@ -175,6 +175,12 @@ function addListenersToWebView(webview, pageAccessor, dispatch) {
       return;
     }
 
+    // If a page is aborted (like hitting BACK/RELOAD while a page is loading), we'll get a
+    // load failure of ERR_ABORTED with code `-3` -- this is intended, so just ignore it.
+    if (Math.abs(errorCode) === 3) {
```
victorporof: Why "Math.abs"?

jsantell: They come in as `-3` -- should maybe save it from here, or add utilities to abstract this difference out, but wondering where else it'll play out, but yeah, in the future, I definitely see some helper functions here.

PR #991, opened 2016-08-23
Use certificate error page when using a weak SSL key. r=vporof

jsantell:


victorporof:
@jsantell there's some failures on CI, planning to look into them, need some help?

Issue #193, opened 2016-04-20
Logging in on twitter.com doesn't work


victorporof:
Looks like we're not handling redirects properly? There's a spew of `command {"payload":{"url":"https://twitter.com/"},"type":"PROFILE_COMMAND_VISITED"}` when trying to log in. @ncalexan maybe you know what's up?

ncalexan:
I do not think we're handling redirects properly; but this spew is because
we can't yet differentiate between iframe loads and main frame loads.
These are just telling the profile, "I visited here", and then after the
redirect, "I visited here".

Mossop:
This is just a bug in electron: https://github.com/electron/electron/issues/3471

Mossop:
Looks like Electron are waiting for a new Chrome refactor to happen which is tracked here: https://bugs.chromium.org/p/chromium/issues/detail?id=368813.

Given the number of blocking bugs I wouldn't expect us to see this fixed until after our three months are up so I think we should fork electron to just take the fix ourselves.

PR #1019, opened 2016-08-26
Update to electron 1.3.4

Mossop:
This forks electron to include a patch that fixes twitter login. There is a possibility that this breaks some uses of preload scripts but I could find nothing broken in my testing.


bgrins:
Works for me in local testing

Mossop:
This fixes #193.

PR #1021, opened 2016-08-26
Make the additional processes run with stock node

Mossop:
There are a number of parts to this:

For running and packaging we copy the system node to the lib directory.

We split into two package.json files. The package.json in the app directory
contains only dependencies that can't be webpacked either because of native
code or because webpack isn't clever enough. The top-level package.json
dependencies key contains the other dependencies that are needed by the app code
and can be webpacked. The top-level package.json devDependencies contains any
other dependencies needed by build/test code.

This structure means that app/node_modules contains only the dependencies that
actually ship as-is in the final package. Part of webpack now copies all of that
into the lib directory to be used by the app.

The test test-dependencies replaces test-packages and test-devdeps and does a
thorough verification that other than a few exceptions the above rules are met.

Eslints import plugin isn't smart enough for a two package.json structures so
one of its rules had to be replaced.

All of this means we can get rid of module rebuilding for electron since the
native code now runs in pure node.


Mossop:
@bgrins Please just look at the last three commits, the first is identical to what you reviewed previously.

I'm not sure the babel changes are necessary but they seem useful anyway.

CI was failing with file copy errors for node and node_modules, it seems that the copy-webpack-plugin isn't great when used with a lot of files (https://github.com/kevlened/copy-webpack-plugin/issues/59) and their suggested fix with graceful-fs didn't help so I just abandoned webpack for those bits. It means no file watching for that but that doesn't seem like a big problem.

Also added some docs on what the two `package.json` files are for.

docs/package manifests.md
```diff
@@ -0,0 +1,19 @@
+Tofino uses a two package manifest setup where node modules required to ship as-is with the
```
bgrins: This looks fine to me - maybe name the file `package-manifests` to be consistent with other files in the dir

bgrins:
Looks fine to me

PR #1033, opened 2016-08-30
Build system changes to streamline having multiple frontends

victorporof:


build/task-build-preload.js
```diff
@@ -0,0 +1,24 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+import colors from 'colors/safe';
+import { default as preloadConfig, SRC_DIR } from './config/webpack.preload';
+import { makeDevConfig as dev, makeProdConfig as prod } from './config/webpack.base';
+
```
bgrins: Consistency nit: extra line can be removed

build/utils/const.js
```diff
@@ -6,14 +6,13 @@ import path from 'path';
...
 export const SRC_DIR = path.join(ROOT, 'app');
 export const LIB_DIR = path.join(ROOT, 'lib');
-export const APP_SHARED_DIR = path.join(SRC_DIR, 'shared');
+export const PACKAGED_DIST_DIR = path.join(ROOT, 'dist');
 
+export const APP_SHARED_DIR = path.join(SRC_DIR, 'shared');
 export const BUILD_CONFIG_PATH = path.join(SRC_DIR, 'build-config.json');
-
-export const PACKAGED_DIST_DIR = path.join(ROOT, 'dist');
+export const BROWSER_FRONTENDS_PATH = path.join(BUILD_SYSTEM_DIR, 'config', 'frontends.json');
```
bgrins: Personally, I'd find it more straightforward if instead of this pointing to a JSON file it was an array like:

    export const BROWSER_FRONTENDS = [
      'browser',
      'browser-alt',
    ];

Just because it's not going to change frequently and I don't see much value in having the extra file / JSON reading

bgrins:
This looks good to me.  See the comment about the list of frontends

build/utils/const.js
victorporof: I'd like having this list in a separate file. That way whenever a new frontend is added, all that is needed is editing that array and creating a new webpack config file. I personally don't see any relevant difference between having a js file exporting an array, or a json file that's being loaded. Feel free to let me know if you feel strongly about this.

PR #1044, opened 2016-09-01
Make some improvements in swiping, more lenient on the Y-axis, use th…

jsantell:
…e OS-level swipe preference, and set the wheel event listener to being passive. r=vporof

Still doesn't feel completely right, the touch begin event doesn't always trigger, it seems. But better.


victorporof:
This can land now, but a rebase is needed.

jsantell:
Adding this change along with general gesture support in browser-modern in #1071

PR #1041, opened 2016-09-01
Preliminary changes allowing sharing relevant code between browser frontends

victorporof:
Depends on https://github.com/mozilla/tofino/pull/1064

PR #1047, opened 2016-09-02
Various improvements for shared components

victorporof:
Depends on https://github.com/mozilla/tofino/pull/1041


victorporof:
Depends on https://github.com/mozilla/tofino/pull/1041

victorporof:
See https://github.com/mozilla/tofino/pull/1041#issuecomment-244404462

app/ui/shared/widgets/overview/summaries/star.jsx
```diff
@@ -0,0 +1,50 @@
...
+const STAR_IMAGES = {
+  full: 'glyph-ratings-full.svg',
+  half: 'glyph-ratings-half.svg',
+  empty: 'glyph-ratings-empty.svg',
+};
+
+class Star extends Component {
+  constructor(props) {
+    super(props);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
```
bgrins: From https://facebook.github.io/react/docs/pure-render-mixin.html:

> Furthermore, shouldComponentUpdate skips updates for the whole component subtree. Make sure all the children components are also "pure".

Is it safe to assume that Btn is pure?  If so, why is it using isEqual in the previous commit?  https://github.com/mozilla/tofino/pull/1047/commits/04261c74ec00a77ea263b3ff8e2fdd0a413170a6#diff-1f66465f034f1cd156f8ff2bc8be49b3R48

app/ui/shared/widgets/overview/cards/base-card.jsx
```diff
@@ -71,53 +72,60 @@ const SUMMARY_STYLE = Style.registerStyle({
...
-
-  if (props.page.location.startsWith(Endpoints.TOFINO_PROTOCOL)) {
-    badgeImage = TOFINO_LOGO;
-  } else {
-    backgroundImage = props.backgroundImage || props.page.meta.image_url;
-    badgeImage = props.badgeImage || props.page.meta.icon_url;
+class BaseCard extends Component {
+  constructor(props) {
+    super(props);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
```
bgrins: From https://facebook.github.io/react/docs/pure-render-mixin.html:

> This only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences.

Is this OK for props.page?

app/ui/shared/widgets/overview/summaries/star.jsx
victorporof: `Btn` is taking regular js objects as props, so shallow compares won't work because that would end up always allowing rerenders. For components that take in immutable.js objects or js primitives as props, PureRenderMixin is adequate.

app/ui/shared/widgets/overview/cards/base-card.jsx
victorporof: Yes, since it's an immutable.js object.

PR #1064, opened 2016-09-09
Fix the build system to avoid having dangling IO streams which prevent the main process from closing

victorporof:
This should fix the test suite stalling in the frontend-rewrite-related PRs.

Please see individual commits.

### DIFF
--- a/build/task-build-browser.js
+++ b/build/task-build-browser.js
@@ -3,10 +3,9 @@
 
 import colors from 'colors/safe';
 import fs from 'fs-promise';
+import path from 'path';
 
-import { BROWSER_FRONTENDS_PATH } from './utils/const';
-import { makeDevConfig as dev, makeProdConfig as prod } from './config/webpack.base';
-import { getBuildConfig } from './utils';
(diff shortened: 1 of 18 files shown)


build/task-build-browser.js
```diff
@@ -3,10 +3,9 @@
 
 import colors from 'colors/safe';
 import fs from 'fs-promise';
+import path from 'path';
 
-import { BROWSER_FRONTENDS_PATH } from './utils/const';
-import { makeDevConfig as dev, makeProdConfig as prod } from './config/webpack.base';
-import { getBuildConfig } from './utils';
+import { BUILD_WEBPACK_CONFIS_PATH, BROWSER_FRONTENDS_PATH } from './utils/const';
```
jsantell: I think `BUILD_WEBPACK_CONFIS_PATH` is a typo

victorporof: CONFIGS yes, not CONFIS

victorporof: Luckily this typo is present everywhere else, so there's no type error.

victorporof:
Fixed typo.

Issue #1068, opened 2016-09-09
Deleting the lib folder should cause everything to be rebuilt on `npm start` or `npm run dev`


victorporof:
This has been broken for quite a while, but it used to work.

PR #1069, opened 2016-09-09
Rebuild if the lib directory is missing

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1068

The problem was caused by the build-modules task, which always created a lib directory *before* we got a chance to test whether or not it was deleted.

Regressed by https://github.com/mozilla/tofino/commit/3d8d6314995d6a16442cc58d385f8509e6a97b6e

### DIFF
--- a/build/tasks.js
+++ b/build/tasks.js
@@ -4,6 +4,7 @@
 import colors from 'colors/safe';
 import { logger } from './logging';
 import { safeGetBuildConfig } from './utils';
+import { buildDirectoryExists, deleteBuildConfigHashes } from './utils/rebuild';
 import argv from './utils/argv';
 
 /**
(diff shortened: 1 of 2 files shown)


build/task-build-modules.js
```diff
@@ -15,6 +16,11 @@ function copy(src) {
 export default async function() {
   logger.info(colors.cyan('Building node environment...'));
 
+  // Check if the `lib` directory exists. If it doesn't, need to rebuild.
+  if (!(await buildDirectoryExists())) {
```
bgrins: Should this be happening instead in Tasks.build?  Seems like the buildConfigHashes stuff is relevant to the other build tasks and not this one, so maybe it should happen here to be more clear: https://github.com/mozilla/tofino/blob/master/build/tasks.js#L41

victorporof: Correct, it should happen before everything else.

bgrins: OK, in that case then please move this block to Tasks.build

victorporof: Will do! R+ otherwise?

bgrins: Yes

PR #1071, opened 2016-09-09
Add swipe gestures to browser-modern, and use passive wheel handler, as

jsantell:
well as OS level preference from Electron 1.3.3+. r=vporof


app/ui/browser-modern/lib/gestures.js
```diff
@@ -0,0 +1,72 @@
+/*
```
victorporof: I think this file should be in ui/shared/util

PR #1079, opened 2016-09-10
Make location autocompletions immutable Record instances

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1076

### DIFF
--- a/app/ui/shared/widgets/list-item.jsx
+++ b/app/ui/shared/widgets/list-item.jsx
@@ -11,8 +11,10 @@ specific language governing permissions and limitations under the License.
 */
 
 import React, { Component, PropTypes } from 'react';
+import shallowEqual from 'fbjs/lib/shallowEqual';
 import isEqual from 'lodash/isEqual';
 import omit from 'lodash/omit';
+import without from 'lodash/without';
 
 import Style from '../style';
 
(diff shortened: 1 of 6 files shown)


app/ui/shared/widgets/list-item.jsx
```diff
@@ -22,7 +24,10 @@ const LIST_ITEM_STYLE = Style.registerStyle({
 
 class ListItem extends Component {
   shouldComponentUpdate(nextProps) {
-    return !isEqual(this.props, nextProps);
+    // Children in this component may contain immutable.js props, which are
```
bgrins: The warnings are one thing, but also this is changing the behavior of shouldComponentUpdate to return false when child components re-render, and I think that's important to capture in this comment.

victorporof: You bet!

test/unit/lint/test-dependencies.js
```diff
@@ -32,8 +32,13 @@ describe('dependencies', () => {
       `app/ui${all}${valid}`,
     ];
 
+    const implicit = [
+      // These aren't specified in package.json but the are installed anyway.
+      'fbjs',
```
bgrins: This is a little odd - I'm not sure about if we should be doing this or instead explicitly relying on fbjs in package.json.  Feel free to land for now to clear out your pending patch list but @linclark what do you think about this?

victorporof: I can also `npm install fbjs --save` to remove this.

app/ui/shared/widgets/list-item.jsx
victorporof: Updated the comment to 
```
    // Since this component uses plain js objects as props, we need to perform
    // deep quality checks on them to make sure we don't unnecessarily rerender.
    // However, there is no need to re-render this parent component when the
    // children change, which would happen when doing naive deep equality.
    // Furthermore, children in this component may contain immutable.js props,
    // which are throwing warnings when accessing some of their properties
    // while doing deep equality checks using lodash's `deepEqual`.
```

Issue #1057, opened 2016-09-07
Use webview's ‘update-target-url’ events instead of adding our own listeners in content


victorporof:
http://electron.atom.io/docs/api/web-view-tag/#event-update-target-url

PR #1082, opened 2016-09-10
Use webview's `update-target-url` events instead of adding our own listeners in content

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1057


Mossop:
👍

Issue #1056, opened 2016-09-07
Use webview's `findInPage` instead of rolling our own implementation


victorporof:
http://electron.atom.io/docs/api/web-view-tag/#event-found-in-page

PR #1083, opened 2016-09-10
Use webview's `findInPage` instead of rolling our own implementation

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1056


Mossop:
👍

Issue #1087, opened 2016-09-10
Autocomplete List Stays Open When Switching Tabs


darkwing:
1.  Open 2-3 tabs
2.  Enter a partial URL so autocomplete list shows
3.  Select a different tab using CONTROL+TAB

The Autocomplete list stays open.  Maybe we just close it, or only show it when the user navigates back to that tab?

victorporof:
I can't reproduce this with the steps described, but these do:

1. Open 2-3 tabs
2. Select first tab, enter a partial URL so autocomplete list shows
3. Select a different tab using CONTROL+TAB
4. Enter a partial URL so autocomplete list shows again
5. Select a first tab using CONTROL+TAB

The autocomplete list is open and shows the new results instead of the old ones, even though the text in the location bar is the old one.

Expected: either the autocomplete list isn't open anymore, or it's open but shows the old results.

PR #1091, opened 2016-09-11
Properly handle the autocompletions list when switching tabs

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1087

Depends on https://github.com/mozilla/tofino/pull/1079

This dependency graph might help: 
screen shot 2016-09-12 at 12 40 55 pm

### DIFF
--- a/app/ui/shared/widgets/autocompleted-search.jsx
+++ b/app/ui/shared/widgets/autocompleted-search.jsx
@@ -48,10 +48,11 @@ class AutocompletedSearch extends Component {
   handleInputKeyDown = e => {
     const { showAutocompletions, selectedIndex } = this.state;
     const { autocompletionResults } = this.props;
+    const hasAutocompletionResults = autocompletionResults && autocompletionResults.size;
 
     switch (e.key) {
       case 'Enter': {
-        if (showAutocompletions && autocompletionResults.size) {
+        if (showAutocompletions && hasAutocompletionResults) {
           const data = this.props.autocompletionResults.get(this.state.selectedIndex);
           this.props.onAutocompletionSelect(data);
           // Don't call any `onKeyDown` event handler on a parent component.
@@ -69,7 +70,7 @@ class AutocompletedSearch extends Component {
         }
         break;
       case 'ArrowUp':
-        if (showAutocompletions && autocompletionResults.size) {
+        if (showAutocompletions && hasAutocompletionResults) {
           if (selectedIndex === 0) {
             this.setState({ selectedIndex: autocompletionResults.size - 1 });
           } else {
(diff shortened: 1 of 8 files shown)


app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -69,9 +70,9 @@ class AutocompletedSearch extends Component {
         }
         break;
       case 'ArrowUp':
-        if (showAutocompletions && autocompletionResults.size) {
+        if (showAutocompletions && hasAutocompletionResults) {
           if (selectedIndex === 0) {
-            this.setState({ selectedIndex: autocompletionResults.size - 1 });
+            this.setState({ selectedIndex: hasAutocompletionResults - 1 });
```
bgrins: This is meant to be an int

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -81,8 +82,8 @@ class AutocompletedSearch extends Component {
         break;
       case 'Tab':
       case 'ArrowDown':
-        if (showAutocompletions && autocompletionResults.size) {
-          if (selectedIndex === autocompletionResults.size - 1) {
+        if (showAutocompletions && hasAutocompletionResults) {
+          if (selectedIndex === hasAutocompletionResults - 1) {
```
bgrins: Ditto

app/ui/shared/widgets/autocompleted-search.jsx
victorporof: Oops!

victorporof: Well, it is an int, but the variable is poorly named.

bgrins:
The fact that we keep it open with the old results is interesting and different from other browsers.  In particular, afaict other browser treat the autocomplete list *not* as per-page but a single global thing that gets closed and cleared when switching tabs.

But I think it leads to some UX things, like when I do:

* Open two tabs
* Type something in URL bar A to show autocomplete list
* Switch tabs (ctrl+tab)
* Type something in URL bar B to show autocomplete list
* Switch back (ctrl+tab)

The autocomplete isn't focused but it looks like it is (i.e. pressing down arrow doesn't switch between results).

Another existing issue that I'll file separately: the list doesn't disappear even when I click elsewhere in the browser (on another tab or even just somewhere in the nav bar).  Seems the only way to get it to go away is by pressing ESC - but once I've clicked elsewhere the URL bar loses focus so ESC doesn't close it.

bgrins:
I'm OK to handle the focus issue in https://github.com/mozilla/tofino/pull/1091#issuecomment-247046366 as a follow up if you want to track that separately

victorporof:
@bgrins the focus issue was there even before this patch. Indeed it would be a great idea to support it to make the autocompletion-list-per-page thing nice from a UX pov.

Issue #1016, opened 2016-08-25
Cannot increase/decrease font size


darkwing:
I've recently given into being an old man and have started using COMMAND(and)+ to increase font sizes on websites.  Would really love to be able to increase and decrease the font size for better readability.

PR #1097, opened 2016-09-11
Add ability to increase/decrease font size

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1016

Depends on https://github.com/mozilla/tofino/pull/1092

This dependency graph might help: 
screen shot 2016-09-12 at 12 40 55 pm


app/ui/browser-modern/actions/page-effects.js
```diff
@@ -111,6 +112,61 @@ export function toggleCurrentPageDevtools() {
...
+
+export function performPageZoomIn(pageId) {
+  return (dispatch, getState) => {
+    let zoomLevel = PagesSelectors.getPageZoomLevel(getState(), pageId);
+    zoomLevel += PageState.ZOOM_LEVEL_10_PERCENT;
+    dispatch(setPageZoomLevel(pageId, zoomLevel));
+  };
+}
+
+export function performCurrentPageZoomIn() {
```
jsantell: Would be less code with a setZoom that took a value or a delta `setZoom(val)`/`setZoom({ delta: -10 })`, but already written, so up to you for maintainability

Issue #1099, opened 2016-09-11
Clicking on anchor links doesn't update the page's location


victorporof:
Neither in the app state nor in the location bar.

E.g. go to http://electron.atom.io/docs/api/web-view-tag/ and click on the anchor link next to "Example", to go to http://electron.atom.io/docs/api/web-view-tag/#example

PR #1100, opened 2016-09-11
Update the location on hash changes or local anchor links

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1099


app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -100,6 +112,8 @@ class Page extends Component {
 
     this.webview.addEventListener('page-title-set', e => {
```
jsantell: Does this always fire on hash loc changes?

victorporof: Nope!

PR #1107, opened 2016-09-12
Reimplement net/cert error pages in browser-modern theme. Fixes #1072…

### LINKED ISSUE #1072: Handle certificate errors in browser-modern

jsantell:
…. r=vporof

The location bar isn't updating on an error, even though Search.jsx is receiving new props, and componentShouldUpdate returning true upon this. Any ideas? Can handle in a follow up.


.eslintrc
```diff
@@ -72,7 +72,7 @@
     "babel/generator-star-spacing": ["error", { before: false, after: true }],
 
     "react/display-name": ["error", { ignoreTranspilerName: true }],
-
+    "react/prefer-stateless-function": ["off"],
```
jsantell: https://medium.com/modus-create-front-end-development/component-rendering-performance-in-react-df859b474adc#.u5375qyp7

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -73,33 +80,75 @@ class Page extends Component {
...
     });
 
+    this.webview.addEventListener('did-finish-loading', () => {
+      // Event fired only when page is successfully loaded
+      this.props.dispatch(PageActions.setPageState(this.props.pageId, {
+        load: PageState.STATES.LOADED,
+      }));
+    });
+
+    this.webview.addEventListener('did-stop-loading', () => {
```
jsantell: Previously, this was marking the page as LOADED, even in event of failure.

app/ui/browser-modern/views/browser/page/cert-error.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+const CERTIFICATE_STYLE = Style.registerStyle({
+  backgroundColor: '#ddd',
+  fontWeight: 'bold',
+  display: 'block',
+  padding: '15px',
+});
+
+class CertErrorPage extends Component {
+  render() {
```
victorporof: Need a shouldComponentUpdate here.

.eslintrc
victorporof: No need for this if you add shouldComponentUpdate everywhere.

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,58 @@
...
+  height: '100%',
+  position: 'absolute',
+  color: 'rgb(66, 78, 90)',
+  backgroundColor: '#FBFBFB',
+  fontSize: '120%',
+  overflowY: 'auto',
+});
+
+class ErrorPage extends Component {
+  render() {
```
victorporof: Add shouldComponentUpdate.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -60,6 +63,10 @@ class Page extends Component {
       // May be emitted multiple times for every single frame.
       this.props.dispatch(PageActions.setPageState(this.props.pageId, {
         load: PageState.STATES.LOADING,
+        code: null,
+        description: null,
+        url: null,
+        certificate: null,
```
victorporof: Why reset all of these to null on a new-window event?

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -73,33 +80,75 @@ class Page extends Component {
       this.props.dispatch(UIEffects.focusURLBar(this.props.pageId, { select: isTofino }));
     });
 
+    this.webview.addEventListener('did-finish-loading', () => {
+      // Event fired only when page is successfully loaded
```
victorporof: Nit: fullstop at the end of sentences.
Can this be emitted multiple times for every single frame?

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -73,33 +80,75 @@ class Page extends Component {
...
+      // May be emitted multiple times for every single frame.
+      // Emitted regardless of whether or not a failure occurs.
+      const url = this.webview.getURL();
+      const title = this.webview.getTitle();
+
+      this.props.dispatch(PageActions.setPageDetails(this.props.pageId, {
+        canGoBack: this.webview.canGoBack(),
+        canGoForward: this.webview.canGoForward(),
+        location: url,
+        title,
```
victorporof: Why set the canGoBack and canGoForward props here?
Should the location be update here? The `did-navigate-in-frame` event is better for in-frame navigations. Anything else that isn't handled?

victorporof: Should the location be update here? The `did-navigate-in-frame` event is better for in-frame navigations. Anything else that isn't handled?

victorporof: Can the title ever be different here? Why update it when we have the `page-title-set` event?

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -143,17 +195,19 @@ Page.displayName = 'Page';
 
 Page.propTypes = {
   dispatch: PropTypes.func.isRequired,
+  onMount: PropTypes.func.isRequired,
   pageId: PropTypes.string.isRequired,
-  pageLocation: PropTypes.string.isRequired,
   pageIndex: PropTypes.number.isRequired,
-  onMount: PropTypes.func.isRequired,
+  pageLocation: PropTypes.string.isRequired,
+  pageState: PropTypes.instanceOf(PageState).isRequired,
```
victorporof: Import from shared props for this.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -130,6 +179,9 @@ class Page extends Component {
     return (
       <div id={`browser-page-${this.props.pageId}`}
         className={`browser-page ${PAGE_STYLE}`}>
+        <ErrorPage
+          hidden={this.props.pageState.load !== PageState.STATES.FAILED}
+          {...this.props.pageState.toJS()} />
```
victorporof: Ugh, please let's not uselessly convert immutable js objects to js objects for no reason. Update the ErrorPage to handle immutable objects instead.

app/ui/browser-modern/views/browser/page/net-error.jsx
```diff
@@ -0,0 +1,54 @@
...
+import Style from '../../../../shared/style';
+
+const CONTAINER_STYLE = Style.registerStyle({
+  width: '700px',
+  margin: '0 auto',
+  display: 'block',
+});
+
+class NetErrorPage extends Component {
+  render() {
```
victorporof: Needs a shouldComponentUpdate.

app/ui/browser-modern/views/browser/page/net-error.jsx
```diff
@@ -0,0 +1,54 @@
...
+CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+*/
+
+import React, { PropTypes, Component } from 'react';
+
+import Style from '../../../../shared/style';
+
+const CONTAINER_STYLE = Style.registerStyle({
+  width: '700px',
```
victorporof: What if the browser is not wide enough for this? This should either flex or use precentage or viewport dimensions.

app/ui/browser-modern/views/browser/page/net-error.jsx
```diff
@@ -0,0 +1,54 @@
...
+  margin: '0 auto',
+  display: 'block',
+});
+
+class NetErrorPage extends Component {
+  render() {
+    return (
+      <div className={CONTAINER_STYLE}>
+        <h1>Could not connect</h1>
+        <p>Tofino can't find the server at <strong>{this.props.url}</strong></p>
```
victorporof: Please don't use tags for styling :)

app/ui/browser-modern/views/browser/page/net-error.jsx
```diff
@@ -0,0 +1,54 @@
...
+        <ul>
+          <li>
+            Check the address for typing errors such as <strong>ww</strong>.example.com
+            instead of <strong>www</strong>.example.com
```
victorporof: Ditto here. Have a class instead of `strong`.

app/ui/browser-modern/views/browser/page/cert-error.jsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+
+CertErrorPage.displayName = 'CertErrorPage';
+
+CertErrorPage.propTypes = {
+  url: PropTypes.string,
+  code: PropTypes.number,
+  description: PropTypes.string,
+  certificate: PropTypes.object,
```
victorporof: Don't use object prop types. Define a shape in shared props, or use immutable.js objects all the way.

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,58 @@
...
+import NetErrorPage from './net-error';
+import CertErrorPage from './cert-error';
+import Style from '../../../../shared/style';
+
+const ERROR_PAGE_STYLE = Style.registerStyle({
+  width: '100%',
+  height: '100%',
+  position: 'absolute',
+  color: 'rgb(66, 78, 90)',
+  backgroundColor: '#FBFBFB',
```
victorporof: Use the css variables in the theme file for colors.

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+ErrorPage.displayName = 'ErrorPage';
+
+ErrorPage.propTypes = {
+  url: PropTypes.string,
+  code: PropTypes.number,
+  description: PropTypes.string,
+  certificate: PropTypes.object,
```
victorporof: Ditto, no object prop types please.

app/ui/browser-modern/views/browser/page/cert-error.jsx
victorporof: Aren't all these props required? Then mark them as such.

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,58 @@
...
+  hidden: PropTypes.bool,
```
victorporof: Ditto, required props should be marked.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -73,33 +80,75 @@ class Page extends Component {
...
+      };
+
+      // Most cert errors are 501 from this event; the true, more descriptive
+      // error description can be retrieved from the main process. Just check
+      // here roughly if the error looks like a cert error.
+      if (Math.abs(errorCode) === 501 || /CERT/.test(errorDescription)) {
+        const { error, certificate } = await getCertificateError(validatedURL);
+        state.certificate = certificate;
+        state.description = error;
+      }
```
victorporof: All of this should be handled in a saga instead. Logic in component views should be kept at a minimum.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -73,33 +80,75 @@ class Page extends Component {
...
+        code: errorCode,
+        description: errorDescription,
+        url: validatedURL,
+      };
+
+      // Most cert errors are 501 from this event; the true, more descriptive
+      // error description can be retrieved from the main process. Just check
+      // here roughly if the error looks like a cert error.
+      if (Math.abs(errorCode) === 501 || /CERT/.test(errorDescription)) {
+        const { error, certificate } = await getCertificateError(validatedURL);
```
victorporof: This should *definitely* be in a saga, it's causes side effect.

victorporof:
@jsantell Setting the location bar text is handled via a side-effect now. React-managed components prove to be difficult to handle especially in our case where there is complex interaction with the underlying dom element (<input>). As such, any interaction with this dom element is done via the corresponding side effect actions.

victorporof:
Clearing review, there are a few things that need to be addressed.

app/ui/browser-modern/views/browser/page/index.jsx
victorporof: We're already resetting the entire page on navigation, see the page sagas.

app/ui/browser-modern/views/browser/page/cert-error.jsx
```diff
@@ -0,0 +1,90 @@
...
+import Style from '../../../../shared/style';
+
+const CONTAINER_STYLE = Style.registerStyle({
+  width: '700px',
+  margin: '0 auto',
+  display: 'block',
+});
+
+const CERTIFICATE_STYLE = Style.registerStyle({
+  backgroundColor: '#ddd',
```
victorporof: Background colors should be theme variables.

app/ui/browser-modern/views/browser/page/cert-error.jsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+    if (cert && cert.fingerprint) {
+      certElements.push(<li key="fingerprint">Fingerprint: {cert.fingerprint}</li>);
+    }
+
+    return (
+      <div className={CONTAINER_STYLE}>
+        <h1>Your connection is not secure</h1>
+        <p>
+          The owner of <strong>{this.props.url}</strong> has configured
```
victorporof: No `<strong>`.

app/ui/browser-modern/views/browser/page/index.jsx
victorporof: If this can be emitted multiple times, then a LOADED state can be set *after* a FAILED state being set.

app/ui/browser-modern/model/page-state.js
```diff
@@ -14,6 +14,10 @@ import Immutable from 'immutable';
 
 const PageState = Immutable.Record({
   load: undefined,
+  code: undefined,
+  description: undefined,
+  certificate: undefined,
+  url: undefined,
```
victorporof: Why is there an `url` property in the page state? There's a `location` in the page model. Having both is confusing.

victorporof:
Sorry about all the comments 😨

.eslintrc
jsantell: I don't understand -- my reasoning for (re)removing this is turning stateless function components into React components is tedious when needed, and there's no perf overhead for doing so

victorporof: Stateless function components don't have `shouldComponentUpdate` methods by definition. We shouldn't use those unless they don't receive any props or do receive props that *aways* change.

victorporof: See the Key Takeaways (TL;DR) in the article you linked to. We should always use pure components.

jsantell: I think we're in agreement then in disabling the preference of stateless function components

victorporof: Disabling it would stop telling you to turn this
```js
class Foo extends Component {
  render() { return <div/> }
}
```
into a stateless function component. That means we can't easily catch class components that don't have a `shouldComponentUpdate` just by linting. It's definitely not the way this rule was intended to be used though, so I guess I agree with removing it. I'd just like to have something that warns us when we have a class component that doesn't have a `shouldComponentUpdate`, no matter what warning that is.

app/ui/browser-modern/views/browser/page/net-error.jsx
jsantell: This was just an error page thrown together in the previous front end, we don't have any mockups so this was just made so we don't have a blank page.

app/ui/browser-modern/views/browser/page/cert-error.jsx
jsantell: Alright, but I disagree with all these theme variables that are for some ugly, unmocked up, barely styled page as a placeholder for a blank page that seems like something broke

app/ui/browser-modern/views/browser/page/cert-error.jsx
jsantell: I just removed all <strong> tags

jsantell:
@victorporof  Addressed most comments -- except the moving the certificate fetching to a saga, and calling the URL bar update action. I'd like to discuss how this should be done first, as I don't completely grok sagas yet and don't want to rewrite this again

victorporof:
@jsantell sure thing, let's talk about it when you get the time.

jsantell:
Added sagas for certificate fetching, and updating the URL bar via effect

app/ui/browser-modern/model/shared-prop-types.js
```diff
@@ -18,18 +18,29 @@ import ProfileModel from './profile';
...
 export const Profiles = ImmutablePropTypes.listOf(Profile.isRequired);
 
+export const SSLCertificate = PropTypes.shape({
+  issuer: PropTypes.string,
+  subjectName: PropTypes.string,
+  serialNumber: PropTypes.string,
+  validStart: PropTypes.string,
+  validExpiry: PropTypes.string,
+  fingerprint: PropTypes.string,
+});
```
victorporof: Any reason this couldn't be a record?

app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -137,3 +141,8 @@ function* capturePage({ pageId, webview }) {
 
   yield put(ProfileEffects.addCapturedPage(pageId, readerResult));
 }
+
+function* getCertificateError({ pageId, url }) {
+  const { error, certificate } = yield Certificate.getCertificateError(url);
```
victorporof: This should be `yield call(Certificate.getCertificateError, url)`. I'm surprised this works as is.

app/ui/browser-modern/views/browser/page/cert-error.jsx
```diff
@@ -0,0 +1,95 @@
...
+import Style from '../../../../shared/style';
+
+const CONTAINER_STYLE = Style.registerStyle({
+  width: '40%',
+  margin: '0 auto',
+  display: 'block',
+});
+
+const CERTIFICATE_STYLE = Style.registerStyle({
+  backgroundColor: '--theme-error-page-cert-background-color',
```
victorporof: This should be `var(--varname)`. Surprised this works otherwise.

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,62 @@
...
+import NetErrorPage from './net-error';
+import CertErrorPage from './cert-error';
+import Style from '../../../../shared/style';
+
+const ERROR_PAGE_STYLE = Style.registerStyle({
+  width: '100%',
+  height: '100%',
+  position: 'absolute',
+  color: '--theme-error-page-color',
+  backgroundColor: '--theme-error-page-background-color',
```
victorporof: Ditto here, `var(--varname)`

app/ui/browser-modern/views/browser/page/error-page.jsx
```diff
@@ -0,0 +1,62 @@
...
+  constructor(props) {
+    super(props);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
+  }
+
+  render() {
+    const code = Math.abs(this.props.pageState.get('code'));
+    const isCertError = code === 501 || // ERR_CERT_AUTHORITY_INVALID
+                        code === 129 || // ERR_SSL_WEAK_SERVER_EPHEMERAL_DH_KEY
+                        /CERT/.test(this.props.pageState.get('description')); // Catch all
```
victorporof: So no reason to move this code into some utility module? It would make things easier to test.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -97,26 +99,66 @@ class Page extends Component {
...
+        code: errorCode,
+        description: errorDescription,
+      };
+
+      // Most cert errors are 501 from this event; the true, more descriptive
+      // error description can be retrieved from the main process. Just check
+      // here roughly if the error looks like a cert error.
+      if (Math.abs(errorCode) === 501 || /CERT/.test(errorDescription)) {
+        this.props.dispatch(PageEffects.getCertificateError(this.props.pageId, validatedURL));
+      }
```
victorporof: There's lots and lots of difficult to test logic in here. I'd move most of it into utility functions in some module.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -97,26 +99,66 @@ class Page extends Component {
...
+        }));
+      }
+    });
+
+    this.webview.addEventListener('did-fail-load', (e) => {
+      // Like 'did-finish-load', but fired when the load failed or was cancelled.
+      // May be emitted multiple times for every single frame.
+      // Only emitted when a failure occurs, unlike `did-stop-loading`.
+      // Is fired before did-finish-load and did-stop-loading when failure
+      // occurs.
```
victorporof: Super nit: keep the "occurs" on the same line.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -177,8 +223,20 @@ Page.displayName = 'Page';
 
 Page.propTypes = {
   dispatch: PropTypes.func.isRequired,
-  pageId: PropTypes.string.isRequired,
   onMount: PropTypes.func.isRequired,
+  pageId: PropTypes.string.isRequired,
+  pageIndex: PropTypes.number.isRequired,
```
victorporof: `pageIndex` isn't being used anywhere, remove it.

app/ui/browser-modern/model/shared-prop-types.js
jsantell: Unlikely

app/ui/browser-modern/views/browser/page/error-page.jsx
jsantell: We could, but spending any more days on reimplenting this feature that was previously landed is something that might drive me mad, and just loading helpful error pages is a substantial UX gain and this blocks other patches

jsantell:
Fixed nits, made cert an Immutable record, any utility stuff should handle in a follow up, this patch is killing me

app/ui/browser-modern/views/browser/page/error-page.jsx
victorporof: :fearful:

Issue #1110, opened 2016-09-13
Ctrl/Cmd+Right on a page with no forward history sets page models in a loading state


victorporof:
Makes stuff look borked.

PR #1111, opened 2016-09-13
Prevent page navigations via keyboard shortcuts when pages aren't capable of navigation

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1110

### DIFF
--- a/app/ui/browser-modern/setup-ipc.js
+++ b/app/ui/browser-modern/setup-ipc.js
@@ -15,6 +15,7 @@ import { ipcRenderer } from '../../shared/electron';
 import * as ContentURLs from '../../shared/constants/content-pages-locations';
 import * as PageActions from './actions/page-actions';
 import * as PageEffects from './actions/page-effects';
+import * as PagesSelectors from './selectors/pages';
 import * as UISelectors from './selectors/ui';
 import * as UIActions from './actions/ui-actions';
 import * as UIEffects from './actions/ui-effects';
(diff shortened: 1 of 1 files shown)


app/ui/browser-modern/setup-ipc.js
```diff
@@ -47,19 +48,25 @@ export default function({ store, userAgentClient }) {
   });
 
   ipcRenderer.on('go-back', () => {
-    if (!UISelectors.getOverviewVisible(store.getState())) {
+    const pageId = PagesSelectors.getSelectedPageId(store.getState());
+    const pageState = PagesSelectors.getPageState(store.getState(), pageId);
```
bgrins: These two lines are a little clumsy compared to `PagesSelectors.getSelectedPage(store.getState()).state`.  Does hiding the `state` property inside getPageState provide much value, given that we will immediately need to dig into the state object's internals?

victorporof: Eh, I have no strong opinions about this. Idiomatic redux would be to not access any properties directly on the app state ever (outside of the reducers of course), and just use selectors. However, I feel like that's a bit too much hassle at times.

Our codebase is somewhat of a mixed bag, strongly but not strictly preferring selectors, so this can go either direction.

bgrins: Yeah to be really strict about that it seems we'd need selectors for getPageCanGoBack, getPageCanRefresh, etc (which seems like overkill), as we eventually end up directly working with the state whether it's accessing `page.state.canGoBack` or `getPageState().canGoBack`. I'd like to hear @linclark's thoughts on how she'd handle this case.

PR #1113, opened 2016-09-13
Don't use redux `connect` to inject props unrelated to a component's render output, to avoid unnecessary re-renders

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1112

### DIFF
--- a/app/ui/browser-modern/views/browser/navbar/index.jsx
+++ b/app/ui/browser-modern/views/browser/navbar/index.jsx
@@ -61,11 +61,14 @@ class NavBar extends Component {
   }
 
   handleOverviewButtonClick = () => {
-    if (this.props.isOverviewVisible) {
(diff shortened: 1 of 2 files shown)


app/ui/browser-modern/views/browser/navbar/index.jsx
```diff
@@ -61,11 +61,14 @@ class NavBar extends Component {
   }
 
   handleOverviewButtonClick = () => {
-    if (this.props.isOverviewVisible) {
-      this.props.dispatch(UIActions.hideOverview());
-    } else {
-      this.props.dispatch(UIActions.showOverview());
-    }
+    this.props.dispatch((dispatch, getState) => {
```
jsantell: Never thought of inlining the thunk middleware stuff like this -- nice!

PR #1119, opened 2016-09-13
Optimistically set page location and update the urlbar before the navigation happens

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1118

Doing this so that the UI shows the intended location instead of just some blank text. This could still happen though when clicking the back/forward buttons, since we don't know the location before the navigation happens; in those cases, just display 'Loading...'

### DIFF
--- a/app/ui/browser-modern/sagas/page-sagas.js
+++ b/app/ui/browser-modern/sagas/page-sagas.js
@@ -19,6 +19,7 @@ import * as UserAgent from '../../shared/util/user-agent';
 import * as PageActions from '../actions/page-actions';
 import * as PageEffects from '../actions/page-effects';
 import * as ProfileEffects from '../actions/profile-effects';
+import * as UIEffects from '../actions/ui-effects';
 import * as EffectTypes from '../constants/effect-types';
 
 export default function() {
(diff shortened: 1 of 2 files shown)


app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -74,6 +75,11 @@ function* destroyPageSession({ page, currentPageCount }) {
 function* naviatePageTo({ pageId, webview, location }) {
   yield put(PageActions.resetPageData(pageId));
   yield put(PageActions.setPageState(pageId, { load: PageState.STATES.PRE_LOADING }));
+  // Optimistically set page location and update the urlbar before the actual
+  // navigation happens, so that the UI shows the intended location instead of
+  // just some blank text.
+  yield put(PageActions.setPageDetails(pageId, { location }));
```
bgrins: These two actions seem to always be used together at least in the current 3 usages (including this one).  Is it possible / desirable to make the ui saga always run this setURLBarValue logic when the setPageDetails action happens with a location property?  I'm not sure if there's a time we'd want to setPageDetails to change the location *without* also updating the associated URL bar, and it would be an easy mistake to make

bgrins:
Works for me, but see the inline comment

app/ui/browser-modern/sagas/page-sagas.js
victorporof: Good point, let's address this in a followup.

victorporof:
Filed https://github.com/mozilla/tofino/issues/1139

Issue #1102, opened 2016-09-12
Restore session when browser starts or is reloaded


victorporof:
Should restore all tabs and windows. Would help a lot with development as well, since reloading the app would clear the app state.

Will work on this.

victorporof:
Depends on https://github.com/electron/electron/issues/7186 or tracking history ourselves as described in https://github.com/mozilla/tofino/issues/996#issuecomment-246672736

victorporof:
Unasigning for now.

victorporof:
Found out about https://github.com/mozilla/tofino/pull/1120#issuecomment-246906935

victorporof:
Still needs a way to inject history, will wait until that's done. Tracking in https://github.com/electron/electron/issues/7186

victorporof:
@Mossop where should the session be stored on disk and when? assuming the app isn't able to exit cleanly, should the app state be stored to disk at regular intervals?

Issue #1101, opened 2016-09-11
Cmd+Shift+T to restore the previous tab (including tab history)


victorporof:
I frequently rely on this in Firefox and Chrome after I closed a tab.

victorporof:
Depends on https://github.com/electron/electron/issues/7186 or tracking history ourselves as described in https://github.com/mozilla/tofino/issues/996#issuecomment-246672736

victorporof:
Unasigning for now.

victorporof:
Found out about https://github.com/mozilla/tofino/pull/1120#issuecomment-246906935

Issue #1011, opened 2016-08-25
Cmd+Click on back/forward buttons should open/fork in new tab


jsantell:

victorporof:
Depends on https://github.com/electron/electron/issues/7186 or tracking history ourselves as described in https://github.com/mozilla/tofino/issues/996#issuecomment-246672736

victorporof:
Unasigning for now.

victorporof:
Found out about https://github.com/mozilla/tofino/pull/1120#issuecomment-246906935

Issue #996, opened 2016-08-24
Webview forward navigation doesn't always work


bgrins:
STR:

Open https://duckduckgo.com/?q=test
Click on a result and let the page load
Press back
Press forward

Expected:
You end up on the result page

Actual:
Pressing forward doesn't do anything

bgrins:
This is getting through the webViewController alright and this call to webview.goForward() is happening: https://github.com/mozilla/tofino/blob/5a7444f6a034395dfca5f0da40a117c2c453d369/app/ui/browser/views/page/index.jsx#L67

darkwing:
Not sure if it's the same issue but CMD+LEFTARROW doesn't work for me at all, nor does CMD+RIGHTARROW

bgrins:
> Not sure if it's the same issue but CMD+LEFTARROW doesn't work for me at all, nor does CMD+RIGHTARROW

Have a fix for that in #995

Mossop:
This appears to be broken in electron itself and brave is affected too. I'll take it.

victorporof:
AFAIK there's no way to access history for an electron's webview (as an array of urls for example). So we're going to have to keep a list of page histories in our app state anyway to fix https://github.com/mozilla/tofino/issues/1101, https://github.com/mozilla/tofino/issues/1102, https://github.com/mozilla/tofino/issues/1011 and https://github.com/mozilla/tofino/issues/1017 anyway.

If we do that, we might as well stop relying on webview's `goForward` and just perform a normal `setAttribute(src)` where src would be the url next in that page's history, from our app state. We don't have any support for bfcache at the moment at all, so this won't regress anything.

It's a simple workaround instead of a fix.

Mossop:
It's extremely complicated to keep track of history, that is why this bug happens, because electron are attempting to do it themselves instead of relying on the history implementation in chromium and getting themselves out of sync. Plus reloading instead of using back/forward means we lose all the bfcache behaviours which would make history navigations faster.

I'm more inclined to hack on electron and give us a way to bypass electron's broken history implementation and give us a way to access chromium's.

victorporof:
I agree. I tried implementing something that keeps track of history based on the webview navigation events, and it's indeed extremely difficult to properly handle everything right.

I filed https://github.com/electron/electron/issues/7186 for getting access to history instead of doing this ourselves.

Mossop:
This is fixed upstream by https://github.com/electron/electron/pull/7175

[...]

PR #1120, opened 2016-09-13
POC for storing individual page histories in the app state

victorporof:
This would help fixing #1101, #1102, #1011 and #1017, and maybe potentially even https://github.com/mozilla/tofino/issues/996, but I still think tracking history ourselves is a bad idea. We should wait for https://github.com/electron/electron/issues/7186

See https://github.com/mozilla/tofino/issues/996#issuecomment-246705312


victorporof:
`webview.getWebContents().history` gives us the urls in the history. It's undocumented and unsupported api, but alas.

PR #1122, opened 2016-09-14
Add ability to view tab history

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1017


app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -69,28 +69,35 @@ class Page extends Component {
...
       // Only emitted once.
-      const url = this.webview.getURL();
-      const isTofino = url.startsWith(Endpoints.TOFINO_PROTOCOL);
-      this.props.dispatch(PageActions.setPageDetails(this.props.pageId, { location: url }));
-      this.props.dispatch(UIEffects.setURLBarValue(this.props.pageId, url));
-      this.props.dispatch(UIEffects.focusURLBar(this.props.pageId, { select: isTofino }));
+      const pageId = this.props.pageId;
+      const isTofino = e.url.startsWith(Endpoints.TOFINO_PROTOCOL);
+      const history = this.webview.getWebContents().history;
+      const historyIndex = this.webview.getWebContents().getActiveIndex();
```
Mossop: getWebContents() is a sync IPC call to the main process (sucks I know), so please cache the result rather than calling it twice.

victorporof: Ugh, good to know. Thanks!

app/ui/shared/widgets/dropdown-menu-btn.jsx
```diff
@@ -0,0 +1,152 @@
...
+      selectedIndex: -1,
+    };
+  }
+
+  componentDidMount() {
+    window.addEventListener('click', this.handleWindowClick);
+    window.addEventListener('keyup', this.handleWindowKeyUp);
+  }
+
+  handleWindowClick = e => {
```
jsantell: This seems like the most React-y way to handle this, but seems like this could get costly fast with many keyup/click listeners on the window. Not sure of a better way, hmm..

app/ui/shared/widgets/list-item.jsx
```diff
@@ -34,6 +34,24 @@ class ListItem extends Component {
       !isEqual(omit(this.props, ['children']), omit(nextProps, ['children']));
   }
 
+  handleMouseDown = e => {
+    if (this.props.onMouseDown) {
```
jsantell: nit: Setting default props and calling noops rather than conditionals

app/ui/shared/widgets/list.jsx
```diff
@@ -40,17 +40,22 @@ class List extends Component {
 
   render() {
     return (
-      <ul {...this.props}
+      <ul {...omit(this.props, Object.keys(OmittedContainerProps))}
         className={`widget-list ${LIST_STYLE} ${this.props.className || ''}`}>
-        {this.props.children}
+        {this.props.reversed ? [...this.props.children].reverse() : this.props.children}
```
jsantell: If we only have one child, `this.props.children` is not an array/iterable, so `...this.props.children` would throw `children is not iterable`, the reason for `React.Children.toArray(this.props.children)` utility

app/ui/shared/widgets/search.jsx
```diff
@@ -29,7 +30,15 @@ const INPUT_STYLE = Style.registerStyle({
...
-    return !isEqual(this.props, nextProps);
+    // Since this component uses plain js objects as props, we need to perform
+    // deep quality checks on them to make sure we don't unnecessarily rerender.
+    // However, there is no need to re-render this parent component when the
+    // children change, which would happen when doing naive deep equality.
+    // Furthermore, children in this component may contain immutable.js props,
+    // which are throwing warnings when accessing some of their properties
+    // while doing deep equality checks using lodash's `deepEqual`.
+    return !shallowEqual(this.props.children, nextProps.children) ||
+      !isEqual(omit(this.props, ['children']), omit(nextProps, ['children']));
```
jsantell: This is a non-intuitive sCU method, and since using it more than once, should probably go somewhere else sharable. Actually, a `lib/should-component-update` with some common patterns would be helpful, as we use the PureRenderMixin, simple isEqual, and then this more complicated scenario which could all be exported from the same place (and wouldn't need to remember about `fbjs` module path to shallowEqual)

app/ui/shared/widgets/search.jsx
```diff
@@ -92,6 +102,10 @@ Search.propTypes = {
   style: PropTypes.object,
   className: PropTypes.string,
   hidden: PropTypes.bool,
+  children: PropTypes.oneOfType([
```
jsantell: This is probably a good candidate for a SharedPropType

jsantell:
Er, I mean control+click opening the history, not Cmd+Click, which forks a tab.

app/ui/browser-modern/views/browser/page/index.jsx
Mossop: I've done no measurements so can't say but the main process is generally not doing anything so I expect it is unlikely to cause significant jank. It's worth noting that many of the methods on webview cause sync IPC calls to the main process so generally we shouldn't expect it to be cheap to interact with.

victorporof:
@jsantell we have a different issue on file for cmd+click on buttons: https://github.com/mozilla/tofino/issues/1011 it's definitely not in the scope of this PR

app/ui/browser-modern/views/browser/location/index.jsx
```diff
@@ -68,11 +68,11 @@ class Location extends Component {
 
   handleInputKeyDown = e => {
     if (e.key === 'Enter') {
-      this.handleAutocompletionSelect({ uri: e.target.value });
+      this.handleAutocompletionPick({ uri: e.target.value });
     }
   }
 
-  handleAutocompletionSelect = data => {
+  handleAutocompletionPick = ({ data }) => {
```
victorporof: I strongly disagree. When you hover over autocompletions or history items, they're "selected" highlighted. Only when you click on them or press enter they become "chosen", or "picked". I don't care about what other word we use, but "select" is poor because it implies this function is called when list items are hovered.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -86,9 +89,14 @@ class Page extends Component {
...
       }
-      this.props.dispatch(PageActions.setPageDetails(this.props.pageId, { location: e.url }));
-      this.props.dispatch(UIEffects.setURLBarValue(this.props.pageId, e.url));
-      this.props.dispatch(ProfileEffects.addRemoteHistory(this.props.pageId));
+      const pageId = this.props.pageId;
+      const webContents = this.webview.getWebContents();
+      const history = webContents.history;
+      const historyIndex = webContents.getActiveIndex();
+      this.props.dispatch(PageActions.setPageDetails(pageId, { location: e.url }));
+      this.props.dispatch(PageActions.setLocalPageHistory(pageId, history, historyIndex));
```
victorporof: Maybe? File an issue.

app/ui/shared/widgets/dropdown-menu-btn.jsx
victorporof: Why would this get costly? Sure I can imagine a world where there's a lot of listeners because we use hundreds of DropdownMenuBtn for some reason, but why would things get costly?

app/ui/shared/widgets/list-item.jsx
victorporof: Nice idea! Will do.

app/ui/shared/widgets/list.jsx
victorporof: Good catch!

app/ui/shared/widgets/search.jsx
victorporof: Will do.

app/ui/shared/widgets/search.jsx
victorporof: Yup!

victorporof:
@jsantell in what scenario would things get out of sync? can you please elaborate?

PR #1134, opened 2016-09-14
Add a badge displaying SSL status in the URL bar. Fixes #128. r=vporof

### LINKED ISSUE #128: Add HTTPS Indicator
There's no indication of HTTP vs. HTTPS.  An icon and a very light green background would be sweet.

jsantell:
This does not handle the scenario where it loads a page with SSL errors and fails, since it'll still display a green lock (although the page doesn't load, so no bad things can happen).

Once #1107 lands, can do a follow up to parse any certificate errors available and display invalid SSL icons even when the page doesn't load.

### DIFF
--- a/app/ui/browser-modern/views/browser/location/index.jsx
+++ b/app/ui/browser-modern/views/browser/location/index.jsx
@@ -16,6 +16,7 @@ import { connect } from 'react-redux';
 
 import Style from '../../../../shared/style';
 import Btn from '../../../../shared/widgets/btn';
+import SecurityBadge from './security-badge';
 import AutocompletedSearch from '../../../../shared/widgets/autocompleted-search';
 import AutocompletionListItem from './autocompletion-list-item';
 
@@ -80,6 +81,10 @@ class Location extends Component {
     // TODO
   }
 
+  handleConnectionButtonClick= () => {
+    // TODO
+  }
+
   handleBookmarkButtonClick = () => {
     const pageId = this.props.pageId;
     const bookmarked = this.props.pageIsBookmarked;
@@ -95,6 +100,12 @@ class Location extends Component {
           imgWidth="16px"
           imgHeight="16px"
           onClick={this.handleInfoButtonClick} />
+        <SecurityBadge title="Connection"
+          url={this.props.pageLocation}
+          pageState={this.props.pageState}
+          imgWidth="16px"
+          imgHeight="16px"
+          onClick={this.handleConnectionButtonClick} />
         <AutocompletedSearch ref={e => this.awesomebar = e}
           className={LOCATION_BAR_INPUT_STYLE}
           defaultValue={this.props.pageLocation}
@@ -121,6 +132,7 @@ Location.propTypes = {
   pageId: PropTypes.string.isRequired,
   pageLocation: PropTypes.string.isRequired,
   pageIsBookmarked: PropTypes.bool.isRequired,
+  pageState: SharedPropTypes.PageState,
   locationAutocompletions: SharedPropTypes.LocationAutocompletions,
   onNavigate: PropTypes.func.isRequired,
 };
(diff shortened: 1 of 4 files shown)


app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+import Btn from './btn';
+
+class SecurityBadge extends Component {
+  constructor(props) {
+    super(props);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
+  }
+
+  render() {
+
```
victorporof: Nit: extra newline

app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+    // Since we have unconfigurable security settings, if we access an HTTPS page and it loads,
+    // it is considered secure. Otherwise, all HTTP pages are insecure. Also, since
+    // we block all mixed content, this is even easier.
+    const image = /^https:/.test(this.props.url) ? 'ssl-secure.svg' : 'ssl-insecure.svg';
```
victorporof: I wonder if this should be moved to some helper function.

app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+    // If we have a tofino:// page, or the URL is empty (occurs during
+    // some page loads in our state), just hide the icon
+    let hidden = this.props.hidden;
+    if (/^tofino:/.test(this.props.url) || this.props.url === '') {
```
victorporof: ...especially because of this.

app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+      </Btn>
+    );
+  }
+}
+
+SecurityBadge.displayName = 'SecurityBadge';
+
+const OmittedContainerProps = {
+  url: PropTypes.string.isRequired,
+  image: PropTypes.string,
```
victorporof: This prop is both omitted and set on the Btn component. Should move it to the propTypes.

app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+  hidden: PropTypes.bool,
+};
+
+SecurityBadge.propTypes = {
+  ...omit(Btn.propTypes, Object.keys(OmittedContainerProps)),
```
victorporof: This should be just ...OmittedContainerProps, not sure why you're using `omit` here.

victorporof:
Just look into fixing the tests, everything else looks good.

app/ui/shared/widgets/security-badge.jsx
jsantell: Yeah, it's weird since some properties don't want to pass in Btn, some you do, and some you do want to pass in after modifying.

app/ui/shared/widgets/security-badge.jsx
jsantell: Probably can take a Page and turn that into security info, but no need for that yet

app/ui/shared/widgets/security-badge.jsx
```diff
@@ -0,0 +1,62 @@
...
+    let hidden = this.props.hidden;
+    if (/^tofino:/.test(this.props.url) || this.props.url === '') {
+      hidden = true;
+    }
+
+    return (
+      <Btn {...omit(this.props, Object.keys(OmittedContainerProps))}
+        image={image}
+        hidden={hidden}>
+        {React.Children.toArray(this.props.children)}
```
victorporof: I didn't know about this: `React.Children.toArray` 👍

app/ui/shared/widgets/security-badge.jsx
victorporof: We should pass in complex data structures as props to child components. A page model contains *many things*, most of which would be unrelated to this component's rendering. Therefore, if an unrelated page's property changes for example, we wouldn't want this security badge to rerender just because the new page model would be a new immutable.js object.

Issue #1138, opened 2016-09-15
The bookmarks app menu isn't being populated with bookmarks


victorporof:
A mismatch between the action type sent and the action type expected.

PR #1140, opened 2016-09-15
Properly populate the bookmarks app menu with recent bookmarks

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1138


ncalexan:
r+ -- this is the same as my patch :)

Issue #1125, opened 2016-09-14
Trying to switch to a tab that doesn't exist via cmd/ctrl+n logs an error


victorporof:
If we have 3 tabs open and type Cmd/Ctrl+4 an error is logged in the console.

PR #1144, opened 2016-09-15
Don't try to switch via keyboard shortcuts to tabs that don't exist at certain indices

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1125


app/ui/browser-modern/selectors/pages.js
```diff
@@ -40,6 +40,14 @@ export function getPageByIndex(state, index) {
   return getPageById(state, getPageIdByIndex(index));
 }
 
+export function hasPageWithId(state, id) {
```
jsantell: This is unused at the moment

PR #1147, opened 2016-09-15
Switch to last tab on cmd/ctrl+9,r=vporof

### LINKED ISSUE #1146: cmd/ctrl+9 should always select the last tab
In Firefox/Chrome even if you have more or less than 9 tabs, pressing ctrl+9 selects the last one

bgrins:
Fixes #1146


app/main/hotkeys.js
```diff
@@ -33,6 +33,7 @@ const HOTKEYS = new Map([
   ['CmdOrCtrl+6', bw => () => bw.webContents.send('select-tab-index', 5)],
   ['CmdOrCtrl+7', bw => () => bw.webContents.send('select-tab-index', 6)],
   ['CmdOrCtrl+8', bw => () => bw.webContents.send('select-tab-index', 7)],
+  ['CmdOrCtrl+9', bw => () => bw.webContents.send('select-tab-index', -1)], // last tab
```
victorporof: Why not send a `select-last-tab` action instead of complicating the `select-tab-index` actions?

bgrins: sure, makes sense

PR #1149, opened 2016-09-15
Update eslint-config-airbnb to version 11.1.0

victorporof:
Greenkeeper has been trying to update our airbnb config for quite a while now, but we kept deleting those PRs: https://github.com/mozilla/tofino/pull/1145. It's about time we did something about this.

### DIFF
--- a/test/unit/user-agent-service/test-visited.js
+++ b/test/unit/user-agent-service/test-visited.js
@@ -21,7 +21,7 @@ const { DB_ADD, TEMPIDS } = helpers;
 const perfTesting = false;
 
 describe('Storage.recentlyStarred', () => {
-  it('Stars and unstars.', async () => {
+  it('Stars and unstars.', async function() {
     const db = new ProfileDatomStorage('.');
     const A = await db.recentlyStarred();
     expect(mori.count(A)).toBe(0);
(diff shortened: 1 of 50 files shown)


test/unit/user-agent-service/test-visited.js
```diff
@@ -21,7 +21,7 @@ const { DB_ADD, TEMPIDS } = helpers;
 const perfTesting = false;
 
 describe('Storage.recentlyStarred', () => {
-  it('Stars and unstars.', async () => {
+  it('Stars and unstars.', (async) () => {
```
Mossop: This doesn't look right, there are a few other instances throughout too.

victorporof:
Jordan can you please take a quick look at the frontend changes?

victorporof:
@Mossop please take a look at this again.

test/unit/user-agent-service/test-visited.js
```diff
@@ -21,7 +21,7 @@ const { DB_ADD, TEMPIDS } = helpers;
 const perfTesting = false;
 
 describe('Storage.recentlyStarred', () => {
-  it('Stars and unstars.', async () => {
+  it('Stars and unstars.', async function() {
```
Mossop: It seems like it's a bug in the rule if this is needed. Fine to take it like this but is it worth checking if it is reported upstream?

victorporof: Will report it upstream. I twas definitely a bug to wrap `async` in parens.

jsantell:
This requires no parens in single-arg arrow functions, and no async arrow functions? I don't see the benefits for this strictness, and my opinions of death-by-a-thousand-paper-cuts on our many strict linting rules are well known, but the code changes look fine.

PR #1172, opened 2016-09-19
Propogate click events from webviews and close popups like autocomple…

### LINKED ISSUE #1123: Autocomplete list doesn't hide when clicking away from it in the browser
From #1091:

STR:
* Type something in a URL bar to get autocomplete list to open
* Click anywhere in the browser (web contents, nav bar...)

Expected:
* Autocompletes go away

The autocomplete list doesn't disappear even when I click elsewhere in the browser. Seems the only way to get it to go away is by pressing ESC - but once I've clicked elsewhere the URL bar loses focus so ESC doesn't close it.

jsantell:
…te list when clicking outside the chrome area. Fixes #1123. r=vporof

How's something like this for handling clicks? The webview event doesn't emit a click event on its own; also, have not tested, but wonder if content can stop click events from hitting this, and if so, how common is that.

### DIFF
--- a/app/ui/browser-modern/views/browser/page/index.jsx
+++ b/app/ui/browser-modern/views/browser/page/index.jsx
@@ -204,6 +204,26 @@ class Page extends Component {
           break;
       }
     });
+
+    // We cannot bind click events on the actual webview node, but mousedown
+    // works, so convert that to a propogated click.
+    this.webview.addEventListener('mousedown', e => {
+      this.handleWebviewClick(e);
+    });
+  }
+
+  /**
+   * Intercept and propogate clicks from the webview so we can react (like close popups)
+   * on events from webview.
+   */
+  handleWebviewClick() {
+    const ev = new MouseEvent('click', {
+      bubbles: true,
+      cancelable: true,
+      view: window,
+      relatedTarget: this.webview,
+    });
+    this.webview.dispatchEvent(ev);
   }
 
   render() {
--- a/app/ui/shared/widgets/autocompleted-search.jsx
+++ b/app/ui/shared/widgets/autocompleted-search.jsx
@@ -37,11 +37,32 @@ class AutocompletedSearch extends Component {
     this.shouldComponentUpdate = ComponentUtil.shouldMixedPropsComponentUpdate.bind(this, {
       immutables: ['dataSrc'],
     });
+    this.handleWindowClick = this.handleWindowClick.bind(this);
 
     this.state = {
       showSelectionList: false,
       selectedIndex: 0,
     };
+  }
+
+  componentWillMount() {
+    if (typeof window === 'object') {
+      window.addEventListener('click', this.handleWindowClick);
+    }
+  }
+
+  componentWillUnmount() {
+    if (typeof window === 'object') {
+      window.removeEventListener('click', this.handleWindowClick);
+    }
+  }
+
+  handleWindowClick() {
+    // This component's button won't be available if hasn't rendered yet.
+    // We also can't call `setState` on components which haven't rendered yet.
+    if (this.state.showSelectionList && this.inputbar) {
+      this.setState({ showSelectionList: false });
+    }
   }
 
   handleInputChange = e => {


app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -204,13 +204,30 @@ class Page extends Component {
 
     this.webview.addEventListener('ipc-message', e => {
```
bgrins: Looks like we could do:

        this.webview.addEventListener('mousedown', e => {

        });

And that fires, but 'click' doesn't for some reason.

jsantell: Mousedown would be sufficient for this, I think. Can propogate as 'click' if needed

jsantell: Also wouldn't need the preload script which seems shaky at best

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -35,13 +35,30 @@ class AutocompletedSearch extends Component {
...
 
     this.state = {
       showSelectionList: false,
       selectedIndex: 0,
     };
   }
 
+  componentWillMount() {
+    if (typeof window === 'object') {
+      window.addEventListener('click', this.handleWindowClick);
```
victorporof: Why the need to check if the window global exists? I thought you said this was an eslint error.

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -35,13 +35,30 @@ class AutocompletedSearch extends Component {
...
+    }
+  }
+
+  componentWillUnmount() {
+    if (typeof window === 'object') {
```
victorporof: Please add something similar for the history dropdown.

victorporof:
Please also handle the escape key, either here or in a followup: https://github.com/mozilla/tofino/blob/master/app/ui/shared/widgets/dropdown-menu-btn.jsx

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -35,13 +35,30 @@ class AutocompletedSearch extends Component {
...
+      window.removeEventListener('click', this.handleWindowClick);
+    }
+  }
+
+  handleWindowClick() {
+    this.setState({ showSelectionList: false });
```
victorporof: React will warn if setState is called before this component renders.

app/ui/shared/widgets/autocompleted-search.jsx
jsantell: This hits in our unit tests, it wasn't linting, where `window` does not exist

jsantell:
This already handles the escape key scenario

app/ui/shared/widgets/autocompleted-search.jsx
jsantell: @victorporof  I'm not sure what you mean by this -- a window safety? or unmount clean up? The closing on escape/click seems to be working fine for it. Can you clarify in a follow up?

app/ui/shared/widgets/autocompleted-search.jsx
jsantell: Fixed!

app/ui/browser-modern/views/browser/page/index.jsx
jsantell: This seems to be working OK

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -37,13 +37,34 @@ class AutocompletedSearch extends Component {
...
+  }
+
+  componentWillUnmount() {
+    if (typeof window === 'object') {
+      window.removeEventListener('click', this.handleWindowClick);
+    }
+  }
+
+  handleWindowClick() {
+    // This component's button won't be available if hasn't rendered yet.
```
victorporof: There's no button in this component.

Issue #1095, opened 2016-09-11
Ability to make Tofino the default browser


victorporof:

jsantell:
Electron doesn't give us anything like this, and this looks like we'll need platform specific settings for each. Few other npm modules that do similar, but all just call the browser's flags to let them do the setting, so makes sense for us to expose an arg to do similar:
https://github.com/gokulkrishh/set-default-browser/blob/master/index.js

jsantell:
Actually, in Firefox, we just set protocol handlers https://dxr.mozilla.org/mozilla-central/source/browser/components/shell/nsMacShellService.cpp#64

Electron has similar for Windows/OSX atleast:
http://electron.atom.io/docs/all/#appsetasdefaultprotocolclientprotocol-path-args-macos-windows

PR #1175, opened 2016-09-19
Ability to set Tofino as default browser on OSX

jsantell:
…ixes #1095. r=vporof,mossop

Would like to try this on Windows too. @mossop, how does this look?

Some issues:
* May not work in Linux -- Electron API seems OSX/Windows only
* OSX runs in 'single instance' mode -- you can't open up a new instance of Tofino (easily). When receiving new links for Windows, it sounds like it opens a whole new instance, not just a browser window, with its own main process. In which case, we need to parse process.argv. Is this correct?
* Opening a link when Tofino is closed opens the default Electron window if you make the development browser default. Ran into this with Brave during opening before (not with a link or as default), so seems like some boilerplate code. If a packaged build is the default, it opens Tofino but not with a URL.

With clarification on how Windows does/should handle this, I think supporting parsing the process args and starting with that URL open should be a follow up.

screen shot 2016-09-19 at 12 06 52 pm


jsantell:
OSX does not receive any additional process.argv args when opening a file (when it's closed), and should use the open-file/open-url handlers (but can't seem to recreate). https://github.com/electron/electron/issues/6310

Windows -should- receive process.argv for args for these events https://github.com/electron/electron/issues/1746

jsantell:
After talking with @Mossop, it sounds like having this work in Windows 10 is hard, and the Electron APIs give us the ability on older Windows atleast. I think we should turn this into OSX only, and handle Windows in a follow up, if it's possible, and not even ask about default browsers on non-OSX for now (because setting as default, and it not working is incredibly frustrating)

app/main/protocols.js
```diff
@@ -10,10 +10,11 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
...
 
-import { protocol } from 'electron';
+import { protocol, app } from 'electron';
 import * as endpoints from '../shared/constants/endpoints';
 
 const tofinoColonSlash = new RegExp(`^${endpoints.TOFINO_PROTOCOL}:/`);
+const DEFAULT_PROTOCOLS = ['http', 'https', 'tofino'];
 
 export function registerStandardSchemes() {
   protocol.registerStandardSchemes([endpoints.TOFINO_PROTOCOL, 'http', 'https']);
```
victorporof: This should use DEFAULT_PROTOCOLS.

app/main/protocols.js
```diff
@@ -10,10 +10,11 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
-import { protocol } from 'electron';
+import { protocol, app } from 'electron';
 import * as endpoints from '../shared/constants/endpoints';
 
 const tofinoColonSlash = new RegExp(`^${endpoints.TOFINO_PROTOCOL}:/`);
+const DEFAULT_PROTOCOLS = ['http', 'https', 'tofino'];
```
victorporof: Don't hardcode 'tofino', use the predefined endpoints.

app/ui/browser-modern/index.js
```diff
@@ -11,6 +11,7 @@ specific language governing permissions and limitations under the License.
 */
 
 /* eslint import/imports-first: "off" */
+/* eslint no-alert: "off" */
```
victorporof: lol

app/ui/browser-modern/index.js
```diff
@@ -43,3 +44,10 @@ const userAgentClient = setupUserAgentClient({
 
 setupFrontend({ store, userAgentClient });
 setupIPCCommunication({ store, userAgentClient });
+
+// Fired on the first browser window when default browser is not Tofino.
+ipcRenderer.once('should-set-default-browser', () => {
```
victorporof: Shouldn't this be in the `setupIPCCommunication` module?

app/ui/browser-modern/index.js
```diff
@@ -43,3 +44,10 @@ const userAgentClient = setupUserAgentClient({
...
+  if (confirm('Would you like to be Tofino your default browser?')) {
```
victorporof: This should be an UI effect if we really want to write it properly.

On the `should-set-default-browser`, an CONFIRM_DEFAULT_BROWSER effect should be dispatched, handled in the UI sagas module, which on confirmation sends a `set-default-browser` ipc message.

jsantell:
Ok, now it works when opening links when Tofino's closed (see `URL_QUEUE`), and only does this in production on OSX for now. Few other changes. @Mossop @victorporof this is hairy so please review again, thanks!

app/main/browser-window.js
```diff
@@ -71,15 +85,39 @@ export async function createBrowserWindow(userAgentClient, onload) {
...
-    if (!browser.isDestroyed()) {
-      browser.show();
-      if (BUILD_CONFIG.development && error) {
-        browser.openDevTools({ detach: true });
+    browser.show();
+    if (BUILD_CONFIG.development && error) {
+      browser.openDevTools({ detach: true });
+    }
+    // If Tofino is not the default browser
+    if (shouldCheckDefaultBrowser) {
```
victorporof: s/shouldCheck/shouldSet?

app/main/browser-window.js
```diff
@@ -71,15 +85,39 @@ export async function createBrowserWindow(userAgentClient, onload) {
...
+  // already the default browser.
+  const shouldCheckDefaultBrowser = browserWindows.length === 0 &&
+                                    process.env.NODE_ENV === 'production' &&
+                                    process.platform === 'darwin' &&
+                                    !protocols.isDefaultBrowser();
+
   // 'window-ready' is called if an error occurred loading the client, or once
   // the client has connected to the User Agent Service correctly.
   browser.once('window-ready', error => {
+    browser.IS_READY = true;
```
victorporof: We have a Deferred class in ui/shared/util/deferred.js, seems like it would work better than IS_READY booleans. Should move that in the app/shared and use it here.

app/main/browser-window.js
```diff
@@ -119,6 +157,40 @@ export async function closeBrowserWindow(bw) {
...
+ * resolves when the BrowserWindow's frontend is set up and
+ * able to receive and handle ipc commands
+ */
+export async function waitUntilReady(bw) {
+  if (bw.IS_READY) {
+    return bw;
+  }
+  return new Promise(resolve => {
+    bw.once('window-ready', () => resolve(bw));
+  });
```
victorporof: See comment above about using promises through the Deferred class.

app/ui/browser-modern/constants/effect-types.js
```diff
@@ -50,3 +50,4 @@ export const GET_CERTIFICATE_ERROR = 'GET_CERTIFICATE_ERROR_EFFECT';
 export const FOCUS_URL_BAR = 'FOCUS_URL_BAR_EFFECT';
 export const SET_URL_BAR_VALUE = 'SET_URL_BAR_VALUE_EFFECT';
 export const SHOW_DOWNLOAD_NOTIFICATION = 'SHOW_DOWNLOAD_NOTIFICATION_EFFECT';
+export const CONFIRM_DEFAULT_BROWSER = 'CONFIRM_DEFAULT_BROWSER';
```
victorporof: The string should be 'CONFIRM_DEFAULT_BROWSER_EFFECT'

app/ui/browser-modern/sagas/ui-sagas.js
```diff
@@ -10,8 +10,11 @@ CONDITIONS OF ANY KIND, either express or implied. See the License for the
 specific language governing permissions and limitations under the License.
 */
 
+/* eslint no-alert: "off" */
```
victorporof: Nit: localize this comment to the confirmDefaultBrowserFunction
```
/* eslint-disable no-alert */
... code ...
/* eslint-enable no-alert */

app/main/browser-window.js
jsantell: This isn't setting the default browser, just if it should ask the user. shouldAsk is probably better

app/main/browser-window.js
jsantell: The reason for the boolean is scenarios in which we've missed the window-ready event

jsantell: Gonna let this one go for now:
1) Not sure how we can get around some sort of flag for BW's that are already ready -- store the once promise on the bw itself? `browser.once('window-ready', resolve); browser._ready = deferred` ?
2) I think this stuff will be changed when working on #1210, to handle BW lifetimes similar to Brave [here](https://github.com/brave/browser-laptop/blob/master/app/cmdLine.js)

Can change to whatever, actually prefer deferreds, but this code's going to be a bit transient for a few days

PR #1203, opened 2016-09-20
Update electron to v1.4.0. Fixes #996

### LINKED ISSUE #996: Webview forward navigation doesn't always work

Mossop:


bgrins:
Nice!  Confirmed it fixes #996

PR #1213, opened 2016-09-20
Add a way to do integration testing inside the app and a basic integration test.

Mossop:
This adds a command line argument to the application that when present loads mocha and runs it on the test directories provided. They run in the main process and can open browser windows with a bit of work and would then be able to inject javascript etc' to perform at least some rudimentary testing.


Mossop:
I'd be interested in some feedback on this approach. On the upside the tests run inside the app main process so we don't have to worry about cross process communication problems like we do with spectron. On the other hand it's difficult to do serious testing since all the actual main process code is unobtainable (hidden inside webpack) so all we can do is open browser windows and then inject javascript to do things to them. Maybe that is enough for some basic checks that UI elements are working in the real app?

Either way right now webdriver tests seem to be broken everywhere while this is at least testing something basic.

What do you think @jsantell and @victorporof

build/task-test.js
```diff
@@ -14,8 +14,11 @@ const TEST_DIR = path.join(Const.ROOT, 'test');
 const DEFAULT_UNIT_TESTS = path.join(TEST_DIR, 'unit');
 const DEFAULT_RENDERER_TESTS = path.join(TEST_DIR, 'renderer');
 const DEFAULT_WEBDRIVER_TESTS = path.join(TEST_DIR, 'webdriver');
+const DEFAULT_INTEGRATION_TESTS = path.join(TEST_DIR, 'integration');
 const MOCHA = path.join(Const.NODE_MODULES_DIR, '.bin', 'mocha');
 const ELECTRON_MOCHA = path.join(Const.NODE_MODULES_DIR, '.bin', 'electron-mocha');
+const ELECTRON = path.join(Const.ROOT, '.electron', 'electron') +
```
victorporof: Use the constants module in build/utils for electron paths.

test/integration/utils.js
```diff
@@ -0,0 +1,53 @@
+import path from 'path';
+
+import { BrowserWindow } from 'electron';
+
+const UI_DIR = path.join(__dirname, '..', '..', 'lib', 'ui');
```
victorporof: Ditto, use the consts. Let's not hard code any more paths.

test/integration/utils.js
```diff
@@ -0,0 +1,53 @@
...
+const BROWSER_CHROME_URL = fileUrl(path.join(UI_DIR, 'browser-modern', 'browser.html'));
+
+function fileUrl(str) {
+  let pathName = path.resolve(str).replace(/\\/g, '/');
```
victorporof: A version of this function is already defined in app/shared/paths-util. Unless I'm missing something, you should import that.

test/integration/utils.js
```diff
@@ -0,0 +1,53 @@
...
+        // These have to be hardcoded :(
+        port: '9090',
+        host: 'localhost',
+        version: 'v1',
+      });
+    });
+
+    browser.once('window-ready', err => {
+      if (err) {
+        reject();
```
victorporof: Please always reject with an actual error instance.

test/integration/utils.js
```diff
@@ -0,0 +1,53 @@
...
+
+  // Windows drive letter must be prefixed with a slash.
+  if (pathName[0] !== '/') {
+    pathName = `/${pathName}`;
+  }
+
+  return encodeURI(`file://${pathName}`);
+}
+
+export function openBrowserWindow() {
```
victorporof: Why did this function have to be written? It's very similar to the one in the main process code. Unless I'm again missing something, should generalize some things into smaller utility methods so we don't write the same or similar code twice.

app/main/browser.js
```diff
@@ -54,9 +55,39 @@ app.on('ready', async function() {
...
 
-  await BW.createBrowserWindow(userAgentClient, () => {
+  const bw = await BW.createBrowserWindow(userAgentClient, () => {
     instrument.event('browser', 'READY', 'ms', Date.now() - browserStartTime);
   });
+
+  const options = parseArgs();
+  if (options.test) {
+    bw.once('window-ready', () => {
+      const Mocha = require('mocha'); // eslint-disable-line
```
victorporof: Let's avoid polluting browser.js with inline code. This should be in an utility module or just another file.

test/integration/utils.js
Mossop: So this gets to the root of a problem with this approach. I can import app/shared/paths-utils here but it will be a different instance of the module to the one in the webpack bundle. In this case that doesn't matter, the module doesn't store any state but I worry that for other cases it will and we'll get confused when we expect them to be the same instance.

victorporof: That's a good point. I had similar problems when trying to make `npm run serve` great again. The LIBDIR wasn't declared when importing from the app folder, since that global was only added in webpacked files.

jsantell:
This is slick, but not sure what we should be testing with this style test, for the reasons you mentioned. Maybe good for testing actual main process code that's actually running electron? Testing anything in the browser window itself seems difficult, but we could test things like app events and browser window utilities, etc. (so I don't know if integration is the most descriptive name for this style)

Mossop:
So an alternative here is to take a similar approach but have the test code run in the renderer process and somehow inject the window object into the test's scope. That way you have direct access to the UI DOM and can manipulate as you like. Would that be more useful do you think @jsantell?

jsantell:
@Mossop that sounds like our renderer tests, which loads our frontend in an electron renderer process with mocha tests (so tests have direct access to DOM), but the electron main process is just a shell app, not our own

victorporof:

Issue #1223, opened 2016-09-21
Make the new tab button clickable area larger


victorporof:
It's hard to click at times by people who don't use keyboard shortcuts.

PR #1225, opened 2016-09-21
Make tab hovering and clicking more accurate

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1223 among other things.

Currently clicking and hovering tabs is very imprecise. This is due to the complex overlapping between them. This patch attempts to make everything explicit and precise.

At the same time, it adds the 'ghost tab' background when hovering the new tab button by using the same TabVisuals component used for normal tabs.


jsantell:
Lots of tricky CSS magic that seems like it'd be good to have tested somehow. Not sure best way to do that :\

r+

Issue #1210, opened 2016-09-20
Handle opening links in Tofino on Windows when it's default browser


jsantell:
Bit more complicated than the OSX bit due to non-single instance enforcement, which is working out general weirdness of setting defaults. Will have to use [app.makeSingleInstance](http://electron.atom.io/docs/all/#appmakesingleinstancecallback) and translate process.argv to what OSX is using (open-file, open-url events).

jsantell:
This is related to #1095, which we are not turning on the prompt to set as default on Windows yet, users can set it on their own and this should allow it to work on Windows, so moving to `next`

PR #1230, opened 2016-09-22
Properly handle opening files when OSX is default

jsantell:
* `userAgentClient` is now attached to `electron.app` and we can use that as a global to determine state of the application startup, and simplify interfaces like in browser-window
* Now handles opening via command line on OSX/Windows(?) via `open -a Tofino https://mozilla.org` **needs confirmed in Windows**
* Windows should now be able to open links when it's the default if it's closed or open already -- **needs confirmed**

** To confirm on Windows **

- [ ] Handles `open -a Tofino https://mozilla.org` or whatever the CLI equiv on Windows is
- [ ] When set as default browser, should open a tab for a link while open
- [ ] When set as default browser, should open a tab for a link while closed
- [ ] When opening an incoming link, it should not create a new instance and open it in an existing window/create a new window

This does **not** ask Windows users if they want to set Tofino as a default browser, due to issues with Windows 10 defaults.


jsantell:
This also moves `Deferred` to a shared util, gets rid of the URL_QUEUE as we can now wait for app initialization, and removes the passing in of a callback for `createBrowserWindow` since it also returns a promise.

app/main/command-line.js
```diff
@@ -0,0 +1,82 @@
...
+        return argv[index + 1];
+      }
+    }
+  }
+  return undefined;
+};
+
+// For macOS, there are events like open-url instead
+if (!isDarwin) {
+  const openUrl = getUrlFromCommandLine(process.argv);
```
Mossop: we already use yargs for parsing the command line in the main process, can't we reuse that?

Mossop: Note that the yargs stuff has special handling for electron where sometimes you have an extra random argument.

app/main/command-line.js
```diff
@@ -0,0 +1,82 @@
...
+import * as BW from './browser-window';
+import * as protocols from './protocols';
+
+const { app } = electron;
+const isDarwin = process.platform === 'darwin';
+
+// Checks an array of arguments if it can find a url
+const getUrlFromCommandLine = argv => {
+  if (argv) {
+    if (argv.length === 2 && !argv[1].startsWith('-')) {
```
Mossop: It's not clear to me what this case is for

jsantell: On OSX, it's for `./path/to/tofino http://mozilla.org`

app/main/command-line.js
jsantell: Totally, honestly I don't know of all the expected ways to open this via command line cross platform

jsantell:
Fixes #1210

app/main/browser.js
```diff
@@ -42,6 +44,12 @@ certs.setupCertificateHandlers();
 spawn.startContentService();
 spawn.startUserAgentService(userAgentClient);
 
+// Constant promise used to signal elsewhere in the main process
+// that the application has launched and finished loading
+// the initial browser window
+const appInitializedDeferred = new Deferred();
+app.INITIALIZED = appInitializedDeferred.promise;
```
Mossop: I don't like adding new bits of data to the electron API like this. Can we just expose this sort of thing in its own module?

app/main/command-line.js
Mossop: Well we define it ourselves on windows. I believe OSX is just <exe> <url> and we can make windows do that too. Yargs takes care of stripping out the exe and potential app directory for you though.

jsantell: We can define strict CLI usage, but also there's when you open a link, Tofino is a default app, somewhere somehow Tofino is called with an extra argument of a URL, which we don't have control over. That being said, shouldn't be an issue

app/main/browser.js
jsantell: SGTM

app/main/command-line.js
Mossop: I'm saying we do have control over it in Windows. It is part of what goes into the registry, the exact command line to call.

jsantell: Ah ok, I understand now

jsantell:
Migrating this to just handle the orchestration on OSX, and handle leaving implementing URL parsing only for non-OSX in a follow up in #1210

app/main/command-line.js
```diff
@@ -0,0 +1,77 @@
...
+    const appAlreadyStartedShouldQuit = app.makeSingleInstance(argv => {
+      // Someone tried to run a second instance, we should focus our window.
+      if (isDarwin) {
+        BW.focusOrOpenWindow();
+      } else {
+        BW.focusOrOpenWindow(getUrlFromCommandLine(argv));
+      }
+    });
+    if (appAlreadyStartedShouldQuit) {
+      app.exit(0);
```
Mossop: We do something similar in response to squirrel events in `index.js`. I wonder if those two should live together. Maybe we'll just do that in the other issue.

jsantell: That's probably a better place for this, I agree. I'll handle in the other issue

app/main/browser-window.js
```diff
@@ -143,7 +161,11 @@ export async function createBrowserWindow(userAgentClient, onload) {
     browser.show();
   }
 
-  return browser;
+  // Emit an event to the main process so we can mark the app
+  // as initialized
+  ipcMain.emit('main:new-window-created', browser);
```
jsantell: +1

PR #1236, opened 2016-09-23
Perform automatic update checks and install when found. Fixes #625

### LINKED ISSUE #625: Installed builds should auto-update themselves

Mossop:
This is pretty rough and ready but it works on Windows. If you could test on OSX that would be great, there is an available update in the releases right now so shortly after starting up you should see the prompt to install it.

### DIFF
--- a/app/main/updater.js
+++ b/app/main/updater.js
@@ -1,0 +1,57 @@
+/*
+Copyright 2016 Mozilla
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of the
+License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed
+under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+*/
+
+import os from 'os';
+import { setInterval } from 'timers';
+import { app, autoUpdater, dialog } from 'electron';
+import { version as appVersion, name as appName } from '../../package.json';
+
+// Check once an hour
+const UPDATE_CHECK_TIME = 1000 * 60 * 60;
+
+// Exe name is set to the same as the package name
+app.setAppUserModelId(`com.squirrel.${appName}.${appName}`);
+
+try {
+  autoUpdater.setFeedURL(`https://tofino-update-server.herokuapp.com/update/${os.platform()}_${os.arch()}/${appVersion}`);
+} catch (e) {
+  // This happens when the application isn't code signed on OSX.
+}
+
+autoUpdater.on('update-downloaded', () => {
+  dialog.showMessageBox({
+    type: 'question',
+    title: 'Update Available',
+    message: 'An update is available, would you like to install it?',
+    buttons: ['Yes', 'No'],
+    defaultId: 0,
+    cancelId: 1,
+  }, response => {
+    if (response === 0) {
+      autoUpdater.quitAndInstall();
+    }
+  });
+});
+
+function checkForUpdates() {
+  try {
+    autoUpdater.checkForUpdates();
+  } catch (e) {
+    // This happens if the app isn't installed
+  }
+}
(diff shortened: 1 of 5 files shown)


bgrins:
This is failing on travis with: https://travis-ci.org/mozilla/tofino/jobs/162308500#L1496

bgrins:
I get a crash when running this locally with:

> ./dist/mac/tofino.app/Contents/MacOS/tofino 

2016-09-26 15:49:27.423 tofino[13833:131480] App Transport Security has blocked a cleartext HTTP (http://) resource load since it is insecure. Temporary exceptions can be configured via your app's Info.plist file.
main ERROR: Error: The resource could not be loaded because the App Transport Security policy requires the use of a secure connection.
    at Error (native)

bgrins:
On Windows 7 I see two issues, one big and one small:

1) I got the 'update is available prompt' after running `npm run package` then double clicking on `dist/win/tofino Setup 0.1.0`.  Clicked OK then it closed and reopened into an entirely back screen.  I can reproduce this every time I accept the update

2) there's some kind of text editor gif playing while the app is loading.  If it's easy to change that to another asset we could switch it to something else

app/main/updater.js
```diff
@@ -0,0 +1,57 @@
...
+import { version as appVersion, name as appName } from '../../package.json';
+
+// Check once an hour
+const UPDATE_CHECK_TIME = 1000 * 60 * 60;
+
+// Exe name is set to the same as the package name
+app.setAppUserModelId(`com.squirrel.${appName}.${appName}`);
+
+try {
+  autoUpdater.setFeedURL(`https://tofino-update-server.herokuapp.com/update/${os.platform()}_${os.arch()}/${appVersion}`);
```
victorporof: The OS platform is already set in the build config. Probably a good idea to add arch there as well now that we need it, and use that instead?

app/main/updater.js
```diff
@@ -0,0 +1,57 @@
...
+} catch (e) {
+  // This happens when the application isn't code signed on OSX.
+}
+
+autoUpdater.on('update-downloaded', () => {
+  dialog.showMessageBox({
+    type: 'question',
```
victorporof: Since this is actual UI, it should jump the appropriate hooks so that this it's handled in the frontend sagas, passing through the redux store.

app/main/updater.js
```diff
@@ -0,0 +1,57 @@
...
+    title: 'Update Available',
+    message: 'An update is available, would you like to install it?',
+    buttons: ['Yes', 'No'],
+    defaultId: 0,
+    cancelId: 1,
+  }, response => {
+    if (response === 0) {
+      autoUpdater.quitAndInstall();
```
victorporof: This should be a request coming from the frontend.

victorporof: See the way setting the default browser works.

app/main/updater.js
bgrins: This UI is running in the main process, though.  Unless/until we move the update UI into a content page there's no good way to interact with redux

victorporof: See the way setting the default browser works.

1. main process sends ipc message to renderer process "should-quit-and-update"
2. renderer process receives ipc message and dispatches "CONFIRM_QUIT_AND_UPDATE" effect
3. said effect is handled in the appropriate ui saga
4. saga sends ipc message to main process "show-message-box" with the appropriate details (this will later be replaced with better ui; it also has the benefit of standardizing a way of showing arbitrary message boxes from the ui)
5. upon confirmation, send ipc message to renderer process "${message-box-id}-confirmed" or something similar; if dismissed or cancelled, "${message-box-id}-cancelled"
6. the above mentioned messages are handled in the appropriate ui saga

This might involve a lot of boilerplate, but is very testable and easily changed in the future when we decide to use different/better ui.

bgrins: It seems OK to have some UI that the normal frontend doesn't know about any actions taken in it will never affect the renderer, and is contained within a UI which is only ever rendered by main process.  If we end up wanting to show a UI for this in a content page then that's another story, of course.  

It's just going to be hard to test auto updating in general, especially if it involved quitting and restarting the process.

PR #1251, opened 2016-09-26
Switch to tar.bz2 packages for linux. Fixes #1248

### LINKED ISSUE #1248: Nightly builds don't work on linux
On linux, zip files don't preserve file permissions and you can't successfully launch tofino. (I imagine that's the same for Mac?)

On windows, tofino is completely black. UI "works" as I'm able to close the browser by clicking blindly on the close button on the upper right, but everything is black.

Mossop:


build/task-package.js
```diff
@@ -29,7 +31,7 @@ const fileChanges = {
     copies: [
       [path.join(Const.PACKAGED_DIST_DIR, `win${archIfNotx64}`,
                  `${EXENAME} Setup ${getAppManifest().version}${archIfNotx64}.exe`),
-       path.join(Const.PACKAGED_DIST_DIR, `tofino-win32-${ARCH}.exe`)],
```
bgrins: Will `tofino-${PLATFORM}-${ARCH}` end up the same as `tofino-win32-${ARCH}`?

Mossop: Yes, PLATFORM is just os.platform() which is `win32` on windows.

build/task-package.js
```diff
@@ -40,24 +42,25 @@ const fileChanges = {
...
+       path.join(Const.PACKAGED_DIST_DIR, `${PACKAGENAME}.dmg`)],
     ],
   },
   linux: {
-    zips: [
-      [path.join(Const.PACKAGED_DIST_DIR, 'linux'),
-       path.join(Const.PACKAGED_DIST_DIR, `tofino-${PLATFORM}-${ARCH}.zip`)],
+    moves: [
+      [path.join(Const.PACKAGED_DIST_DIR,
+                 `${APPNAME}-${getAppManifest().version}.tar.bz2`),
```
bgrins: Nit: we should assign `getAppManifest().version` into a `VERSION` const so this fits onto one line (and replace the 4 other usages of `getAppManifest().version` in this file with the var).

Also, there's one use of `getManifest().version()` in this file.  Is that an intentional difference?

Mossop: Not sure about the difference, the other pull request for auto updates adds a test that makes sure the two stay in sync so after that it doesn't matter much so I'll just make these consistent.

PR #1239, opened 2016-09-24
Perf improvements

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1197 and https://github.com/mozilla/tofino/issues/1196 and a few other things. See my email to tofino-dev for more info.

Issue #1193, opened 2016-09-20
Write tests for the new frontend


victorporof:
We have no tests. We're bad people.

victorporof:
I bet we can migrate over at least some of the old tests.

victorporof:
Duping to https://github.com/mozilla/tofino/issues/1519

PR #1257, opened 2016-09-27
Implement pinned tabs

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1204

Depends on https://github.com/mozilla/tofino/pull/1239

@Mossop you wanted this the most so please take a look and see if the functionality you need is there. Double-clicking will pin a tab.

@bgrins UI stuff.

screen shot 2016-09-27 at 5 30 52 pm


bgrins:
First impression on usage:

* double clicking to pin is quick and works for now without context menu
* In FF you can't close a pinned tab with ctrl+w, but it does close here

Mossop:
This looks nice and it's interesting that you can use it as a hacky way to reorder your tabs.

One feature that is missing that we could do in a follow-up is that in Firefox clicking a link that is to a different domain always opens a new tab.

app/ui/browser-modern/model/pages.js
```diff
@@ -13,15 +13,22 @@ specific language governing permissions and limitations under the License.
...
-  // We're using separate Map and List immutables instead of an OrderedMap,
+  // We're using separate Map, Set and List immutables instead of an OrderedMap,
   // because passing in only ids to certain components will guarantee an
   // easy way of avoiding re-renders. A single sorted data structure containing
   // all pages will trigger unwanted re-renders when a sequence of pages is
   // needed (e.g. rendering a tab bar) but certain properties that don't matter
   // on those pages change (e.g. a page `meta` property that's not relevant).
   // Since we want to avoid deep equality checks and rely on shallow comparison,
-  // mapping wouldn't be an option either since it creates new objects.
+  // mapping or getting keys aren't viable option either since those operations
```
bgrins: 'options'

app/ui/browser-modern/reducers/pages.js
```diff
@@ -171,6 +187,18 @@ function setPageState(state, pageId, pageState) {
   });
 }
 
+function setPageUIState(state, pageId, pageUIState) {
+  return state.withMutations(mut => {
+    for (const [key, value] of Object.entries(pageUIState)) {
+      if (!(key in PageUIState.prototype)) {
+        logger.warn(`Skipping setting of \`${key}\` on page state.`);
```
bgrins: Is there an expected scenario where this would happen?  Just wondering if we should instead throw in this case to make sure setting invalid values doesn't creep into the codebase.

app/ui/browser-modern/views/browser/tabbar/tab-pointer-area.jsx
```diff
@@ -44,11 +44,14 @@ class TabPointerArea extends Component {
 
   render() {
     return (
+      /* eslint-disable jsx-a11y/no-static-element-interactions */
```
bgrins: Uber nit: this comment is unbalanced (one is inside of the paren and the other afterwards)

app/ui/browser-modern/views/browser/tabbar/tab.jsx
```diff
@@ -124,14 +125,16 @@ Tab.displayName = 'Tab';
...
+  isPinned: PropTypes.bool.isRequired,
   isActive: PropTypes.bool.isRequired,
   isBeforeActive: PropTypes.bool.isRequired,
   isAfterActive: PropTypes.bool.isRequired,
 };
 
 function mapStateToProps(state, ownProps) {
   const page = PagesSelectors.getPageById(state, ownProps.pageId);
-  const pageIndex = PagesSelectors.getPageIndexById(state, ownProps.pageId);
+  const pageIndex = page ? PagesSelectors.getPageIndexById(state, ownProps.pageId) : -1;
```
bgrins: Why was `: -1` added here?

app/ui/browser-modern/views/browser/tabbar/tab.jsx
```diff
@@ -81,38 +85,35 @@ class Tab extends Component {
...
     } else {
       this.props.dispatch(PageActions.setSelectedPage(this.props.pageId));
     }
   }
 
-  handleTabClose = e => {
-    this.props.dispatch(PageEffects.destroyPageSession(this.props.pageId));
-    e.stopPropagation();
+  handleTabDoubleClick = () => {
+    this.props.dispatch((dispatch, getState) => {
```
bgrins: Why do we need to get the state and run the selector here?  `this.props.isPinned` should be up to date here, right?

So we could do this instead:

    if (this.props.isPinned) {
      this.props.dispatch(PageActions.setPagePinned(this.props.pageId));
    } else {
      this.props.dispatch(PageActions.setPageUnpinned(this.props.pageId));
    }

app/ui/browser-modern/views/browser/tabbar/tab.jsx
bgrins: aka, what's the case where page will be falsy?

app/ui/browser-modern/model/pages.js
victorporof: Yup!

app/ui/browser-modern/reducers/pages.js
victorporof: Not at the moment, and you have a good point. Will update.

app/ui/browser-modern/views/browser/tabbar/tab-pointer-area.jsx
victorporof: Thank you! This is the exact sort of thing that would make my eye twitch.

app/ui/browser-modern/views/browser/tabbar/tab.jsx
victorporof: Correct, this should never happen. There are cases in other places where there might be  no selected tab momentarily, between when closing the last tab and replacing it with a new one. But not here.

app/ui/browser-modern/views/browser/tabbar/tab.jsx
victorporof: Good catch!

victorporof:
Thanks for the fantastic review Brian, as always.

No testing strategy yet, just moving on as we have so far. Will have one soon part of https://github.com/mozilla/tofino/issues/1193

victorporof:
Landed!

Issue #1241, opened 2016-09-24
Shouldn't be able to reload app in production builds by default


victorporof:
Nor should the browser toolbox be accessible. A pref disabling these functionalities is desirable.

PR #1259, opened 2016-09-27
Should not have a 'Reload App' menu item when Tofino is packaged. Fix…

jsantell:
…es #1241. r=vporof

### DIFF
--- a/app/main/browser-window.js
+++ b/app/main/browser-window.js
@@ -89,8 +89,8 @@ export async function createBrowserWindow(userAgentClient, onload) {
   // window opened, we're in a packaged build, on OSX (for now), and this isn't
   // already the default browser.
   const shouldAskToSetDefaultBrowser = browserWindows.length === 0 &&
-                                       process.env.NODE_ENV === 'production' &&
-                                       process.platform === 'darwin' &&
+                                       !BUILD_CONFIG.development &&
+                                       BUILD_CONFIG.platform === 'darwin' &&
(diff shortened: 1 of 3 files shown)


app/main/menu/menus.js
```diff
@@ -21,6 +21,7 @@ import electron from 'electron';
 import items from './submenu-items';
 
 const APP_NAME = electron.app.getName();
+const isProduction = process.env.NODE_ENV === 'production';
```
victorporof: This is the second time I'm seeing us checking NODE_ENV instead of the build config for determining if something is in production. Unless I'm missing something, we should use the build config instead. This environment variable doesn't guarantee a packaged build.

There are good reasons for using a development flag in the build config, we can reiterate outside of this PR if needed.

A much better alternative would be to have a 'production' property on the build config. This would also fix the annoying default browser prompt in local production builds.

app/main/browser-window.js
```diff
@@ -89,7 +89,7 @@ export async function createBrowserWindow(userAgentClient, onload) {
   // window opened, we're in a packaged build, on OSX (for now), and this isn't
   // already the default browser.
   const shouldAskToSetDefaultBrowser = browserWindows.length === 0 &&
-                                       process.env.NODE_ENV === 'production' &&
+                                       !BUILD_CONFIG.development &&
                                        process.platform === 'darwin' &&
```
victorporof: Should check BUILD_CONFIG.platform here.

PR #1262, opened 2016-09-28
Replace existing User Agent Service with implementation in NPM package datomish-user-agent-service

ncalexan:


ncalexan:
@jsantell you and discussed this, so you might be the best person to give it a first review.  The s/uri/url/ patch is fallout from my UA service changes earlier; it makes things that aren't tested against an actual service endpoint work.  Content pages and the searchbar, for example.

app/main/spawn.js
```diff
@@ -63,23 +66,33 @@ export function startUserAgentService(userAgentClient, options = {}) {
...
   if (userAgentClient) {
-    userAgentClient.connect({ port, host, version });
+    connected = userAgentClient.connect({ port, host, version });
   }
+
+  const stop = () => {
+    child.kill('SIGTERM');
+  };
+
+  return [connected, stop, child];
```
jsantell: This is an unusual return signature for js; as an object `{ connected, stop, child }` would be more idiomatic

ncalexan: Oh, you're totally right.  I'll bump this before landing.  Still need to get review on the UAS implementation.

Issue #1264, opened 2016-09-28
Empty results in the location autocompletions list


victorporof:
![screen shot 2016-09-28 at 10 23 04 am](https://cloud.githubusercontent.com/assets/248899/18905902/91d555b6-8565-11e6-8727-50dcfe88b746.png)

![screen shot 2016-09-28 at 10 23 29 am](https://cloud.githubusercontent.com/assets/248899/18905913/9b33b684-8565-11e6-8f24-694d20075112.png)

I started noticing this happening after https://github.com/mozilla/tofino/pull/834

@ncalexan ?

ncalexan:
I think you already figured this all out, but what's happening is that the UAS is returning a `{ title: "", url: <URL> }` pair and there's a uri/url name mismatch, so it's not being shown at all.

PR #1265, opened 2016-09-28
Fix uri/url mismatches

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1264

### DIFF
--- a/app/ui/browser-modern/views/browser/location/index.jsx
+++ b/app/ui/browser-modern/views/browser/location/index.jsx
@@ -68,12 +68,12 @@ class Location extends Component {
 
   handleInputKeyDown = e => {
     if (e.key === 'Enter') {
-      this.handleAutocompletionPick({ uri: e.target.value });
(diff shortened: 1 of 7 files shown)


app/ui/browser-modern/views/browser/location/index.jsx
```diff
@@ -68,12 +68,12 @@ class Location extends Component {
 
   handleInputKeyDown = e => {
     if (e.key === 'Enter') {
-      this.handleAutocompletionPick({ uri: e.target.value });
+      this.handleAutocompletionPick({ url: e.target.value });
```
ncalexan: Unrelated, but this also needs to be fixed -- it should be `{ data: { url: e.target.value } }`, if I read the function below correctly.

ncalexan:
@victorporof thanks for fixing my errors -- I had an equivalent patch at https://github.com/mozilla/tofino/pull/1262/commits/b8e9d537a20c003f74af057f13e635dbd48824c6.

PR #1282, opened 2016-09-28
Display a tofino icon on tofino:// pages and a globe for an indetermi…

### LINKED ISSUE #1275: SSL Icon Display Idea
The SSL icon in the address bar is one of three things:
- Doesn't show when not SSL
- Shows as green when is valid SSL
- Shows as broken lock when invalid SSL

The tiny UI jolt of "not showing, now showing" between every page load is noticeable.  I wonder if we could make the lock icon there persistent, but use a very light grey for when a page isn't on SSL.  Less jank.

jsantell:
…nate state for SSL icons. Fixes #1275. r=vporof

Fixes propogating props for SecurityBadge as well.

The Tofino icon is white, which means you can't see it. When loading/connecting to a page, it uses the globe. Neither is probably preferable in the log run, but we can swap out when we have the appropriate icons.


app/ui/browser-modern/views/browser/location/security-badge.jsx
```diff
@@ -29,19 +29,18 @@ class SecurityBadge extends Component {
...
     return (
       <Btn title="Connection"
-        hidden={this.props.hidden}
-        image={this.props.image}
         imgWidth="16px"
         imgHeight="16px"
-        onClick={this.props.onClick} />
+        image={this.props.image}
+        onClick={this.props.onClick}
+        {...this.props} />
```
victorporof: There's no reason to pass in `{...this.props}`, since this isn't a widget.

app/ui/browser-modern/views/browser/location/security-badge.jsx
```diff
@@ -52,28 +51,24 @@ function mapStateToProps(state, ownProps) {
...
-  // use http -> https redirects, we shouldn't penalize them for a quick interstitial),
-  // just hide the icon
-  let hidden = false;
-  if (!pageLocation ||
-      pageLocation.startsWith(TOFINO_PROTOCOL) ||
+  let image = 'ssl-insecure.svg';
+  if (pageLocation.startsWith(TOFINO_PROTOCOL)) {
+    image = 'logo-tofino.png';
+  } else if (!pageLocation ||
+  // If the URL is empty (occurs during some page loads in our state),
```
victorporof: This occurs only when navigating back/forward. Technically this shouldn't be the case anymore since we have we keep local tab history.

app/ui/browser-modern/views/browser/location/security-badge.jsx
```diff
@@ -52,28 +51,24 @@ function mapStateToProps(state, ownProps) {
...
-  // If we have a tofino:// page, if the URL is empty (occurs during
-  // some page loads in our state), or if the page is still loading (many sites
-  // use http -> https redirects, we shouldn't penalize them for a quick interstitial),
-  // just hide the icon
-  let hidden = false;
-  if (!pageLocation ||
-      pageLocation.startsWith(TOFINO_PROTOCOL) ||
+  let image = 'ssl-insecure.svg';
+  if (pageLocation.startsWith(TOFINO_PROTOCOL)) {
+    image = 'logo-tofino.png';
```
victorporof: We have better icons in the branding/iconset directory. Should use one from there.

app/ui/browser-modern/views/browser/location/security-badge.jsx
```diff
@@ -29,19 +29,18 @@ class SecurityBadge extends Component {
   render() {
     return (
       <Btn title="Connection"
-        hidden={this.props.hidden}
-        image={this.props.image}
         imgWidth="16px"
         imgHeight="16px"
-        onClick={this.props.onClick} />
+        image={this.props.image}
```
victorporof: Uber nit: no reason to move this prop around.

app/ui/browser-modern/views/browser/location/security-badge.jsx
jsantell: It doesn't propogate `style` otherwise

app/ui/browser-modern/views/browser/location/security-badge.jsx
jsantell: Updated comment

app/ui/browser-modern/views/browser/location/security-badge.jsx
victorporof: There's no requirement to propagate `style`, since this isn't a widget. If we followed your logic, all browser frontend components that contain buttons or other widgets should pass in all props to their children; this wouldn't make sense. I don't see why we would need to make an exception for this component.

If you mean `className` instead of `style`, then I see what's going on. But I still disagree that we should pass in all props to the button component (no need for pageId in buttons). Just pass in className.

Issue #1292, opened 2016-09-29
Failed prop type: Required prop `code` and `description` was not specified in `NetErrorPage`


victorporof:
Very interesting warning, only happens when using swipe gestures to navigate back from a certificate error page.

STR:
1. Go to https://badssl.com/
2. Navigate to the "expired" menu item (first one)
3. Swipe to navigate back.

I assume the component isn't updated quickly enough with the new state, because it waits for the appropriate `hidden` prop on the parent component.

PR #1293, opened 2016-09-29
Fix `code` and `description` prop warnings on the NetErrorPage when swipe-navigating back from a page with certificate errors

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1292

See issue for explanation.
This is the most amazing bug. I'm out. *drops mic*


jsantell:
Yeesh, I suspect maybe more issues like this w/r/t page loading state. LGTM!

PR #1297, opened 2016-09-29
Don't allow clicks to go through to the webview when autocomplete popup is opened

### LINKED ISSUE #1207: Clicking outside the autocompletions list to close it should prevent other nodes from reacting to that click
E.g.: click over a link in `tofino://history` to close the autocompletions list. I would expect the completions list to be modal. Same goes for the dropdown history list.

bgrins:
Fixes #1207.

There are two approaches here, the first (can be seen at https://github.com/mozilla/tofino/commit/44cc3e89561e0df852f842a5754fbd1bdea96b9f) binds to the mouseDown on the subcomponent then stops propagation up to the window.

The second (which can be seen as this PR) sets up a ref to the search container so we can do container.contains(e.target) as a lighter-weight way to check if we should prevent the mousedown from happening.

There's a third possible approach here which would be sending a change over to the webview to tell it to use `pointer-events: none` when the autocomplete popup is opened.  This would have the benefit of not showing hover states on things in the page when the popup is opened, but it also feels more heavyweight and less flexible if we want to somehow change this behavior in the future.

I'm leaning towards the approach that's currently in the PR, but @victorporof lmk what you think.

### DIFF
--- a/app/ui/shared/widgets/autocompleted-search.jsx
+++ b/app/ui/shared/widgets/autocompleted-search.jsx
@@ -47,20 +47,24 @@ class AutocompletedSearch extends Component {
 
   componentWillMount() {
     if (typeof window === 'object') {
-      window.addEventListener('click', this.handleWindowClick);
+      window.addEventListener('mousedown', this.handleWindowMousedown);
     }
   }
 
   componentWillUnmount() {
     if (typeof window === 'object') {
-      window.removeEventListener('click', this.handleWindowClick);
+      window.removeEventListener('mousedown', this.handleWindowMousedown);
     }
   }
 
-  handleWindowClick = () => {
+  // We need to use mousedown since preventing the event on click won't stop
+  // it from passing through to the webview.  Note that this will only be fired
+  // when the mousedown happened outside the selection list
+  handleWindowMousedown = (e) => {
     // We can't call `setState` on components which haven't rendered yet.
     if (this.state.showSelectionList && this.inputbar) {
       this.setState({ showSelectionList: false });
+      e.preventDefault();
     }
   }
 
@@ -127,6 +131,13 @@ class AutocompletedSearch extends Component {
     this.props.onKeyDown(e);
   }
 
+  // The the list has been moused down, then don't let that bubble up
+  // to the window, because that would cause the list to become closed
+  // before the item gets selected.
+  handleMouseDownOnSelectionList = (e) => {
+    e.stopPropagation();
+  }
+
   handleMouseOverChildComponent = component => {
     const index = component.props['data-index'];
     this.setState({ selectedIndex: index });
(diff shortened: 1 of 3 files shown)


app/ui/shared/widgets/search.jsx
```diff
@@ -96,6 +96,7 @@ class Search extends Component {
   render() {
     return (
       <div {...omit(this.props, Object.keys(OmittedContainerProps))}
+         ref={e => this.container = e}
```
victorporof: Bad formatting here.

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -47,20 +47,26 @@ class AutocompletedSearch extends Component {
...
-      window.removeEventListener('click', this.handleWindowClick);
+      window.removeEventListener('mousedown', this.handleWindowMousedown);
     }
   }
 
-  handleWindowClick = () => {
-    // We can't call `setState` on components which haven't rendered yet.
-    if (this.state.showSelectionList && this.inputbar) {
+  // We need to use mousedown since preventing the event on click won't stop
+  // it from passing through to the webview.
```
victorporof: Not sure I understand this comment. Preventing click events won't stop the click events from propagating? Am I reading this right? Why "prevent" and not "stop propagation"?

app/ui/shared/widgets/autocompleted-search.jsx
```diff
@@ -47,20 +47,26 @@ class AutocompletedSearch extends Component {
...
+  handleWindowMousedown = (e) => {
+    // We can't call `setState` on components which haven't rendered yet,
+    // and we only want to hide / prevent if the target is outside of the
+    // select list.
+    if (this.state.showSelectionList && this.inputbar &&
+        !this.inputbar.container.contains(e.target)) {
       this.setState({ showSelectionList: false });
+      e.preventDefault();
```
victorporof: Shouldn't this be e.stopPropagation? Both? What is e.preventDefault giving us exactly, I thought this was supposed to stop propagation to the other dom nodes, not prevent default actions?

victorporof:
Need to do the same thing for the back/forward button history lists.

app/ui/shared/widgets/autocompleted-search.jsx
bgrins: This is just from my testing, I haven't found it documented anywhere.  But if you preventDefault() on a click event that happened in a webview it doesn't stop it from happening on the page but if you preventDefault() on a mousedown it does.

stopPropagation shouldn't help here - the event target will be something in the webview and it's already bubbled all the way up to the window by now.

victorporof: Ok, please update the comment to make it a bit more descriptive.

bgrins:
So i’ve done some further testing here and I think mouse events *do* go through when closing awesomebar in FF / chrome.  For instance, try clicking on a link on this page while the popup is opened, and it does follow the link while closing the popup.

I clearly remember it *not* doing that in some cases but can't find that now -- @victorporof do you remember when this click is prevented?

victorporof:
@bgrins I don't remember specific when the click is prevented and when it's not. In my experience, the mouse events are always swallowed when clicking somewhere to close the awesomebar, and that would be what I'd always expect to happen. I think it's reasonable to assume that the number of cases where clicks do go through are of little importance (or, to rephrase, it is more important to prevent clicks going through than it is not).

victorporof:
As discussed on IRC, a solution using `pointer-events: none` and a `nonInteractive` property on the page state used to apply the said css property is being worked on.

victorporof:
Moved to https://github.com/mozilla/tofino/pull/1357

PR #1319, opened 2016-10-03
Increase UAS tests timeout; make UAC connect faster (in tests) . Fixes #1302.

### LINKED ISSUE #1302: Tests started failing/ timing out after landing #1262
@ncalexan See https://github.com/mozilla/tofino/commits/master

ncalexan:


app/shared/user-agent-client.js
```diff
@@ -17,10 +17,15 @@ import { logger } from './logging';
 import { UserAgentHttpClient } from './user-agent-http-client';
 
 class UserAgentClient extends EventEmitter {
-  constructor() {
+  // You'd think { initialConnectDelayMillis = 50, maxConnectDelayMillis = 10000 } would work.  In
+  // fact, you'd be crying in your beer, just like me: it doesn't when you call simply `new
+  // UserAgentClient()`.
+  constructor(options) {
```
victorporof: I believe `{initialConnectDelayMillis = 50, maxConnectDelayMillis = 10000}` does work, but it doesn't have any default param to assign to when there's no param available. Simply adding ` = {}` at the end will work. This comment is unproductive.

ncalexan: Thanks!  I will update with your suggestion, which I interpret to be ```
constructor({ x = 50, y = ... } = {})
```

As for productive, I disagree.  I commented specifically because it looks like it's not "the right thing", and it might save somebody pain who tries to make it look like "the right thing".  The test error this incites is that renderer tests hang with no message or exception, which is pretty hard to traceback to an innocent style change...

victorporof: Your comment directly addresses js syntax. It has nothing to do with with this being hard to debug, or renderer tests hanging. In fact, none of this is mentioned; as it stands, the comment is a rant about js syntax not being intuitive enough in your opinion. Such a comment has no value.

victorporof: If you're referring to ` = {}` as a "style change", that assertion would be wrong. It's not a stylistic or opinionated syntax preference, it's the only correct js syntax.

victorporof:
@ncalexan This doesn't seem to fix the timeouts. I hate to reping in this issue, but the failures hider local development while also making it a bit more difficult than it should be to land pull requests that are, in fact, green otherwise. Any progress? Is there something more complex to be solved than increasing a timeout? Should the original PR be backed out instead?

ncalexan:
On Thu, Oct 6, 2016 at 5:46 AM, Victor Porof <notifications@github.com>
wrote:

> @ncalexan <https://github.com/ncalexan> This doesn't seem to fix the
> timeouts. I hate to reping in this issue, but the failures hider local
> development while also making it a bit more difficult than it should be to
> land pull requests that are, in fact, green otherwise. Any progress? Is
> there something more complex to be solved than increasing a timeout? Should
> the original PR be backed out instead?
>
I'd prefer to disable the test.  The original change has been live for a
week and I haven't witnessed or heard of any issues connecting for local
devs.  We can make the test more robust and re-enable eventually.

victorporof:
I filed https://github.com/mozilla/tofino/issues/1345 to re-enable the tests. Should we close this PR?

ncalexan:
> I filed #1345 to re-enable the tests. Should we close this PR?

Not quite yet; I feel there's a real problem (the timeout failures all hang in the same way) which I'm still working on locally.

victorporof:
@ncalexan what's the status here?

victorporof:

PR #1322, opened 2016-10-04
Add new tab animations. Fixes #1186

### LINKED ISSUE #1186: Add new tab animations
Creating a new tab should animate the way it appears in the tab bar.

victorporof:

### DIFF
--- a/app/ui/browser-modern/views/browser/tabbar/tab.jsx
+++ b/app/ui/browser-modern/views/browser/tabbar/tab.jsx
@@ -34,7 +34,7 @@ const TAB_STYLE = Style.registerStyle({
   alignItems: 'center',
   overflow: 'hidden',
   boxSizing: 'border-box',
-  width: `${UIConstants.TAB_DEFAULT_WIDTH}px`,
+  width: `${UIConstants.TAB_MIN_WIDTH}px`,
   minWidth: `${UIConstants.TAB_MIN_WIDTH}px`,
   height: `${UIConstants.TAB_HEIGHT}px`,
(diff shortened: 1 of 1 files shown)


app/ui/browser-modern/views/browser/tabbar/tab.jsx
```diff
@@ -34,8 +34,12 @@ const TAB_STYLE = Style.registerStyle({
   alignItems: 'center',
   overflow: 'hidden',
   boxSizing: 'border-box',
-  width: `${UIConstants.TAB_DEFAULT_WIDTH}px`,
-  minWidth: `${UIConstants.TAB_MIN_WIDTH}px`,
+  '&[data-mounted=true]': {
+    transition: 'width 0.2s ease-out',
+    width: `${UIConstants.TAB_DEFAULT_WIDTH}px`,
+    minWidth: `${UIConstants.TAB_MIN_WIDTH}px`,
```
bgrins: `minWidth` can be pulled out of the data-mounted selector and left next to the `width` property as it was before (since being mounted should have nothing to do with that property).  Also, `transition` could be added to the non data-mounted portion but up to you

victorporof:
@bgrins What addons? if there's a better solution that might help us add more animation jazz in the long run, let me know!

bgrins:
> @bgrins What addons? if there's a better solution that might help us add more animation jazz in the long run, let me know!

I was looking specifically at https://facebook.github.io/react/docs/animation.html#animate-initial-mounting

victorporof:
I didn't know about that! Thanks!

PR #1331, opened 2016-10-05
Only select the text in the location bar when first focusing the element

Mossop:


app/ui/shared/widgets/search.jsx
```diff
@@ -43,6 +43,8 @@ class Search extends Component {
 
     this.state = {
       interaction: INTERACTION_TYPES.IDLE,
+      focused: false,
+      focusByMouse: false,
```
victorporof: Instead of `focusByMouse`, it'd probably be better to use an interaction type (see INTERACTION_TYPES) so that the input's value doesn't change when the user interacts with it and the redux store decides it's a good idea to update it (even with the same value).

Mossop: Can you explain this more?

victorporof: When the user is typing, a temporary INTERACTION_TYPE local state is set on the local component state which is assigned as an attribute to the actual input element. In the `setURLBarValue` saga, this attribute is checked before trying to set the url bar value so that the input value isn't changed while the user is typing. Such a behavior makes also sense while the input is being user-selected. So, instead of using `focusByMouse` as yet another local state property, just add another INTERACTION_TYPE, making sure to reset it to IDLE when blurring occurs.

victorporof: This can be handled in a followup if you want to.

Mossop:
@victorporof This is ready for another quick review following your comments. There are still one or two things broken with the interaction but they are pretty edge-casey so I'd prefer to file them as follow-ups.

PR #1341, opened 2016-10-06
Display context menus in webviews for links, images and any content.

### LINKED ISSUE #347: Context Menu: Links
Tracking for MVP milestone.
Should not be considered complete without tests.

Should do the following:
* Open in new tab
* Open in new Window
* Copy link address
* Copy 
* Select All
* Inspect Element

Feel free to split up in multiple issues if you're working on this.

jsantell:
Fixes #347. r=vporof

Ideally, we wouldn't need a thunk in the `displayWebviewContextMenu` action creator, but we need dispatch in the saga, since I can't think of a reasonable way to dispatch via yield/put/call in the bound handlers without the saga terminating (and any way I can think of involves needing a 'close' event on the menu which does not exist AFAICT)


app/ui/browser-modern/actions/page-effects.js
```diff
@@ -248,3 +248,12 @@ export function getCertificateError(pageId, url) {
     url,
   };
 }
+
+export function displayWebviewContextMenu(webview, e) {
+  return dispatch => dispatch({
+    type: EffectTypes.DISPLAY_WEBVIEW_CONTEXT_MENU,
+    webview,
+    e,
+    dispatch,
```
victorporof: Don't use dispatch in sagas. Use `put` instead.

app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -170,3 +177,54 @@ function* getCertificateError({ pageId, url }) {
...
+  menu.append(new MenuItem({
+    label: 'Select All',
+    click: () => webview.selectAll(),
+  }));
+
+  menu.append(new MenuItem({ type: 'separator' }));
+
+  menu.append(new MenuItem({
+    label: 'Inspect Element',
+    click: () => webview.inspectElement(e.x, e.y),
```
victorporof: This should dispatch a `TOGGLE_DEVTOOLS` effect instead.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -41,6 +41,7 @@ class Page extends Component {
   constructor(props) {
     super(props);
     this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
+    this.requestContextData = this.requestContextData.bind(this);
```
victorporof: Please use autobinding methods instead.

app/ui/browser-modern/views/browser/page/index.jsx
```diff
@@ -226,6 +230,11 @@ class Page extends Component {
     this.webview.dispatchEvent(ev);
   }
 
+  requestContextData(e) {
```
victorporof: requestContextData = (e) => { ... }

victorporof:
Here's a patch removing `dispatch`: https://pastebin.mozilla.org/8916539

jsantell:
Me and @victorporof discussed the strategy of using `yield put` and `takeLatest` to only have one saga running at a time, even though it will never 'complete' if the user selects no context menu option

PR #1343, opened 2016-10-06
Open attempts to navigate away from a pinned tab's domain in a new tab

Mossop:
We can catch and cancel attempts to navigate away from the main process only so this sends a message to the main process for every pinned tab asking it to monitor the tab for navigations. Any for a new domain are blocked and a new-tab is opened instead.


app/ui/browser-modern/reducers/pages.js
```diff
@@ -186,6 +187,11 @@ function setPageUIState(state, pageId, pageUIState) {
   return state.withMutations(mut => {
     for (const [key, value] of Object.entries(pageUIState)) {
       mut.update('map', m => m.setIn([pageId, 'uiState', key], value));
+
+      if (key === 'pinned') {
+        const id = state.getIn(['map', pageId, 'uiState', 'webContentsId']);
+        ipcRenderer.send('guest-pinned-state', id, value);
```
victorporof: Reducers should be pure in all cases, this should be handled in a saga.

app/ui/browser-modern/model/page-ui-state.js
```diff
@@ -16,6 +16,7 @@ const PageUIState = Immutable.Record({
   zoomLevel: 0,
   searchVisible: false,
   pinned: false,
+  webContentsId: undefined,
```
victorporof: Not sure this makes sense to be in uiState. Why not keep it directly on the page model?

app/ui/browser-modern/reducers/pages.js
victorporof: To elaborate, pinning a tab should become an effect, handled in the ui sagas, which sends this ipc event to the main process while also dispatching an action updating the local page state.

app/ui/browser-modern/model/page-ui-state.js
Mossop: I realised we don't need to cache this at all so removed it.

app/ui/browser-modern/constants/effect-types.js
```diff
@@ -44,6 +44,8 @@ export const SET_PAGE_ZOOM_LEVEL = 'SET_PAGE_ZOOM_LEVEL_EFFECT';
 export const CAPTURE_PAGE = 'CAPTURE_PAGE_EFFECT';
 export const PARSE_PAGE_META_DATA = 'PARSE_PAGE_META_DATA_EFFECT';
 export const GET_CERTIFICATE_ERROR = 'GET_CERTIFICATE_ERROR_EFFECT';
+export const PIN_TAB = 'PIN_TAB';
+export const UNPIN_TAB = 'UNPIN_TAB';
```
victorporof: By convention, effect action string identifiers have an `_EFFECT` suffix so that they're more easily distinguished in console logs.

PR #1352, opened 2016-10-07
Add hotkeys for cmd+shift+r for hard reload, ignoring cache. Fixes #1…

jsantell:
…299. r=vporof

### DIFF
--- a/app/ui/browser-modern/actions/page-effects.js
+++ b/app/ui/browser-modern/actions/page-effects.js
@@ -102,17 +102,19 @@ export function navigateCurrentPageForward() {
   };
 }
 
-export function navigatePageRefresh(pageId, doc = document) {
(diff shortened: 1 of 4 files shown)


app/ui/browser-modern/actions/page-effects.js
```diff
@@ -102,17 +102,19 @@ export function navigateCurrentPageForward() {
   };
 }
 
-export function navigatePageRefresh(pageId, doc = document) {
+export function navigatePageRefresh(pageId, ignoreCache, doc = document) {
```
victorporof: Nit: any reason why we're not just passing in all options here from `navigateCurrentPageRefresh`?

jsantell: No reason one way or another

Issue #1187, opened 2016-09-20
Cmd+Left/Right should go to beginning/end of inputs when focused


victorporof:
At the moment these shortcuts navigate pages back/forward.

jsantell:
Any ideas how this should be implemented? Preload script with a keydown event on the window?

victorporof:
This should support all inputs, including chrome.

PR #1355, opened 2016-10-08
Preserve native hotkeys when input elements are focused, r=jsantell. Fixes 1187

victorporof:
Close https://github.com/mozilla/tofino/issues/1187


app/ui/preload/active-element.js
```diff
@@ -0,0 +1,29 @@
...
+  ipc.sendToHost('focus-data', null);
+}, true);
+
+document.addEventListener('focus', () => {
+  const target = document.activeElement;
+  // We shouldn't send the element itself, only some data about it, since
+  // app state is intended to be immutable and easily serializable.
+  // Furthermore, we can't send the dom node itself via ipc anyway.
+  if (target) {
+    ipc.sendToHost('focus-data', { nodeName: target.nodeName });
```
jsantell: I think a util to take an element and return the serialized form to send to state would be good to be consistent here in the preload script and in the renderer process, like:

```
serializeNode = node => ({ nodeName: node.nodeName });
```

app/main/hotkeys.js
```diff
@@ -44,16 +42,40 @@ const HOTKEYS = new Map([
...
 
 export function unbindBrowserWindowHotkeys(browserWindow) {
+  handlerSet.forEach(handler => ipcMain.removeListener('guest-active-element', handler));
   shortcut.unregisterAll(browserWindow);
 }
 
+function handleGuestActiveElementChange(browserWindow, event, { details }) {
+  if (!details || details.nodeName !== 'INPUT') {
+    CUSTOM_HANDLED_HOTKEYS.forEach((handlerFactory, accelerator) => {
+      shortcut.register(browserWindow, accelerator, handlerFactory(browserWindow));
```
jsantell: Rather than binding and unbinding the shortcuts (is there overhead here?), we could also wrap the handlers to ignore the call rather than unregistering the shortcuts. I think that'd clean up a lot of this handling

```
['CmdOrCtrl+Left', bw => requiresChromeFocus(() => bw.webContents.send('go-back')]),
```

```
const chromeHasFocus = true;
const requiresChromeFocus = (handler) => {
  if (chromeHasFocus) {
    handler();
  }
}
const handleGuestActiveElementChange = (bw, e, { details }) => {
  chromeHasFocus = !details || details.nodeName !== 'INPUT';
};
```

app/ui/preload/active-element.js
victorporof: Good idea.

app/main/hotkeys.js
victorporof: You'd think that were the case, but I've tried that and no. If it were that easy, I could've handled all of this in the frontend without needing to send the currently active element to the main process at all (prevent the actions from happening there).

In short, registered shortcuts swallow the default from happening even when they don't do anything. So if `CmdOrCtrl+Left` is bound, even though the handler does nothing, there won't be anything going on in the input element at all.

victorporof: I'll add a comment.

PR #1344, opened 2016-10-06
Remove browser-blueprint from the repository. Fixes #1070. r=vporof

jsantell:

PR #1364, opened 2016-10-09
Don't include native modules in frontend code

victorporof:
Fixes https://github.com/mozilla/tofino/issues/1363

Naturally this fails right now. The following modules are used when they shouldn't:
```
[
      +  "path"
      +  "url"
      +  "events"
      +  "assert"
]
```

Most of these will go away after https://github.com/mozilla/tofino/pull/1344 lands. The remaining few are super easy to fix (they shouldn't be in app/ui or app/shared at all, just moving them to app/main should fix things).


victorporof:
Updated the PR to make the tests pass. It doesn't care about removing some asserts from browser-blueprint code because #1344 is expected to land soon anyway.

The only problem with the existing test is that it doesn't follow imports from outside of `app/shared` and `app/ui`. That is, if something imports a module from outside of these directories that does, in fact, import native node modules, then the test added in this PR won't fail. However, this PR still makes things much better than they used to be, so it's a good step in the right direction IMO.

test/unit/lint/test-dependencies.js
```diff
@@ -24,6 +24,52 @@ const native = [
 ];
 
 describe('dependencies', () => {
+  let appPackages;
+  let mainPackages;
+  let mainDevPackages;
+  let prodPackages;
+
+  beforeEach(async function() {
```
Mossop: There isn't much point in doing this before every test, maybe move it to the top-level?

Issue #1321, opened 2016-10-04
Frequent "user-agent-http-client.js POST http://localhost:9090/v1/sessions/start 500 (Internal Server Error)" when rapidly opening new tabs


victorporof:

ncalexan:
@victorporof do you see any logging around this?  This is a helpful spur to get the UAS logging boned up.

ncalexan:
A little logging later, I see
```
caught error { Error: SQLITE_ERROR: cannot start a transaction within a transaction
    at Error (native) errno: 1, code: 'SQLITE_ERROR' }
POST /v1/sessions/start 500 20.640 ms - 106
```
I thought we would block transactions at this point, but that's not so -- we need to process them sequentially.  I'll figure out what the JS version of the talking stick is.

ncalexan:
Should be fixed by https://github.com/mozilla/tofino/commit/c905c10883c0724bd5e55bd81bfdbec46a2c5536.  Don't know why this didn't autoclose; guess my commit message syntax wasn't correct?

PR #1380, opened 2016-10-10
Add Tofino to the registry as a handler for html and http and parse urls passed on the command line.

Mossop:
This adds Tofino as a browser to the windows registry on install and removes on uninstall. It's a bit clunky, you get a UAC escalation request from regedit but it works for now. After that you can choose tofino in the windows default apps list.

The command line handling is a bit clunky too. It might be nice to rework that at some point but it works for now.

Currently as long as the app is already open links open in a new tab. If the app isn't open it fails because of #1321.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -15,6 +15,7 @@ specific language governing permissions and limitations under the License.
 
 import { logger } from '../shared/logging';
 import 'source-map-support/register';
+import { addToWindowsRegistry, removeFromWindowsRegistry } from './protocols';
 
 process.on('uncaughtException', err => {
   logger.error(err.stack);
@@ -29,6 +30,8 @@ process.on('unhandledRejection', (reason, p) => {
 
 import { app } from 'electron';
 import { argParser, parseArgs } from '../shared/environment';
+
+// See https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/using/custom-squirrel-events-non-cs.md#application-startup-commands
 
 const SQUIRREL_VERSION_EVENTS = [
   'squirrel-install',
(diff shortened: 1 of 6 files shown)


app/main/index.js
```diff
@@ -71,6 +74,24 @@ switch (seenEvent) {
     require('./browser');
     break;
 
+  case 'squirrel-install':
+    addToWindowsRegistry().then(() => {
```
jsantell: You can call the promise method `catch(fn)` which is like `then(null, fn)`

package.json
```diff
@@ -60,6 +60,7 @@
     "redux-logger": "2.7.0",
     "redux-saga": "0.12.0",
     "source-map-support": "ncalexan/node-source-map-support#fileUrls-plus",
+    "tmp": "0.0.29",
```
jsantell: This is node-only, should it live in app/package.json?

app/package.json
```diff
@@ -12,7 +12,9 @@
   "license": "Apache-2.0",
   "dependencies": {
     "datomish-user-agent-service": "0.0.2",
+    "fs-promise": "0.5.0",
     "request": "2.65.0",
+    "thenify": "3.2.0",
```
victorporof: Potentially unrelated, I don't understand when we should add dependencies to app/package.json exactly. Is there a reason why these aren't in the top level package.json?

victorporof: Should we migrate some of the top level dependencies here instead? Or vice-versa?

Mossop: I wish we could add comments to the json to explain them. It's basically for packages that webpack fails to package for some reason. In this case thenify relies on any-promise which attempts to require a whole bunch of promise modules that we don't have installed which was causing webpack to fail.

victorporof: Same for fs-promise?

Mossop: Yes, it requires any-promise too.

PR #1386, opened 2016-10-10
Download notification should show full path of download. Fixes #1361.…

### LINKED ISSUE #1361: Download notification should also show download location, not just the file name

jsantell:
… r=vporof"

I think the initial implementation didn't have this because of Notification limitations cuts off most text, atleast on OSX (and doesn't appear in the OSX notifications panel, which I don't really use anyway)

screen shot 2016-10-10 at 12 53 50 pm

### DIFF
--- a/app/ui/browser-modern/sagas/ui-sagas.js
+++ b/app/ui/browser-modern/sagas/ui-sagas.js
@@ -70,12 +70,12 @@ function* setURLBarValue({ urlbar, value, doc, options = { keepSelection: true }
   }
 }
 
-function* showDownloadNotification({ filename, status }) {
+function* showDownloadNotification({ path, filename, status }) {
   let message;
   const title = `Download ${status === 'success' ? 'completed' : 'failed'}`;
 
   if (status === 'success') {
-    message = `Download for ${filename} is completed.`;
+    message = `Download for ${filename} is completed. Can be located at ${path}/${filename}`;
   } else if (status === 'error') {
     message = `Download for ${filename} has failed.`;
   }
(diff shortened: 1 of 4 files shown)


app/ui/browser-modern/sagas/ui-sagas.js
```diff
@@ -70,12 +70,12 @@ function* setURLBarValue({ urlbar, value, doc, options = { keepSelection: true }
...
 }
 
-function* showDownloadNotification({ filename, status }) {
+function* showDownloadNotification({ path, filename, status }) {
   let message;
   const title = `Download ${status === 'success' ? 'completed' : 'failed'}`;
 
   if (status === 'success') {
-    message = `Download for ${filename} is completed.`;
+    message = `Download for ${filename} is completed. Can be located at ${path}/${filename}`;
```
victorporof: Instead of "Can be located at...", to save space, can't we just write "... at:"?
"Download for ${filename} completed at ${path}"

victorporof: Additionally, it might be useful to make this path relative to the user's home directory to save up even more space. Like, show `~/Downloads` instead of the full path to the user's downloads folder.

victorporof: These are nits though and can be handled in followups.

jsantell: I don't think any of these strategies guarantee that we can properly display the whole path anyway.

victorporof: Sure!

PR #1394, opened 2016-10-11
Display time last visited in history.

jsantell:
This might be useful, just a start at it and we can think of how better to organize this.


victorporof:
Nice!

victorporof:
@jsantell are we landing this?

jsantell:
Need a reason to justify this; closing for now

PR #1409, opened 2016-10-13
Add context menu commands for navigating back/forward/refreshing the …

### LINKED ISSUE #349: Context Menu: Standard
Should not be considered complete without tests.
Should have basic options of
* Back
* Reload

jsantell:
…page. Fixes #349. r=vporof"

### DIFF
--- a/app/ui/browser-modern/sagas/page-sagas.js
+++ b/app/ui/browser-modern/sagas/page-sagas.js
@@ -11,7 +11,7 @@ specific language governing permissions and limitations under the License.
 */
 
 import { takeLatest, takeEvery } from 'redux-saga';
-import { put, call, apply } from 'redux-saga/effects';
+import { select, put, call, apply } from 'redux-saga/effects';
 
 import Deferred from '../../../shared/deferred';
 import { ipcRenderer, remote, clipboard } from '../../../shared/electron';
@@ -25,6 +25,7 @@ import * as PageEffects from '../actions/page-effects';
 import * as ProfileEffects from '../actions/profile-effects';
 import * as UIEffects from '../actions/ui-effects';
 import * as EffectTypes from '../constants/effect-types';
+import * as PageSelectors from '../selectors/pages';
 
 const { Menu, MenuItem } = remote;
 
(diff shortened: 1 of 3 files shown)


app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -190,10 +190,36 @@ function* getCertificateError({ pageId, url }) {
...
+function* displayWebviewContextMenu({ e, pageId, webview }) {
   const menu = new Menu();
   const chosen = new Deferred();
 
+  menu.append(new MenuItem({
+    label: 'Back',
+    click: () => chosen.resolve(function* () {
+      yield put(PageEffects.navigatePageBack(pageId));
+    }),
+    enabled: webview.canGoBack(),
```
victorporof: Don't use `canGoBack` and `canGoForward` on webviews, we're not doing this anymore, since these won't work after session restore (even when populating them with history, because electron doesn't offer an official API for this yet, so we don't rely on its canGoBack/canGoForward methods for checking if navigation in history is possible, but we do our own checks instead on the local page history).

Use the page state selectors here: https://github.com/mozilla/tofino/blob/master/app/ui/browser-modern/selectors/pages.js#L127 specifically `getPageCanGoBack`, `getPageCanGoForward` and `getPageCanRefresh`.

To access app state in sagas, see https://yelouafi.github.io/redux-saga/docs/api/index.html#selectselector-args. See the session restore PR for examples.

app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -190,10 +190,36 @@ function* getCertificateError({ pageId, url }) {
...
+  }));
+
+  menu.append(new MenuItem({
+    label: 'Forward',
+    click: () => chosen.resolve(function* () {
+      yield put(PageEffects.navigatePageForward(pageId));
+    }),
+    enabled: webview.canGoForward(),
```
victorporof: Ditto.

app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -190,10 +190,36 @@ function* getCertificateError({ pageId, url }) {
...
+  }));
+
+  menu.append(new MenuItem({
+    label: 'Reload',
+    click: () => chosen.resolve(function* () {
+      yield put(PageEffects.navigatePageRefresh(pageId));
+    }),
+    enabled: true,
```
victorporof: Ditto.

jsantell:
Updated to use `yield select(...)`

PR #1434, opened 2016-10-17
Make sagas testable

victorporof:
Changes preliminary to a PR with frontend tests.


app/ui/browser-modern/views/menus/page-context-menu.js
```diff
@@ -0,0 +1,96 @@
...
+import * as ClipboardEffects from '../../actions/clipboard-effects';
+
+const { Menu, MenuItem } = remote;
+
+/**
+ * Creates a context menu for some web contents given a pageId and some
+ * information on what the context menu should be about. Returns a promise
+ * resolved with an action when one of the menu items is picked.
+ */
+export default class PageContextMenu {
```
jsantell: Nice -- I think this would be better suited under models, though

jsantell: In terms of directory structure, nothing else

jsantell: Maybe a method on the PageModel, actually.

PR #1443, opened 2016-10-18
Content pages cleanups

victorporof:


jsantell:
Not sure what these test failures are about, but everything else looks :+1:

`Error: Cannot resolve module 'fs' `

victorporof:
Bunyan is using fs in content pages, and it can't since we're bundling to web. Isomorphic my ass.

PR #1457, opened 2016-10-23
Various small configuration files cleanups

victorporof:


app/ui/shared/util/redux/configure-store.js
```diff
@@ -16,11 +16,17 @@ import thunk from './middleware/thunk';
 import history from './middleware/history';
 import BUILD_CONFIG from '../../../../build-config';
 
+// In webpacked environments, a `process.env.TEST` global will be created,
+// with the value set to the current `TEST` flag on the environment variables.
+/* eslint-disable no-undef */
+const TEST = process.env.TEST;
```
jsantell: Maybe we should just move this to `constants`, which uses `process.env.TEST` rather than having to explain it always

PR #1473, opened 2016-10-25
Use same webpack config for content service as main for resolving app…

jsantell:
…/shared/environment for the logger to fix content service crash.


jsantell:
r=mossop

PR #1477, opened 2016-10-25
Render the statusbar above the searchbar when it's displayed. Fixes #1407

### LINKED ISSUE #1407: Find bar overlaps the status bar
In Firefox, the status bar is rendered above the find bar in this case when both are open.

jsantell:


app/ui/browser-modern/views/browser/decorations/statusbar.jsx
```diff
@@ -58,10 +60,16 @@ StatusBar.displayName = 'StatusBar';
 
 StatusBar.propTypes = {
   statusText: PropTypes.string,
```
victorporof: Since you're here anyway, might want to make `statusText` isRequired as well, since it's always passed in via props. AFAIK empty strings still registered as properly passed in props.

PR #1504, opened 2016-11-01
Don't overwrite the user's text in location bar if the location bar h…

### LINKED ISSUE #1461: Quickly pressing ctrl+t, ctrl+v doesn't end up pasting the clipboard contents into new tab's url bar
STR:
Copy a URL into the clipboard
Press cmd/ctrl+t and cmd/ctrl+v quickly

Expected:
The new tab's url bar is filled with the clipboard contents

Actual:
The URL remains `tofino://history/`

Related: if I wait to paste until the tab is fully loaded, that won't replace the URL either (appends to the end of tofino://history/ instead) since the selection collapses and moves to the end of the url bar.

jsantell:
…as focus. Fixes #1461. r=vporof

### DIFF
--- a/app/ui/browser-modern/sagas/page-sagas.js
+++ b/app/ui/browser-modern/sagas/page-sagas.js
@@ -54,11 +54,14 @@ export default function*() {
   );
 }
 
-export function* createPageSession({ id, location, options, withUI }) {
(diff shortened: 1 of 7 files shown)


victorporof:
I'm still looking at this, it's a bit hairy. Sorry about the wait.

PR #1540, opened 2016-11-07
Only select the URL bar on navigation if the page is the homepage, and

### LINKED ISSUE #1521: Don't select the URL bar when loading a page
* Load a page.
* Click a link.

Expected:
* The page itself has focus, so Space scrolls, Backspace navigates back, etc.

Actual:
* The URL bar has focus. Hitting any key will modify the visible URL.

jsantell:
do not select the URL bar on history forking. Fixes #1521. r=vporof

I think focusing on did-navigate and did-stop-navigation is for selecting it upon a new tab. As a follow up, we should probably mark page models whether or not they have navigated at all to determine focus (if I manually type in tofino://history, I would expect it to not get focused), but that's a bit more of an edge case.

### DIFF
--- a/test/unit/browser-modern/sagas/page-sagas-test.js
+++ b/test/unit/browser-modern/sagas/page-sagas-test.js
@@ -5,13 +5,14 @@ import expect from 'expect';
 import jsdom from 'jsdom';
 
 import { takeEvery, takeLatest } from 'redux-saga';
-import { apply, select, put } from 'redux-saga/effects';
+import { apply, select, put, call } from 'redux-saga/effects';
 import main, * as PageSagas from '../../../../app/ui/browser-modern/sagas/page-sagas';
 import * as EffectTypes from '../../../../app/ui/browser-modern/constants/effect-types';
 import * as PageActions from '../../../../app/ui/browser-modern/actions/page-actions';
 import * as PageEffects from '../../../../app/ui/browser-modern/actions/page-effects';
 import * as PageSelectors from '../../../../app/ui/browser-modern/selectors/pages';
 import * as UIEffects from '../../../../app/ui/browser-modern/actions/ui-effects';
+import * as URLUtil from '../../../../app/ui/shared/util/url-util';
 import userAgentHttpClient from '../../../../app/shared/user-agent-http-client';
 
 describe('page sagas', () => {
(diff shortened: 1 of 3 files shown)


jsantell:
Adding stubs since forking page sagas aren't tested at all

jsantell:
The fork tests will be easier once #1543 lands too with the url utility

test/unit/browser-modern/sagas/page-sagas-test.js
```diff
@@ -177,4 +177,7 @@ describe('page sagas', () => {
 
     expect(gen.next().done).toEqual(true);
   });
+
+  it('should be able to fork a page back in history');
+  it('should be able to fork a page forward in history');
```
victorporof: Why just stubs? It should be fairly straightforward to write proper tests here.

jsantell:
@victorporof added saga test

test/unit/browser-modern/sagas/page-sagas-test.js
```diff
@@ -177,4 +178,36 @@ describe('page sagas', () => {
...
+        location: forkedLocation,
+      }), { id: forkedId }))
+    );
+
+    expect(gen.next().value).toEqual(
+      put(PageActions.setPageIndex(forkedId, 6))
+    );
+    expect(gen.next().value).toEqual(
+      put(PageActions.setSelectedPage(forkedId))
+    );
```
victorporof: Make sure the generator ends here via `expect(gen.next().done).toEqual(true);`

test/unit/browser-modern/sagas/page-sagas-test.js
```diff
@@ -177,4 +178,36 @@ describe('page sagas', () => {
...
+    gen.next(5); // getPageIndexById
+
+    const createPageSessionCall = gen.next(history).value;
+    const forkedId = createPageSessionCall.CALL.args[0].id;
+    const forkedLocation = URLUtil.createHistoryRestoreUrl(history, 0);
+
+    expect(createPageSessionCall).toEqual(
+      call(PageSagas.createPageSession, Object.assign(PageEffects.createPageSession({
+        location: forkedLocation,
+      }), { id: forkedId }))
```
victorporof: You can pass in an id to `createPageSession`, why use Object.assign?

jsantell: +1

PR #1543, opened 2016-11-07
Restoring a session should use historyrestore URLs to preserve histor…

### LINKED ISSUE #1426: Populate webviews history when performing session restore

jsantell:
…y. Fixes #1426. r=vporof

Should land after #1540 due to having the urlbar have focus with the history restore URLs means those ugly URLs stay around after the real page loads.

### DIFF
--- a/app/ui/browser-modern/reducers/root.js
+++ b/app/ui/browser-modern/reducers/root.js
@@ -11,17 +11,31 @@ specific language governing permissions and limitations under the License.
 */
 
 import * as ActionTypes from '../constants/action-types';
+import * as PagesSelectors from '../selectors/pages';
+import { createHistoryRestoreUrl } from '../../shared/util/url-util';
 
 export default function(state, action) {
   switch (action.type) {
     case ActionTypes.OVERWRITE_APP_STATE:
-      return overwriteAppState(state, action.state);
+      return overwriteAppState(state, action.state, action.options);
 
     default:
       return state;
   }
 }
 
-function overwriteAppState(prevState, nextState) {
-  return nextState;
+function overwriteAppState(prevState, nextState, options = {}) {
+  if (!options.restoreHistory) {
+    return nextState;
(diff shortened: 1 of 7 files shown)


app/ui/browser-modern/reducers/root.js
```diff
@@ -11,17 +11,31 @@ specific language governing permissions and limitations under the License.
...
+  if (!options.restoreHistory) {
+    return nextState;
+  }
+
+  return nextState.withMutations(mut => {
+    const ids = mut.getIn(['pages', 'ids']);
+    for (const pageId of ids) {
+      const history = PagesSelectors.getPageHistoryURLs(mut, pageId);
+      const historyIndex = PagesSelectors.getPageHistoryIndex(mut, pageId);
+      const restoreURL = createHistoryRestoreUrl(history, historyIndex);
```
jsantell: Could also be a method on the Page model, if we wanted to start doing that sort of thing

jsantell:
@victorporof added tests for root reducer and url util

test/unit/browser-modern/reducers/root-reducers-test.js
```diff
@@ -0,0 +1,83 @@
...
+import * as pagesSelectors from '../../../../app/ui/browser-modern/selectors/pages';
+import * as ActionTypes from '../../../../app/ui/browser-modern/constants/action-types';
+import RootState from '../../../../app/ui/browser-modern/model/index';
+import Pages from '../../../../app/ui/browser-modern/model/pages';
+
+describe('root reducers', () => {
+  let state;
+
+  beforeEach(() => {
+    state = undefined;
```
victorporof: Shouldn't state = new RootState() here?

jsantell: We could, although the reducer doesn't have any default initial state

PR #1560, opened 2016-11-16
Use ancestor's session ID when forking a tab, or restoring a recently…

### LINKED ISSUE #1555: We do not currently track ancestor ID in forked/restored tabs

jsantell:
… closed tab. Fixes #1555. r=rnewman,vporof


app/ui/browser-modern/sagas/page-sagas.js
```diff
@@ -77,6 +79,9 @@ export function* destroyPageSession({ id, options = {} }) {
 
   if (!options.withoutUI) {
     const currentPageCount = yield select(PageSelectors.getPageCount);
+    const sessionId = yield apply(userAgentHttpClient, userAgentHttpClient.waitForSession, [page]);
+
+    yield put(PageActions.saveRestorablePage(page.id, sessionId));
```
rnewman: If you're feeling particularly brave, changing `page.id` to `page.uuid` would go a long way to improving clarity.

test/unit/browser-modern/sagas/page-sagas-test.js
```diff
@@ -203,6 +221,7 @@ describe('page sagas', () => {
       call(PageSagas.createPageSession, PageEffects.createPageSession({
         location: forkedLocation,
         id: forkedId,
```
rnewman: … 'cos I can't tell if `forkedId` is a UUID or a session entid.

app/shared/user-agent-http-client.js
```diff
@@ -73,6 +74,9 @@ export class UserAgentHttpClient {
       // does not have its own sessionId, we probably hit a race condition,
       // so wait until the page has its sessionId by listening for this promise.
       return this.PAGE_TO_SESSION.get(page.id);
+    } else if (UUIDUtil.isUUID(page) && this.PAGE_TO_SESSION.has(page)) {
+      // Also allow passing in the pageId itself
+      return this.PAGE_TO_SESSION.get(page);
```
jsantell: I think most of this logic is from the original implementation months ago, unsure if all these forks are still necessary/used

app/ui/browser-modern/sagas/page-sagas.js
jsantell: We should probably reconsider some of this -- this is the first time session IDs are exposed to the frontend (we abstract the page id <-> session id relationship in UserAgentHttpClient), hence just now running into this problem

PR #1562, opened 2016-11-18
Enable drag/drop for tabs. Fixes #1167

### LINKED ISSUE #1167: Drag / drop tabs to reorder inside a single window
I'd like to be able to reorder tabs within a window by dragging them around

bgrins:
This adds tab reordering using https://github.com/gaearon/react-dnd - pretty easy.  I'm not sure the best way to test this without using end to end tests though, open to suggestions.

### DIFF
--- a/app/ui/browser-modern/views/browser/tabbar/tab.jsx
+++ b/app/ui/browser-modern/views/browser/tabbar/tab.jsx
@@ -13,6 +13,7 @@ specific language governing permissions and limitations under the License.
 import React, { Component, PropTypes } from 'react';
 import PureRenderMixin from 'react-addons-pure-render-mixin';
 import { connect } from 'react-redux';
+import { DragSource, DropTarget } from 'react-dnd';
 
 import Style from '../../../../shared/style';
 import TabPointerArea from './tab-pointer-area';
@@ -24,6 +25,8 @@ import * as UISelectors from '../../../selectors/ui';
 import * as PagesSelectors from '../../../selectors/pages';
 import * as PageActions from '../../../actions/page-actions';
 import * as PageEffects from '../../../actions/page-effects';
+
+const DRAG_DROP_TYPE = 'TAB';
 
 const TAB_STYLE = Style.registerStyle({
   // Due to its visual elements, this component's bounds are much larger than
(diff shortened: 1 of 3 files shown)


victorporof:
I've never used react-dnd myself. Does this update the app state accordingly as well?

app/ui/browser-modern/views/browser/tabbar/tab.jsx
```diff
@@ -103,7 +106,11 @@ class Tab extends Component {
   }
 
   render() {
-    return (
+    const {
+      connectDropTarget = _ => _,
+      connectDragSource = _ => _,
```
victorporof: I had a mild stroke looking at this. Default props might be better in this case, but we sure love futuristic js so I have no opinions I guess.

app/ui/browser-modern/views/browser/tabbar/tab.jsx
```diff
@@ -130,6 +137,45 @@ Tab.propTypes = {
...
+  };
+}
+
+const dropTarget = {
+  hover: (props, monitor) => {
+    const draggedPageId = monitor.getItem().pageId;
+    const targetPageId = props.pageId;
+    const targetPageIndex = props.pageIndex;
+    if (targetPageId !== draggedPageId) {
+      props.dispatch(PageActions.setPageIndex(draggedPageId, targetPageIndex));
```
victorporof: I guess it does.

app/ui/browser-modern/views/browser/tabbar/tabs-list.jsx
```diff
@@ -59,4 +62,5 @@ function mapStateToProps(state) {
   };
 }
 
-export default connect(mapStateToProps)(TabsList);
+/* eslint-disable new-cap */
+export default connect(mapStateToProps)(DragDropContext(HTML5Backend)(TabsList));
```
victorporof: wow

bgrins:
One more case we should handle is not allowing a non-pinned tab to drag in front of a pinned tab.  Might be follow up material, but should be relatively easy (only allow changing positions on pinned tabs with other pinned, and vice versa)

app/ui/browser-modern/views/browser/tabbar/tab.jsx
bgrins: Could definitely switch to default props.  Still would need the weirdness with returning the first arg though because of the way this wraps the return value in `render`

PR #1567, opened 2016-11-22
Show Loading title instead of history restore URLs for history restore

### LINKED ISSUE #1564: Restored tabs start with the title "tofino://historyrestore?…"
We should show a more useful title, like the title of the page we're restoring, or "Loading…", or something.

jsantell:
URLs. Fixes #1564. r=rnewman

### DIFF
--- a/app/ui/browser-modern/model/page.js
+++ b/app/ui/browser-modern/model/page.js
@@ -18,8 +18,10 @@ import { logger } from '../../../shared/logging';
 import PageMeta from './page-meta';
 import PageState from './page-state';
 import PageUIState from './page-ui-state';
+import { isHistoryRestoreUrl } from '../../shared/util/url-util';
 
 const VERSION = 1;
+const LOADING_TITLE = 'Loading…';
 
 export default class Page extends register(Immutable.Record({
   id: '',
@@ -31,6 +33,12 @@ export default class Page extends register(Immutable.Record({
   uiState: new PageUIState(),
   history: Immutable.List(),
   historyIndex: -1,
+  getTabTitle() {
+    if (isHistoryRestoreUrl(this.location)) {
+      return LOADING_TITLE;
+    }
+    return this.title || this.meta.title || this.location || LOADING_TITLE;
+  },
 }, `Page_v${VERSION}`)) {
   constructor(data) {
     if (!data || !data.id) {
--- a/app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
+++ b/app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
@@ -127,8 +127,9 @@ TabContents.propTypes = {
 function mapStateToProps(state, ownProps) {
   const page = PagesSelectors.getPageById(state, ownProps.pageId);
   const pageIsPinned = PagesSelectors.getPagePinned(state, ownProps.pageId);
+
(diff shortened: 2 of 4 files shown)


app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
```diff
@@ -127,8 +128,14 @@ TabContents.propTypes = {
 function mapStateToProps(state, ownProps) {
   const page = PagesSelectors.getPageById(state, ownProps.pageId);
   const pageIsPinned = PagesSelectors.getPagePinned(state, ownProps.pageId);
+  let title = page.title || page.meta.title || page.location || 'Loading...';
+
+  if (isHistoryRestoreUrl(title)) {
+    title = 'Loading...';
```
rnewman: In both cases, use ellipsis '…' instead of three periods '...'.

app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
```diff
@@ -127,8 +128,14 @@ TabContents.propTypes = {
 function mapStateToProps(state, ownProps) {
   const page = PagesSelectors.getPageById(state, ownProps.pageId);
   const pageIsPinned = PagesSelectors.getPagePinned(state, ownProps.pageId);
+  let title = page.title || page.meta.title || page.location || 'Loading...';
```
rnewman: Worth lifting all of this and below into the prototype of `page`?

app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
```diff
@@ -127,8 +128,14 @@ TabContents.propTypes = {
...
+
```
rnewman: So if I write a page:

```<html>
  <head>
    <title>tofino://historyrestore is super buggy</title>
    …
```

it'll never show a title. (And I haven't checked, but if the title is the empty string, does JS regard that as false, and fall through?)

I think you want logic like this instead:

```
let title = page.title || page.meta.title;
if (title === undefined) {
    if (page.location && !isHistoryRestoreURL(page.location)) {
        title = page.location;
    } else {
        title = 'Loading…';
    }
}
```

app/ui/browser-modern/views/browser/tabbar/tab-contents.jsx
rnewman: That would make it a little easier to write unit tests for this… including for pages with empty titles and URLs as titles.

jsantell: We don't have any prototype methods in the immutable models as of yet, but could add it for this.

jsantell:
@rnewman added `getTabTitle` method

app/ui/browser-modern/model/page.js
```diff
@@ -31,6 +33,12 @@ export default class Page extends register(Immutable.Record({
   uiState: new PageUIState(),
   history: Immutable.List(),
   historyIndex: -1,
+  getTabTitle() {
+    if (isHistoryRestoreUrl(this.location)) {
+      return LOADING_TITLE;
```
rnewman: Ah, that's the better phrasing my brain refused to give me! 😄

PR #1569, opened 2016-11-22
Don't prompt for setting the default browser in test.

rnewman:
The default browser prompt is why this test was perma-failing on macOS.

The test heuristics right at the end also are apparently wrong (at least on macOS), so I commented those out.

This introduces a build config flag `offerDefault`, and uses that instead of switching on `development`.


rnewman:
@mossop doesn't have a Mac; mostly I want CI for this, but @jsantell, could you take a look?

build/tasks.js
```diff
@@ -67,7 +67,7 @@ const Tasks = {
 
   async serve() {
     // Make sure we'll run in development mode to enable logging.
-    await Lazy.config({ development: true });
+    await Lazy.config({ development: true, offerDefault: false });
```
jsantell: We should add `offerDefault: false` in `build/utils/base-config.js` and only offerDefault: true in `task-package.js`

rnewman:
Landed on master, plus backed out the UAS change :/

3da1bda

